PR #1, opened 2016-07-11
node server with universal rendering

skomputer:
Added a small express-based node server that pre-fetches data for any route, renders the front-end as HTML, and provides the application state to the front-end JS.


aslagle:
:+1:

PR #6, opened 2016-07-25
added borrow links in custom book details and auth links in header

### LINKED ISSUE #3: add login button/form for pre-authenticaton

skomputer:
Closes #3 and #4.


src/server/index.tsx
```diff
@@ -42,7 +42,22 @@ function handleRender(req, res) {
         );
         res.status(200).send(renderFullPage(html, state));
       }).catch(err => {
-        res.status(404).send(err);
+        // if 401, render catalog root
```
aslagle: wouldn't we want to render the login popup?

aslagle: that works, so I guess it happens on the client?

skomputer: yes, the client still routes the url as usual

aslagle:
:+1:

PR #8, opened 2016-08-08
Recommendations

skomputer:


aslagle:
I don't like the importants but it's fine for now. Otherwise looks good :+1:

PR #9, opened 2016-08-10
Made category code more robust in case of missing info in the OPDS feed.

aslagle:
If one of the categories is missing a label or scheme we should render the book without the category instead of throwing an exception.


skomputer:
👍

PR #19, opened 2016-08-10
Heroku

aslagle:
This is the only change I needed to run the app on heroku.

When it runs npm install, it needs to run tsc so it will have the server js. 

I also set three environment variables.

heroku config:set NPM_CONFIG_PRODUCTION=false
to install dev dependencies.

and
heroku config:set SIMPLIFIED_PATRON_HOME_URL="http://qa.circulation.librarysimplified.org/groups/"
heroku config:set SIMPLIFIED_PATRON_CATALOG_BASE="http://qa.circulation.librarysimplified.org"


skomputer:
👍

PR #32, opened 2016-09-02
Moved inline styles to scss.

aslagle:
I updated the version of opds-web-client to https://github.com/NYPL-Simplified/opds-web-client/pull/163, and made the same changes here. I also moved the existing css file to scss.


package.json
```diff
@@ -11,9 +11,9 @@
   "scripts": {
     "prepublish": "typings install && tsc && npm run prod",
     "lint": "tslint -c tslint.json src/*.ts src/*.tsx src/**/*.ts src/**/*.tsx src/**/**/*.ts src/**/**/*.tsx",
-    "test": "npm run lint && tsc && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js",
-    "dev": "rm -rf dist && mkdir dist  && cp node_modules/bootstrap/dist/css/* dist && cp css/* dist && NODE_ENV=development webpack --watch --progress --display-modules --config webpack.config",
-    "prod": "rm -rf dist && mkdir dist && cp node_modules/bootstrap/dist/css/* dist && cp css/* dist && NODE_ENV=production webpack -p --progress --display-modules --config webpack.config",
+    "test": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js",
+    "dev": "rm -rf dist && mkdir dist  && cp node_modules/bootstrap/dist/css/* dist && cp node_modules/opds-web-client/dist/*.css dist && cp -r src/stylesheets lib && NODE_ENV=development webpack --watch --progress --display-modules --config webpack.config",
```
aslagle: I don't think hot reload will work for the css yet, since we have dev and prod sharing a webpack config that builds it to a separate file. I might have to come back to that later.

src/server/index.tsx
```diff
@@ -10,8 +11,8 @@ import nyplLogo from "../images/nypl-logo-transparent";
 
 const app = express();
 const port = process.env.PORT || 3000;
-const homeUrl = process.env.SIMPLIFIED_PATRON_HOME_URL || "http://localhost:6500/groups/";
-const catalogBase = process.env.SIMPLIFIED_PATRON_CATALOG_BASE || "http://localhost:6500";
+const homeUrl = process.env.SIMPLIFIED_PATRON_HOME_URL || "http://circulation.alpha.librarysimplified.org/groups/";
+const catalogBase = process.env.SIMPLIFIED_PATRON_CATALOG_BASE || "http://circulation.alpha.librarysimplified.org";
```
aslagle: This is unrelated - I changed the default so it's easier to work on locally without running a circ manager.

mgiraldo:
👍

PR #36, opened 2016-10-03
Added login with clever and moved configuration to a file.

aslagle:
This branch uses https://github.com/NYPL-Simplified/opds-web-client/pull/169 to support logging in with clever (and I need to publish a new version of opds-web-client for the tests to pass here).

Clever and FirstBook have AuthPlugins with the buttons from the app, and the Clever button links out to clever. The FirstBook button shows the basic auth form. Millennium still uses basic auth as well. 

In addition to auth plugins, I've changed the way this app is configured to use a single environment variable that points to a config file. The auth plugin info goes in the config file as well.

It was hard to get webpack to build in the right auth plugins from the config. I ended up having each auth plugin get its own js file, which exports a global variable, and injecting the list of global variables into the main js. This doesn't work for server-side rendering, and we might want to revisit it later.


src/Config.ts
```diff
@@ -0,0 +1,12 @@
+interface Config {
```
winniequinn: I think this would be easier to understand if we had type synonyms for `FilePath`, `URL`, et cetera.

src/Config.ts
```diff
@@ -1,11 +1,14 @@
+type FilePath = string;
+type URL = string;
```
winniequinn: Maybe we can call this `URLString` instead so it doesn't shadow the built-in type.

PR #39, opened 2016-10-06
Handle errors while logging in with clever

aslagle:


src/auth/components/CleverButton.tsx
```diff
@@ -10,7 +10,7 @@ export interface CleverAuthMethod extends AuthMethod {
 
 export default class CleverButton extends React.Component<AuthButtonProps<CleverAuthMethod>, any> {
   render() {
-    let currentUrl = window.location.href;
+    let currentUrl = window.location.origin + window.location.pathname;
```
aslagle: href includes the hash

PR #48, opened 2017-01-09
Instant Classics compatibility

aslagle:
This branch makes it possible to run the catalog with the instant classics OPDS feed. 

In addition to adding the config file and un-hiding the download buttons, I had to make it possible to turn off URL shortening, since it depends on individual OPDS entries having URLs and instant classics only has feeds.

This branch also assumes that openebooks.net will host a static catalog.html page that links to both the instant classics catalog and the open ebooks catalog. The catalog links in the headers of both these catalogs now point to a URL for the static page, and the original catalog links are hidden.

### DIFF
--- a/config/instantclassics.json
+++ b/config/instantclassics.json
@@ -1,13 +1,8 @@
 {
-  "theme": "./src/themes/openebooks.scss",
-  "homeUrl": "/groups/",
-  "catalogBase": "http://qa.circulation.openebooks.us",
-  "catalogName": "Open eBooks",
-  "appName": "Open eBooks",
-  "authPlugins": {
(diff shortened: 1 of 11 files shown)


config/instantclassics.json
```diff
@@ -0,0 +1,16 @@
+{
+  "theme": "./src/themes/instantclassics.scss",
+  "homeUrl": "/https%3A%2F%2Finstantclassics.librarysimplified.org%2Findex.xml",
```
nypl-quinn: Would it be better if `homeUrl` was provided as a URL and `server/index.tsx` performed URL encoding? I would expect to be able to put a URL into a config option named `homeUrl` but I don't think that's the case here.

aslagle: Fixed in 59e69a3

PR #51, opened 2017-09-19
Update for circ manager 2.0 with library names in URLs and new auth document format

aslagle:


aslagle:
The library URL change is similar to a change I already made in the admin interface.

PR #53, opened 2018-10-04
Updated Open eBooks header links.

aslagle:
DPLA just updated openebooks.net and this changes the catalog header to match.

We removed the catalog landing page that linked to the broken Instant Classics catalog so now the "Catalog" link goes directly to this Open eBooks catalog, and also replaced "Contribute" with "FAQs".

### DIFF
--- a/config/local.openebooks.json
+++ b/config/local.openebooks.json
@@ -11,10 +11,10 @@
   "headerLinks": [
     { "title": "Home", "url": "http://openebooks.net/index.html" },
     { "title": "About", "url": "http://openebooks.net/about.html" },
-    { "title": "Contribute","url": "http://openebooks.net/contribute.html" },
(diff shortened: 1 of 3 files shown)


aslagle:
NYPL hosts this so one of you will also need to get this deployed for us.

leonardr:
Is the catalog change just because the Instant Classics catalog view is broken, or would it be happening anyway?

aslagle:
No, it was happening anyway. We'd rather import open content into Open eBooks itself than have a separate catalog.

EdwinGuzman:
This looks good to me.

So DPLA updated openebooks.net but NYPL hosts the catalog app on openebooks.us? This seems weird to me that they are own/updated by different organizations or is this correct?

aslagle:
Yeah that's right. Open eBooks has a bunch of organizations involved.

PR #56, opened 2018-10-09
Multiple library catalog with registry support and dynamic configuration

aslagle:
This branch is a major refactoring that makes it possible to run the application with multiple libraries from a registry or with a single circulation manager. I removed all configuration files and configuration is now read dynamically from the registry and the circ managers' auth documents.


package.json
```diff
@@ -9,14 +9,14 @@
...
     "test": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js lib/**/**/__tests__/*.js",
-    "dev": "rm -rf dist && mkdir dist && cp node_modules/bootstrap/dist/css/* dist && cp -r src/stylesheets lib && NODE_ENV=development UV_THREADPOOL_SIZE=128 webpack --watch --progress --display-modules --config webpack.config",
-    "prod": "rm -rf dist && mkdir dist && cp node_modules/bootstrap/dist/css/* dist && cp -r src/stylesheets lib && NODE_ENV=production UV_THREADPOOL_SIZE=128 webpack -p --progress --display-modules --config webpack.config",
+    "build-dev": "rm -rf dist && mkdir dist && cp node_modules/bootstrap/dist/css/* dist && cp -r src/stylesheets lib && NODE_ENV=development UV_THREADPOOL_SIZE=128 webpack --watch --progress --display-modules --config webpack.config",
+    "build-prod": "rm -rf dist && mkdir dist && cp node_modules/bootstrap/dist/css/* dist && cp -r src/stylesheets lib && NODE_ENV=production UV_THREADPOOL_SIZE=128 webpack -p --progress --display-modules --config webpack.config",
```
aslagle: I renamed these commands to be more consistent with our other web applications.

src/UrlShortener.ts
```diff
@@ -0,0 +1,46 @@
+import { URL } from "universal-url";
```
aslagle: Much of this file was adapted from code that used to be in CatalogHandler and ContextProvider, but I thought it would be nice to have the functions for expanding and shortening URLs in the same place.

src/auth/OAuthPlugin.ts
```diff
@@ -15,7 +15,7 @@ const CleverAuthPlugin: AuthPlugin = {
           let hash = window.location.hash;
           let accessTokenStart = hash.indexOf(accessTokenKey);
           let accessToken = hash.slice(accessTokenStart + accessTokenKey.length).split("&")[0];
-          let credentials = { provider: "Clever", credentials: "Bearer " + accessToken };
+          let credentials = { provider: "OAuth", credentials: "Bearer " + accessToken };
```
aslagle: The auth plugin code probably needs a substantial rewrite as well, but I didn't want to do that now. It currently will not work if the library has an OAuth provider but no basic auth provider, and I don't think it will work if there's more than one OAuth provider. The only library with an OAuth provider is Open eBooks/Clever, so that's fine for now. If we get another OAuth provider we'll have to revisit this.

src/components/BookDetails.tsx
```diff
@@ -166,7 +156,7 @@ export class BookDetails extends DefaultBookDetails<BookDetailsProps> {
     if (this.isBorrowed()) {
       links.push(
         <div className="app-info">
-          Your book is ready to download in the {this.context.appName} app.
+          Your book is ready to download in your reading app.
```
aslagle: This is a lame message but I'm expecting Amigos will redesign this part of the page soon anyway.

If it still includes the app name at that point we can add configuration for it to the circ manager, and maybe also app store/play store ids so we can link to it.

src/components/CatalogHandler.tsx
```diff
@@ -9,79 +10,95 @@ import Footer from "./Footer";
...
-}
+    collectionUrl = this.context.urlShortener.expandCollectionUrl(collectionUrl) || null;
+    bookUrl = this.context.urlShortener.expandBookUrl(bookUrl) || null;
 
-export function expandCollectionUrl(catalogBase: string, url: string): string {
-  return url ?
-    catalogBase + "/" + url :
-    url;
-}
+    let cssVariables = {};
```
aslagle: The following code sets CSS variables for all theme colors needed by the Sass files. I had to set all the variables rather than just foreground and background since Sass's lighten and darken functions don't work on CSS variables.

This is annoying but it works. I'm open to suggestions for better approaches to dynamic theming, or changes to the way I'm managing these variables. @heart-of-code I think I discussed this with you before and you'll likely be doing additional work on it so I'd particularly like to know what you think.

src/components/__tests__/CatalogHandler-test.tsx
```diff
@@ -47,30 +44,57 @@ describe("CatalogHandler", () => {
...
+    expect(cssVariables["--linkhovercolor"]).to.equal("#1a1a1a");
+    expect(cssVariables["--pagetextcolor"]).to.equal("#000000");
+    expect(cssVariables["--pagetextcolorlight"]).to.equal("#000000");
+    expect(cssVariables["--highlightcolor"]).to.equal("#000000");
+
+    let library = {
+      ...context.library,
+      colors: { background: "#000000", foreground: "#ffffff" }
+    };
+    wrapper.setContext({ ...context, library });
```
aslagle: This `...` syntax is a nice alternative to `Object.assign` and I'd like to start using it more widely. See https://redux.js.org/recipes/usingobjectspreadoperator

src/server/Registry.ts
```diff
@@ -0,0 +1,181 @@
+require("isomorphic-fetch");
```
aslagle: @leonardr I'd appreciate it if you could review this file. You can ignore the rest of the PR unless you're interested.

aslagle: This is the code that handles fetching library information from the registry and the authentication document.

src/server/index.tsx
```diff
@@ -1,96 +1,103 @@
...
-const homeUrl = "/" + encodeURIComponent(config.homeUrl || "groups");
-const catalogBase = process.env.SIMPLIFIED_CATALOG_BASE || config.catalogBase || "http://circulation.alpha.librarysimplified.org";
-const catalogName = config.catalogName || "Books";
-const appName = config.appName || "";
-const distDir = process.env.SIMPLIFIED_PATRON_DIST || "dist";
-const authPlugins = Object.keys(config.authPlugins || {});
-const headerLinks = config.headerLinks || [];
-const logoLink = config.logoLink || "";
-const shortenUrls = config.shortenUrls !== undefined ? config.shortenUrls : true;
+const shortenUrls: boolean = !!process.env.SHORTEN_URLS;
```
aslagle: This is false by default but I should probably change that.

src/stylesheets/book_details.scss
```diff
@@ -2,13 +2,6 @@
   margin-bottom: 10px;
 }
 
-.book-details .main .circulation-links .btn {
-  &.download-button, &.revoke-button {
-    /* Download and revoke buttons are hidden until we have a design for them. */
-    display: none;
```
aslagle: Hopefully we will have a design soon :)

I put this back for now mostly to enable downloading open access content.

src/components/__tests__/CatalogHandler-test.tsx
heart-of-code: the spread operator and object deconstruction are SO nice

src/server/Registry.ts
```diff
@@ -0,0 +1,181 @@
...
+  getDataFromAuthDocument(authDocument: AuthDocument) {
+    let logoUrl;
+    for (const link of authDocument.links) {
+      if (link.rel === "logo") {
+        logoUrl = link.href;
+        break;
+      }
+    }
+    let colors = authDocument["web_colors"];
+    let headerLinks = authDocument["web_header_links"];
```
leonardr: Does Javascript cause a problem if you access a nonexistent element of a dictionary? i.e. is this like Python `authDocument["web_colors"]` or like `authDocument.get("web_colors")`?

aslagle: No, JavaScript just returns `undefined`, and there's no equivalent of Python's `get`. It is a problem if you try to access an element of something and it turns out to not be a dictionary though.

config/openebooks.json
```diff
@@ -1,20 +0,0 @@
...
-  "headerLinks": [
-    { "title": "Home", "url": "http://openebooks.net/index.html" },
-    { "title": "About", "url": "http://openebooks.net/about.html" },
-    { "title": "FAQs","url": "http://openebooks.net/faq.html" },
-    { "title": "Resources", "url": "http://openebooks.net/resources.html" },
-    { "title": "Help Center", "url": "https://openebooks.zendesk.com/" },
-    { "title": "Catalog", "url": "/collection/groups" }
-  ],
-  "logoLink": "http://openebooks.net/index.html"
-}
```
leonardr: Where is this configuration to be kept now?

leonardr: I think I get it now -- you put the header links and labels in two lists.

aslagle: Yes. There are a few things I removed entirely - most of the theme is no longer configurable, only background and foreground colors, and appName and logoLink are no longer used for now.

src/components/CatalogHandler.tsx
heart-of-code: We can chat about approaches to dynamic theming after I get better acquainted with the codebase 😄

heart-of-code: I'll try and take a deep dive this weekend, but my schedule is pretty booked this month (was told I wasn't expected to work on the project until ~January)

aslagle: @heart-of-code That's fine, we can definitely revisit this later :)

src/server/Registry.ts
```diff
@@ -0,0 +1,200 @@
...
+      if (link.rel === "logo") {
+        logoUrl = link.href;
+        break;
+      }
+    }
+    let colors = authDocument["web_colors"];
+
+    let headerLinks = [];
+    for (const link of catalog["links"]) {
+      if (link.rel === "related") {
```
aslagle: It would be better to look at the role here, that will have to be added to opds-feed-parser" first.

src/UrlShortener.ts
```diff
@@ -0,0 +1,46 @@
...
+  private readonly catalogUrl: string;
+  private readonly enabled: boolean;
+
+  constructor(catalogUrl: string, enabled = true) {
+    this.catalogUrl = catalogUrl;
+    this.enabled = enabled;
+  }
+
+  private catalogBase(): string {
+    // The origin and library short name in a catalog URL are safe to remove.
```
EdwinGuzman: Why is this the case?

src/__tests__/UrlShortener-test.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    it("expands collection url", () => {
+      let url = host + collectionUrl;
+      expect(shortener.expandCollectionUrl(collectionUrl)).to.equal(url);
+    });
+
+    it("expands book url", () => {
+      let url = host + "works/" + bookUrl;
+      expect(shortener.expandBookUrl(bookUrl)).to.equal(url);
```
EdwinGuzman: Aside, I really like how you don't need to worry about calling a method in typescript and _not_ include a parameter because it'll complain about the missing param. So there's really no need to test something like `shortener.expandBookUrl()`.

src/auth/OAuthPlugin.ts
EdwinGuzman: Where is the auth plugin code located?

src/auth/components/BasicAuthButtonWithImage.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from "react";
+import { AuthProvider, BasicAuthMethod } from "opds-web-client/lib/interfaces";
```
EdwinGuzman: I don't think `AuthProvider` is being used in this file.

src/auth/components/BasicAuthButtonWithImage.tsx
```diff
@@ -0,0 +1,30 @@
...
+    let image;
+    for (const link of this.props.provider.method.links || []) {
+      if (link.rel === "logo") {
+        image = link.href;
+        break;
+      }
+    }
+    let label = this.props.provider.method.description ? "Log in with " + this.props.provider.method.description : "Log in";
+    return (
+      <button href="#" className="basic-auth-button" aria-label={label} autoFocus={true}>
```
EdwinGuzman: Why `autoFocus` here? The `autoFocus` may not be good for accessibility since the focus suddenly jumps on the button.

EdwinGuzman: Also, `href` is not for `button`s so you don't need it here.

src/server/Registry.ts
```diff
@@ -0,0 +1,200 @@
...
+          // anyway.
+          console.warn(catalogError);
+          throw "This library is not available.";
+        }
+
+        let authDocument = currentEntry && currentEntry.authDocument;
+        try {
+          authDocument = await this.getAuthDocument(catalog);
+        } catch (authDocError) {
+          // If we can't get the authentication document, keep the previous cached
```
EdwinGuzman: Will there be any issues to proceed without an authentication document?

src/server/Registry.ts
```diff
@@ -0,0 +1,200 @@
...
+          // version or proceed without one.
+          console.warn(authDocError);
+        }
+        this.CACHE[library] = { registryEntry, catalog, authDocument, timestamp: new Date().getTime() };
+      } catch (error) {
+        console.warn(error);
+        throw "This library is not available.";
```
EdwinGuzman: The nested try/catch are _throw_ ing me off, but it looks unavoidable.

src/server/Registry.ts
aslagle: We would be missing everything that would normally come from the auth document, like colors and authentication options. But you would still be able to browse the catalog, so it would probably be better than not being able to do anything.

src/auth/components/BasicAuthButtonWithImage.tsx
```diff
@@ -0,0 +1,30 @@
...
+    let image;
+    for (const link of this.props.provider.method.links || []) {
+      if (link.rel === "logo") {
+        image = link.href;
+        break;
+      }
+    }
+    let label = this.props.provider.method.description ? "Log in with " + this.props.provider.method.description : "Log in";
+    return (
+      <button className="basic-auth-button" aria-label={label} autoFocus={true}>
```
EdwinGuzman: How does this work if it doesn't have an `onClick` handler?

aslagle: The `onClick` is here: https://github.com/NYPL-Simplified/opds-web-client/blob/master/packages/opds-web-client/src/components/AuthProviderSelectionForm.tsx#L71

EdwinGuzman: I see. I rather not have `onClick` handlers on elements that don't natively have it or might not support. But it could be a different PR.

aslagle: I could pass it down to the button component instead.

EdwinGuzman: Okay, if possible. Thanks.

aslagle:
The only auth providers with images are first book and clever. You can set them up with fake info to test that the button images show up and clicking them gives you the appropriate form or redirect, but you'd need to use real credentials for those services if you want to test actually logging in.

src/auth/components/BasicAuthButtonWithImage.tsx
aslagle: I tested the autofocus again, without it when you click something that gives you the popup, the focus stays where you were on the page, e.g. in the navigation. Then you have to go through the rest of the navigation before you get to the popup.

PR #57, opened 2018-12-21
Added config file option

aslagle:
This branch adds a third way to run the application - with a config file mapping library paths to circulation manager URLs.

When running with a single library, I also removed the requirement to specify the library name since it can be found in the auth document.


src/components/ContextProvider.tsx
```diff
@@ -37,6 +37,9 @@ export default class ContextProvider extends React.Component<ContextProviderProp
       if (bookUrl) {
         path += `/book/${this.urlShortener.prepareBookUrl(bookUrl)}`;
       }
+      if (!path) {
+        path = "/";
```
aslagle: This was a probably unrelated bug I found where the catalog link sometimes didn't work.

src/server/index.tsx
```diff
@@ -1,96 +1,111 @@
...
 import { match, RouterContext } from "react-router";
 import { singleLibraryRoutes, multiLibraryRoutes } from "../routes";
 import ContextProvider from "../components/ContextProvider";
 import buildInitialState, { State } from "opds-web-client/lib/state";
-import Registry from "./Registry";
+import LibraryDataCache from "./LibraryDataCache";
 import UrlShortener from "../UrlShortener";
 import { LibraryData } from "../interfaces";
 
-const app = express();
```
aslagle: Most of this code is the same but I wrapped it all in a function so I could use async/await.

src/server/index.tsx
```diff
@@ -1,96 +1,111 @@
...
+
+  let registryBase = process.env.REGISTRY_BASE;
+
+  const circManagerBase = process.env.SIMPLIFIED_CATALOG_BASE;
+  let routes = circManagerBase ? singleLibraryRoutes : multiLibraryRoutes;
+
+  const config = {};
+  const configFile = process.env.CONFIG_FILE;
+
+  if (configFile) {
```
aslagle: This part is new.

aslagle:
CC @alsrlw @jonathangreen

README.md
```diff
@@ -2,18 +2,25 @@
 A Circulation Manager web interface for library patrons.
 
 ## Running the application
-This application requires either a [library registry](https://github.com/NYPL-Simplified/library_registry) or a [circulation manager](https://github.com/NYPL-Simplified/circulation) to run. By default, it expects a library registry to be running at http://localhost:7000.
-Any circulation manager you'll be using with the app also needs a configuration setting to turn on CORS headers. In the admin interface, go to the Sitewide Settings section under System Configuration and add a setting for "URL of the web catalog for patrons". For development, you can set this to "*", but for production it should be the real URL where you will run the catalog.
+There are three ways to run this application: with a [library registry](https://github.com/NYPL-Simplified/library_registry), with a single library on a [circulation manager](https://github.com/NYPL-Simplified/circulation), and with a configuration file for multiple circulation manager URLs. By default, it expects a library registry to be running at http://localhost:7000.
```
EdwinGuzman: So the default option is the first option you mentioned? Might a bulleted list be easier to read?

README.md
```diff
@@ -2,18 +2,25 @@
...
 
 Once you have a library registry or circulation manager, run `npm install` in this repository to set up dependencies.
 
-Then run either `npm run dev` or `npm run prod` to start the application. `npm run dev` will watch the code for changes and rebuild the front-end code, but won't reload the server code.
+Then you can run either `npm run dev` or `npm run prod` to start the application. `npm run dev` will watch the code for changes and rebuild the front-end code, but won't reload the server code.
 
-To configure a library registry url, set the environment variable `REGISTRY_BASE`. To use a circulation manager, set `SIMPLIFIED_CATALOG_BASE` and `SIMPLIFIED_CATALOG_NAME`.
+Set one of the following environment variables:
+To configure a library registry url, set the environment variable `REGISTRY_BASE`.
+To use a circulation manager, set `SIMPLIFIED_CATALOG_BASE`.
```
EdwinGuzman: You don't need `SIMPLIFIED_CATALOG_NAME` anymore?

src/components/ContextProvider.tsx
EdwinGuzman: It didn't work because it didn't route to the home link?

README.md
aslagle: Yes.

README.md
aslagle: Right, it gets that from the circ manager now.

src/components/ContextProvider.tsx
aslagle: Yeah, I think the problem was that it ended up with `""` as the href in a link, so when you clicked it didn't do anything.

aslagle: The problem was when you clicked the "Catalog" link in the header.

src/server/LibraryDataCache.ts
```diff
@@ -105,60 +123,76 @@ export default class Registry {
...
-        rootUrl = link.href;
-        break;
+  async getRegistryEntry(library): Promise<void> {
+    const template = await this.getLibraryUrlTemplate();
+    const libraryUrl = template.replace("{uuid}", library);
+    try {
+      const registryResponse = await fetch(libraryUrl);
+      const registryCatalog = await registryResponse.json();
+      const registryEntry = registryCatalog.catalogs[0];
+      if (!registryCatalog.catalogs || registryCatalog.catalogs.length !== 1) {
```
EdwinGuzman: Are you retrieving just the first catalog of potentially many from `registryCatalog.catalogs`? Why? And for this line, are you checking specifically for `!== 1` because you expect multiple ones?

jonathangreen:
I did some testing of this on the LYRASIS test servers, looks good to me.

EdwinGuzman:
Works well. I tested starting the catalog up with the Library Registry and a single library with a Circulation Admin but not the config file and multiple Circulation Admins.

src/server/index.tsx
```diff
@@ -1,96 +1,111 @@
...
-        if (!shortenUrls || preparedCollectionUrl) {
-          if (library) {
-            res.redirect(302, "/" + library + "/collection/" + preparedCollectionUrl);
-          } else {
-            res.redirect(302, "/collection/" + preparedCollectionUrl);
+    } else {
+      configText = fs.readFileSync(configFile, "utf8");
+    }
+    for (let entry of configText.split("\n")) {
+      let [path, circManagerUrl, ...ignore] = entry.split("|");
```
EdwinGuzman: Do you _have_ to deconstruct the rest of the values (if any)? `let [path, circManagerUrl]  = entry.split("|");` won't work? Just curious.

aslagle: You're right, that would work too.

README.md
```diff
@@ -2,18 +2,31 @@
...
+There are three ways to run this application:
+* with a [library registry](https://github.com/NYPL-Simplified/library_registry)
+* with a single library on a [circulation manager](https://github.com/NYPL-Simplified/circulation)
+* with a configuration file for multiple circulation manager URLs
+
+By default, it expects a library registry to be running at http://localhost:7000.
+
+Any circulation manager you'll be using with the app also needs a configuration setting to turn on CORS headers. In the admin interface, go to the Sitewide Settings section under System Configuration and add a setting for "URL of the web catalog for patrons". For development, you can set this to "*", but for production it should be the real URL where you will run the catalog. If you are using a library registry, this configuration will automatically be created when you registry libraries with the registry, but you need to configure the URL in the registry by running `bin/configuration/configure_site_setting --setting="web_client_url="http://library.org/{uuid}"` (replace the URL with your web client URL). Otherwise, you'll need to create a sitewide setting for it in the circulation manager.
```
EdwinGuzman: I think you're missing an extra `"` in the configuration command for the `web_client_url` key.

README.md
```diff
@@ -2,18 +2,31 @@
...
+Then you can run either `npm run dev` or `npm run prod` to start the application. `npm run dev` will watch the code for changes and rebuild the front-end code, but won't reload the server code.
 
-To configure a library registry url, set the environment variable `REGISTRY_BASE`. To use a circulation manager, set `SIMPLIFIED_CATALOG_BASE` and `SIMPLIFIED_CATALOG_NAME`.
+Set one of the following environment variables:
+* To configure a library registry url, set the environment variable `REGISTRY_BASE`.
+* To use a circulation manager, set `SIMPLIFIED_CATALOG_BASE`.
+* To use a configuration file, set `CONFIG_FILE` to point to a local file or a remote URL. Each line in the file should be a library's desired URL path in the web catalog and the library's circ manager URL, separated by a pipe character. For example:
+```
+library1|http://circulationmanager.org/L1
+library2|http://circulationmanager.org/L2
```
EdwinGuzman: I initially got a bit confused at first because I thought this option allowed for multiple circulation manager URLs as in multiple circulation managers and multiple urls for each circulation manager. I'm not sure if that's intended but it seems like one circulation manager and multiple URLs is what is explained here. In any case, both situations do work.

aslagle: Yeah, this doesn't care if the libraries are on the same circ manager or not.

src/components/ContextProvider.tsx
EdwinGuzman: Does something similar need to be done in the `opds-web-client` repo?

aslagle: I don't think so since that repo doesn't have URL shortening - it would never end up with an empty string as the collection.

PR #58, opened 2019-02-07
Test update sass

alsrlw:
This includes a minor change to update the version of node-sass to work with node:8.x and a code condition to allow for comments (with #) in the configuration of the library config list file. In passing, do we need to check for blank lines and skip them as well?

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -36,8 +36,10 @@ const initialize = async () => {
       configText = fs.readFileSync(configFile, "utf8");
     }
     for (let entry of configText.split("\n")) {
-      let [path, circManagerUrl] = entry.split("|");
-      config[path] = circManagerUrl;
+      if ( entry.charAt(0) != '#' ) {
+        let [path, circManagerUrl] = entry.split("|");
+        config[path] = circManagerUrl;
+      }
     }
   }
 
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -36,7 +36,7 @@ const initialize = async () => {
       configText = fs.readFileSync(configFile, "utf8");
     }
     for (let entry of configText.split("\n")) {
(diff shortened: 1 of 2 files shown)


src/server/index.tsx
```diff
@@ -36,8 +36,10 @@ const initialize = async () => {
       configText = fs.readFileSync(configFile, "utf8");
     }
     for (let entry of configText.split("\n")) {
-      let [path, circManagerUrl] = entry.split("|");
-      config[path] = circManagerUrl;
+      if ( entry.charAt(0) != '#' ) {
```
aslagle: The travis build is failing due to a couple lint errors - it's a best practice in JS to use `!==` instead of `!=` (and `===` instead of `==`), and we have it configured to want double quotes instead of single quotes which is kind of arbitrary but just for consistency.

aslagle:
This looks good except for the lint errors and I do think it would be fine to skip blank lines.

Thanks Robert!

src/server/index.tsx
alsrlw: I've made the changes and added a check to ensure the line is not blank or null (I think!).

src/server/index.tsx
```diff
@@ -36,8 +36,10 @@ const initialize = async () => {
       configText = fs.readFileSync(configFile, "utf8");
     }
     for (let entry of configText.split("\n")) {
-      let [path, circManagerUrl] = entry.split("|");
-      config[path] = circManagerUrl;
+      if ( entry || entry.charAt(0) !== "#" ) {
```
aslagle: Shouldn't this be `&&` instead of `||`?

alsrlw: Gee, duh! Sorry, I had originally phrased that condition differently and forgot to change the boolean when I saw the preferred way to handle it. Node noob at work :-)

PR #59, opened 2019-05-02
Updated packages

EdwinGuzman:
Updating packages, mainly:
- `opds-web-client`
- `typescript`
- `webpack`
- `@types` instead of `typings`

Also updated unit tests based on the `fetch-mock` mocking package.


EdwinGuzman:
Managed to get the production javascript build to go from 3.5MB down to 1.5MB.

EdwinGuzman:
The app should build and run just as before so there should be no noticeable changes. No new features were added but some unit tests were updated.

src/server/__tests__/LibraryDataCache-test.ts
```diff
@@ -40,42 +41,40 @@ describe("LibraryDataCache", () => {
...
           href: template
         }]
       };
-      let mockFetch = stub().returns(new Promise<any>((resolve) => {
-        resolve({ json: () => {
-          return registryCatalog;
-        }});
-      }));
-      fetch = mockFetch as any;
+      // Originally was { json: () => registryCatalog } for the return promise value
```
aslagle: We don't need to keep track of this - git always has the history if we really want to know what it used to be.

EdwinGuzman: Yeah, this was more for me while updating the tests to see the change. Can remove.

heart-of-code:
> Looks good to me but I haven't tested it.
> 
> @heart-of-code, do you want to make sure you can run this before it's merged?

Looks good! Built and ran fine

PR #61, opened 2019-05-27
Library Data CSS

heart-of-code:
Pulls in the CSS hrefs from the authentication document and adds them to the HTML header


aslagle:
Looks great! Could you just also update src/server/__tests__/LibraryDataCache-test.ts so one of the tests uses your new code?

heart-of-code:
@aslagle thanks for the callout, updated the `getDataFromAuthDocumentAndCatalog` test to include the css links!

PR #63, opened 2019-07-18
Adding tslint-react-a11y.

EdwinGuzman:
Resolves [SIMPLY-2135](https://jira.nypl.org/browse/SIMPLY-2135).
I set the `option`s to take in `aria-selected={false}` because those values were not updating when a new value was selected (the form's select value isn't stored and only used when submitting). I think it's fine for now and we can review it again when bigger updates are being made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## CHANGELOG
 
+### 0.2.1
+#### Added
+- `tslint-react-a11y` to review accessibility for React components.
+
 ### 0.2.0
 #### Updated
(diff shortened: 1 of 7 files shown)


aslagle:
I was going to suggest just disabling the aria-selected rule, but it doesn't look like the package supports it :(

EdwinGuzman:
Right, it's a package that is using a different package as its base and only enabling the accessibility rules. This may need to change in the future as `tslint` will stopped being maintained and preference will be to use `eslint` (after 2020).

PR #64, opened 2019-12-03
Prettier

kristojorg:
## Prettier
Similar to my PR in `opds-web-client`, this sets up prettier on the master branch of this repo. 

## Tslint to Eslint
It also transitions the repo from using tslint to eslint. Tslint is no longer being maintained, and the previous maintainers have suggested migrating to eslint for all linting. So, this does that. I attempt to get the accessibility rules ported by implementing `jsx-a11y`, but I can't be sure it is exactly the same as our previous accessibility rules. Ultimately, the way to solve this issue will be to implement `react-axe`, which is the most up to date and comprehensive way to ensure accessibility compliance (according to the maintainers of jsx-a11y and our previously used `react-a11y` tslint rules. So implementing react-axe is on my to do list. 

## Typescript upgrade
We also needed to upgrade typescript to get all of this working (and also will be needed for future updates). So this updates typescript and fixes associated errors. 

Let me know if you all have any issues with this for now.


kristojorg:
I'm closing this because it isn't ready yet. Typescript upgrading is giving me some errors.

PR #65, opened 2019-12-04
prettier

kristojorg:
Ok this PR just runs prettier on all ts, tsx, and json files. It does not set up prettier as a pre-commit hook or change the eslint/tslint rules to comply with prettier. Doing so requires more work, which I am doing in the `beta` branch. But I wanted to format all the files on master so that the beta branch changes don't look like I changed every single file.


kristojorg:
I'll need you to merge, thanks !

PR #72, opened 2020-02-05
Updating the README for documentation on app start up.

EdwinGuzman:
Resolves [SIMPLY-2542](https://jira.nypl.org/browse/SIMPLY-2542)
Small PR to add more details on how to start the app. Mostly came up from asking @adriana-alter to review and try all combinations herself, and finding that some don't work (mostly missing that libraries should be registered in order for the app to work).

Also added a separate file for the license.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,52 +1,52 @@
 # circulation-patron-web
-A Circulation Manager web interface for library patrons.
+A Circulation catalog web interface for library patrons.
 
-## Running the application
+## Installing
+Once you have a [Library Registry](https://github.com/NYPL-Simplified/library_registry) or [Circulation Manager](https://github.com/NYPL-Simplified/circulation), run `npm install` in this repository to install the dependencies.
+
+## Running the Application
+Once the dependencies are installed, the following two base commands can be used to start the application:
+
+* `npm run dev` - This command will watch the code for changes and rebuild the front-end code, but won't reload the server code.
+* `npm run prod` - This will generate the build that will be used in production servers.
+
+The application will start at the base URL of `localhost:3000`.
+
+### Application Startup Configurations
 There are three ways to run this application:
-* with a [library registry](https://github.com/NYPL-Simplified/library_registry)
-* with a single library on a [circulation manager](https://github.com/NYPL-Simplified/circulation)
-* with a configuration file for multiple circulation manager URLs
+* with a [Library Registry](https://github.com/NYPL-Simplified/library_registry)
+* with a single library on a [Circulation Manager](https://github.com/NYPL-Simplified/circulation)
+* with a configuration file for multiple Circulation Manager URLs
 
-By default, it expects a library registry to be running at http://localhost:7000.
-
-Any circulation manager you'll be using with the app also needs a configuration setting to turn on CORS headers. In the admin interface, go to the Sitewide Settings section under System Configuration and add a setting for "URL of the web catalog for patrons". For development, you can set this to "*", but for production it should be the real URL where you will run the catalog. If you are using a library registry, this configuration will automatically be created when you register libraries with the registry, but you need to configure the URL in the registry by running `bin/configuration/configure_site_setting --setting="web_client_url=http://library.org/{uuid}"` (replace the URL with your web client URL). Otherwise, you'll need to create a sitewide setting for it in the circulation manager.
-
-Once you have a library registry or circulation manager, run `npm install` in this repository to set up dependencies.
-
-Then you can run either `npm run dev` or `npm run prod` to start the application. `npm run dev` will watch the code for changes and rebuild the front-end code, but won't reload the server code.
+By default, this application expects a Library Registry to be running at http://localhost:7000.
 
 Set one of the following environment variables when running the application:
-* To configure a library registry url, set the environment variable `REGISTRY_BASE`.
-   * This is the default setting which will point to the Library Registry located at `localhost:7000`. The libraries can be viewed in the app (running locally) by going to `localhost:3000/{uuid}` where `uuid` is the `uuid` of the library.
-* To use a circulation manager, set `SIMPLIFIED_CATALOG_BASE`.
-   * Point this environment variable to the url of the Circulation Manager, for example `SIMPLIFIED_CATALOG_BASE=http://localhost:6500 npm run prod`. This will load the main library in the Circulation Manager in the app (running locally) by going to `localhost:3000`.
-* To use a configuration file, set `CONFIG_FILE` to point to a local file or a remote URL. Each line in the file should be a library's desired URL path in the web catalog and the library's circ manager URL, separated by a pipe character. For example:
+* `REGISTRY_BASE` - to use a Library Registry
+   * Example: `REGISTRY_BASE=localhost:7000 npm run prod`
+   * This is the default setting which will point to the Library Registry located at `localhost:7000`. The libraries can be viewed in the app (running locally) by going to `localhost:3000/{uuid}` where `uuid` is the `uuid` of the library. Get the `uuid` from the Library Registry admin.
+   * A Circulation Manager should be running at the same time and the library must be registered to this Library Registry (setting "Registry Stage" to "production").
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,52 +1,52 @@
...
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+The following environment variables can also be set to further configure the application.
+* Set `SHORTEN_URLS=false` to stop the app from removing common parts of the circulation manager URLs from the web app's URLs.
+* Set `CACHE_EXPIRATION_SECONDS` to control how often the app will check for changes to registry entries and circ manager authentication documents.
 
-   http://www.apache.org/licenses/LICENSE-2.0
+## Manager, Registry, and Application Configurations
+Any Circulation Manager you'll be using with the app also needs a configuration setting to turn on CORS headers. In the Circulation Manager interface, go to the Sitewide Settings section under System Configuration (`/admin/web/config/sitewideSettings`) and add a setting for "URL of the web catalog for patrons". For development, you can set this to "*", but for production it should be the real URL where you will run the catalog.
```
adriana-alter: As in, you need to do this step no matter which of the set-up methods you're using?  If so, let's put it at the beginning of the instructions.

EdwinGuzman: Good point.

README.md
```diff
@@ -1,52 +1,52 @@
...
-* To use a circulation manager, set `SIMPLIFIED_CATALOG_BASE`.
-   * Point this environment variable to the url of the Circulation Manager, for example `SIMPLIFIED_CATALOG_BASE=http://localhost:6500 npm run prod`. This will load the main library in the Circulation Manager in the app (running locally) by going to `localhost:3000`.
-* To use a configuration file, set `CONFIG_FILE` to point to a local file or a remote URL. Each line in the file should be a library's desired URL path in the web catalog and the library's circ manager URL, separated by a pipe character. For example:
+* `REGISTRY_BASE` - to use a Library Registry
+   * Example: `REGISTRY_BASE=localhost:7000 npm run prod`
+   * This is the default setting which will point to the Library Registry located at `localhost:7000`. The libraries can be viewed in the app (running locally) by going to `localhost:3000/{uuid}` where `uuid` is the `uuid` of the library. Get the `uuid` from the Library Registry admin.
+   * A Circulation Manager should be running at the same time and the library must be registered to this Library Registry (setting "Registry Stage" to "production").
```
adriana-alter: The Library Registry admin has `internal_urn`, not `uuid`.

adriana-alter:
Good news: this is a lot easier to read.

Bad news:

Screen Shot 2020-02-05 at 2 59 09 PM

Screen Shot 2020-02-05 at 3 27 48 PM

Screen Shot 2020-02-05 at 3 27 40 PM

README.md
EdwinGuzman: It's `urn:uuid:.....` but it should be updated. `internal_urn` is the variable name and that should probably be updated in the Library Registry so it has better names rather than snake case labels.

EdwinGuzman:
Yeah, but what steps are you doing to get those errors? All of them?

adriana-alter:
The first screenshot is the result of following the circ manager steps; the other two are from the library registry steps.  (Haven't tried the config file version yet, but not optimistic.)  I'm sure the problem is something I'm doing, since it works for you...but obviously the instructions are still really unclear!  Let's go through it in person when I'm back in the office; show me what you're doing?

README.md
```diff
@@ -1,52 +1,51 @@
...
-A Circulation Manager web interface for library patrons.
+A Circulation catalog web interface for library patrons.
 
-## Running the application
-There are three ways to run this application:
-* with a [library registry](https://github.com/NYPL-Simplified/library_registry)
-* with a single library on a [circulation manager](https://github.com/NYPL-Simplified/circulation)
-* with a configuration file for multiple circulation manager URLs
+## Installing
+Once you have a [Library Registry](https://github.com/NYPL-Simplified/library_registry) or [Circulation Manager](https://github.com/NYPL-Simplified/circulation), run `npm install` in this repository to install the dependencies.
```
aslagle: It's not necessary to have a library registry, unless something changed. In the previous list of three ways, only the first one required a library registry.

aslagle: Ah I see now it says "or" instead of "and" :)

kristojorg:
I definitely struggled with this when getting set up (I thought it was just me haha) and honestly haven't tried anything beyond the `CIRCULATION_MANAGER_BASE` setting. This will help me a bunch when I eventually go to make sure my changes work with all the various config options... so thank you!!

EdwinGuzman:
@adriana-alter made the minor changes.

README.md
```diff
@@ -1,52 +1,52 @@
...
```
kristojorg: It might be helpful here to explain the difference between a Library Registry and a Circulation Manager and how they interact. I'm still not totally clear on this myself.

- Do you always need a registry? How would an app get the registry data if not from a registry? How would it get the CM data if not from a CM? Why can you run with one _or_ the other?
- What data / services does each one provide?
- What is their relationship to real-world deployments? Is there one global registry generally? Does each library system run their own registry? Same question for CM.
- Does each instance of `circulation-patron-web` have its own CM? Does one CM manage an instance of `circulation-patron-web` as well as `OPDS-web-client`, `circulation-web` and the mobile apps? 
- Why would I run an app with multiple CMs in a config file? Is this for testing/dev purposes or is this a real life use case?

Sorry for all the questions, but I think this info would really help future devs like myself!

EdwinGuzman: Nope, no need to be sorry, they're good questions to ask. I'll include answers in the readme but some questions may be outside the scope of this repo, or at least very detailed explanations (and it's probably answered somewhere else already).

I think for the last question, @aslagle could provide more details:

> Why would I run an app with multiple CMs in a config file? Is this for testing/dev purposes or is this a real life use case?

aslagle: I think this was a request from Robert, or maybe Jonathan, or both. 

It's for real-life deployments, when an organization wants to use the same instance for multiple libraries but isn't running their own registry already. It's easier to configure a text file than run a whole registry when you don't need it for anything else.

EdwinGuzman: Thanks.

README.md
```diff
@@ -1,52 +1,58 @@
...
-## License
+### Application Startup Configurations
+There are three ways to run this application:
+* with a [Library Registry](https://github.com/NYPL-Simplified/library_registry)
+* with a single library on a [Circulation Manager](https://github.com/NYPL-Simplified/circulation)
+* with a configuration file for multiple Circulation Manager URLs
 
-```
-Copyright © 2015 The New York Public Library, Astor, Lenox, and Tilden Foundations
+By default, this application expects a Library Registry to be running at http://localhost:7000. For all three variations below, make sure that (1) the Circulation Manager is running at the same time and that (2) the library is registered to the Library Registry (setting "Registry Stage" to "production").
```
aslagle: (2) isn't necessary unless you're using the library registry option.

README.md
alsrlw: @kristojorg It was definitely me, I believe. Since we have multiple CMs deployed, and the web client service can easily serve multiple libraries, it is a way to cost-effectively use resources. It'll be a question in the future to see how many want to deploy a web client. So sharing one web client among potentially multiple CMs provides some flexibility.

kristojorg: I see I see. Would then the users be faced with urls like `domain/library/:libraryUrl/collection/:collectionUrl/...`? What I'm getting at is when you're in the `circulation-patron-web` interface, how are you meant to switch between libraries that are in the config file?

EdwinGuzman: The users are not meant to switch libraries _in_ the app. They would have to go to the base url directly. So `domain/library1`, `domain/library1`, etc.

EdwinGuzman:
If all is good I'll merge this in, but do need some approvals.

README.md
kristojorg: I see

PR #75, opened 2020-03-23
Version 1.0.0: Update opds-web-client and clean up download links a bit.

aslagle:
Most of this is updating opds-web-client to the latest version, which also required updating Typescript and Redux.

I also made a small change to make the download links less confusing - if there are multiple links with the same mime type, it will only show one, since patrons have no way to tell them apart. If there's one with an unknown type, it won't show at all, since patrons won't know how to open the file anyway.

The docker stuff is in a separate PR (https://github.com/NYPL-Simplified/circulation-patron-web/pull/74) which will hopefully be merged first, and then we can start tagging releases, so I made this 1.0.0. Kristo's stuff can be 2.0.0 once it's ready.


EdwinGuzman:
Version `1.0.0` usually means that that working version is production-ready and stable, but I don't think we've been strictly following this for semantic versioning. So is version 1.0.0 mostly for docker deployments?

aslagle:
Yeah, we didn't really pay much attention before. I am deploying this for production though, and I think it's reasonably stable.

PR #77, opened 2020-04-01
Remove server

kristojorg:
This is a draft PR to begin a discussion before I put in the work to do this. Ignore the changes, the `remove-server` branch is based on my `beta` branch, I haven't changed anything else in there yet.

### Reasons to remove the server
Right now the app is server-rendered, and I think there are a few compelling reasons to remove that functionality:

1. **Smaller codebase:** In general, minimizing the surface area of the code is an important goal for a project like this. Keeping the codebase simple will make it possible for a distributed team of engineers to maintain and improve it, especially when there are contractors coming and going. Removing the server would cut out a significant amount of error-prone code.
1. **Little added value:** Generally, server-rendering a react app is a performance optimization. In this case, it doesn't fetch all the required resources on the server, so it just shows the user a loading screen marginally faster. 
1. **Simpler deployments**: Right now we have to maintain a server for each CPW deployment. Both the code for that has to be maintained, and the devops takes work too. Without a server, we can simply deploy static files, which is far easier, faster, and less buggy.

### My proposal
I suggest we transition to codebase to use create-react-app. We would both remove the server, and most of the build configuration in favor of the more standard create-react-app config. It will allow us to do everything we need to do, but make the codebase much simpler and easier for new devs to become productive in. We would also benefit from the work the react team puts in to CRA, and would be able to upgrade seamlessly when they release new versions. The downside of this is we lose the ability to have custom configuration (unless we eject from CRA) - though I don't think we are doing anything currently that couldn't be done with a CRA-based project.

There are two alternative routes we could explore, however:
1. [Next.js](https://nextjs.org/): This is a framework for server-rendered react apps. I've used it many times, and it's quite good. The Zeit team does a great job maintaining it. If we wanted to keep the server rendering, this would be a really good choice. It also allows more configuration than CRA. While it would be significantly less complicated than our current setup, it's still slightly more than a CRA setup. It is also pretty proven tech, used by [many successful companies](https://nextjs.org/showcase) including Netflix, Hulu, Ticketmaster, etc. 
1. [Gatsby](https://www.gatsbyjs.org/docs/): This is a framework for (mostly) statically built websites, which our app falls into. I've used this many times as well, and it's fantastic. The basic premise is that it renders every page to a static html file at build time, which can then be served as static files. So it's actually faster than a server-rendered app, without the added devops complexity of deploying a server. The only pages that wouldn't be statically rendered are the dynamic ones (like the my-books page). It also has a really strong plugin system to enable any number of things from analytics to progressive-web-app functionality. The downside is that it would require a bit of extra work to connect the CM api to gatsby's graphql-based data layer, though once it's done, the dev experience is phenomenal.

### Questions
Am I missing something RE why the app was server rendered in the first place? James English mentioned that @aslagle might have chosen to server-render for security reasons? 

Let me know what you all think about this. I think moving to CRA or Next.js wouldn't be too much work and would help the app out a ton. Just want to make sure I hear your thoughts first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,27 +1,42 @@
 ## CHANGELOG
 
+### 3.0.0
+
+- removed server to render only with client
+
 ### 1.0.0
+
 #### Updated
+
 - Updated opds-web-client and Typescript to more recent versions.
 - Changed the download buttons to remove duplicate mime types and hide unknown mime types.
 
 #### Added
+
 - The application has a docker submodule and Docker images are now automatically built in Docker hub.
 
 ### 0.2.2
+
 #### Updated
+
 - Updated documentation explaining how to run the application. Added a separate LICENSE file.
 
 ### 0.2.1
+
 #### Added
+
 - `tslint-react-a11y` to review accessibility for React components.
 
 ### 0.2.0
+
 #### Updated
+
 - Updating to React 16.
 
 ### 0.1.0
+
 #### Updated
+
 - Updated packages include Webpack and Typescript.
 - Using `@types` instead of typings.
-- Updated unit tests with updated `fetch-mock` package. 
+- Updated unit tests with updated `fetch-mock` package.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-patron-web",
-  "version": "2.0.0",
+  "version": "3.0.0",
   "description": "patron UI for Circulation Manager",
   "repository": {
     "type": "git",


EdwinGuzman:
Can you point this PR to the right branch to only see the changes for this PR?

kristojorg:
I actually haven’t made any changes yet. I wanted to just get a discussion going. Maybe I should open an issue instead? Jonathan said he thought a draft PR with no changes would be the simplest, but I can move this to issues if that’s better

kristojorg:
Ok I updated the changelog so that I could base this on the beta branch. This way it's more clear there haven't been any changes yet.

aslagle:
I don't know what James was talking about. I feel like he must be thinking of opds-web-client, which has a demo server solely because I wanted it to use it with OPDS feeds that we don't control and can't add CORS headers to. That doesn't matter for this app, since it's only meant to be used with circ managers.

I actually did a static version for one of the two web catalogs I recently set up for DPLA, but I used opds-web-client rather than this. [It's here](https://github.com/dpla/biblioboard-web-catalog)

It was another developer who originally did the server-side rendering stuff - I was around then but I don't remember his reasoning. In my opinion the biggest advantage of having a server is that you can have much nicer URLs. I don't think it's really possible to do that with a purely static site, though even a simple Nginx or Apache server should be enough to get that.

You might also want to look at how circulation-web does it - it uses the circ manager as its server, but that's mainly just to handle the URLs, and also CSRF tokens. I don't know what the scope of your work is, but if you'll be doing more on this, it would be good for you to get familiar with that anyway since most changes here should be done there as well to keep things consistent.

The other issue is if you want to run this with a library registry, which no one is currently doing anyway afaik, having a server can reduce the load on the registry, which I've heard has recently had problems due to traffic from the mobile apps. There are probably better ways to deal with that though. And I suppose that also applies to the circ manager.

kristojorg:
Thanks for the details Amy. I expected as much RE the security reasons. I hadn't found one anyway. 

Regarding the URLs, in CPW at least I have full control over the URL scheme via react-router. Unless I'm missing what you're referring to, I don't think that should be an issue.

I'll take a look at circulation-web and at your static version, thanks for those tips. 

In terms of the library registry, I have a couple ideas there:
1. First, I was thinking we should audit how this is being used and remove any config options that aren't useful right now... that might include the library registry option as well as the option to run with multiple libraries under nested routes: `/lib1/book/...` and `/lib2/book/...` instead of simply `/book/...`. I was going to bring that up another time, but if it turns out we don't need the library registry type config, it might make sense to simply remove it.
2. The other option is to fetch the required config at build time, and then simply trigger a rebuild whenever something in the registry changes via a webhook. I would then provide the fetched registry info as a simple json file or the like to the web app.

aslagle:
I just meant you need a server to route dynamic paths to your HTML, to even load the react app in the first place. It can be extremely simple but I think there has to be something.

You won't learn much from auditing the current usage since it hasn't had much use at all yet - almost everything is only used in development at this point. Supporting multiple libraries somehow is definitely a requirement though. The library registry option makes sense for multiple libraries if you already have a library registry, but most people don't and wanted the config file so I added that too.

kristojorg:
Ah yes, I see what you mean. Can either use a very simple nginx server (there are standard docker images for these), or deploy the files to a CDN, which I presume AWS has the functionality to allow.

Sounds good RE the library registry and multiple library routes. I'll try to figure out how the registry could be queried once at build time to reduce the load there.

EdwinGuzman:
It doesn't matter now but maybe this would've worked better as an issue rather than a PR. We can keep any discussion here.

Thanks, @aslagle , for the details. Regarding the other update, I think Next.js might be the way to go if we want to keep supporting a server, multiple library routes, and the library registry.

Besides handling routes, the server right now is used to read environment variables and get data from the LibraryDataCache to set up the app accordingly. If you remove the server and set everything up statically, how would the app be built appropriately?

jonathangreen:
@kristojorg from the LYRASIS perspective supporting multiple libraries is necessary. In the deployment that have currently for testing we have 5 libraries, but given that we have circulation managers with hundreds of libraries that will be potentially using this we need to support a way to have multiple libraries.

kristojorg:
@EdwinGuzman I agree an issue would be a better place, but it looks like issues are disabled on this repo. I did just speak with Leonard however and it looks like we'll be moving these types of discussions to jira from now on. 

In terms of the points you brought up... I'm not against using Next.js, but I do tend to think it's a bit overkill for the project, especially since it does complicate the deployment process (you're deploying a server instead of just static files). More specific answers:
- The multi library routes are currently handed client-side. Doesn't depend on the server to make that happen.
- The LibraryDataCache, as far as I can tell, simply fetches and caches the auth document and registry base. I think fetching this data could be done at build time for a static app, which would mean that we wouldn't have to have the cache anymore since it's only ever being requested on build, which would be a minimal load on the registry.

Basically, I definitely don't think it's necessary to have a server at all, but if you all would rather we keep it, we can do that

kristojorg:
@jonathangreen I think I was unclear in what I was suggesting RE removing multi-library support. I wasn't suggesting removing the ability to have one CM serve multiple libraries and corresponding `circulation-patron-web` front ends. I was suggesting we move away from having one front-end deployment serve multiple libraries at the same time. This would just mean that each library has it's own circulation-patron-web deployment, even though they might share the same CM. 

If we move to static-files instead of server-rendering, creating these deployments becomes trivial, and then they could be hosted at whatever url you wanted, so you could maintain the setup of `http://basedomain.org/library1/` and `http://basedomain.org/library2/` or you could serve them at different domains: `http://library1catalog.org/` and `http://library2catalog.org`.

So I suppose what I should have said is "changing the way we support multiple library routes" rather than removing it outright.

jonathangreen:
I am concerned that this removes development complexity and pushes it over to ops complexity.

We have 100+ libraries, we would need to have automation to build deployments for each library, upload them to hosting, and then do that again every time there is an update to the web catalog that needs to be pushed out. As this scales out to 1000s of libraries, I'd like to see the deployment plan for this as it sounds on the surface to be harder to deploy and update then it is now. 

I don't mean to de-rail the discussion about server rendering, and it feels like the thread about supporting multiple libraries has maybe diverged enough that it should have its own discussion in a separate issue so not to distract from this one.

kristojorg:
Yep, agreed let's move it over to an issue in Jira once we get the whole scope of work discussion out of the way. I hear what you're saying, though, and that's good feedback. I think it would be less complex, but it won't be hard to setup a demo of what it would look like so we can know for sure once we refactor or remove the server code.

alsrlw:
Just to add a +1 to measure weight: the deployment factor will impact multiple hosting providers. In addition to the Lyrasis libraries Jonathan mentions, Amigos will also likely have a large numbers of libraries that are interested. Automation makes seeming big things small-ish, but it'd be good to see what the deployment implications are.

kristojorg:
For now, in terms of removing the server, it sounds like we need to decide if we would rather use static files (via possibly create-react-app) or have a server (via next.js).

The pro of having a server is that it (maybe) more directly maps to how we are querying and caching the registry and auth document right now. I say maybe because I would have to figure out how to implement the caching in next.js. I think it would be slightly different. Another pro would be marginally improved time to first render (which right now means improved time to first loading screen).

The downside of having a server is that it increases devops complexity significantly. Static, client-rendered sites can simply be deployed in an S3 bucket, which means zero server logs/errors, zero management of resources, lower cost, etc. 

The downside of client-rendering is that I'd have to update the LibraryDataCache to run at build time and save the information to a json file which is then included with the front end. 

I think it's pretty clear what I'm in favor of, but I'm just as happy to go with next.js honestly. It's a really great framework and easy to use.

alsrlw:
I have some questions related to the authentication_document caching and rebuilds: how frequently and how initialized. If I recall correctly, library-changeable factors (minimally, logo and client color scheme) are in the auth document. It sounds like any change like that for a library needs to result in a rebuild. If there will be a separate channel (Jira ticket?) for discussion of ops, I'll save the discussion  for that to be created.

PR #78, opened 2020-04-09
Docker

kristojorg:
This PR moves the docker configuration directly into the repository (instead of the previous submodule) and updates the README to reflect that. It also simplifies the `Dockerfile` a bit. The idea is that we will connect this repository to Docker Hub and set up automatic builds. 

@aslagle I know you previously added the docker submodule, but let me know if you have any issues with this route.

The updated release process for master would be:
1. Tag a release
1. Push the tag to github
1. Create a github release for the tag
1. Docker Hub will build a corresponding image
1. Consumers can then update their deployments to the new image


README.md
```diff
@@ -68,11 +68,57 @@ The following environment variables can also be set to further configure the app
...
+Whether running the container from a Docker Hub image, or a local one, you will need to provide at least one environment variable to specify the circulation manager backend, as described in [Application Startup Configurations](#Application-Startup-Configurations). You can also provide the other optional environment variables when running your docker container. There are two ways to run the container: (1) via the command line, and (2) via `docker-compose` with a `docker-compose.yml` file.
+
+When running the image with the `CONFIG_FILE` option, you will want to provide the file's directory to the container as a volume, so the container can access the file on your host machine. We do this in both examples below.
+
+### From the command line
+
+This command will download the image from NYPL's Docker Hub repo, and then run it with the `CONFIG_FILE` option and the name `patronweb`. If you would like to run your locally built image, subtitle `nypl/patron-web` with the name of the image you built previously (also `patronweb` in the example above).
+
+```
+docker run --rm --name patronweb -d \
```
EdwinGuzman, line 110: I get this error when running this `docker: Conflicting options: --restart and --rm.`

README.md
```diff
@@ -68,11 +68,57 @@ The following environment variables can also be set to further configure the app
...
+  --restart=unless-stopped \
+  -e "CONFIG_FILE=/config/cm-libraries.conf" \
+  -v "./config:/config" nypl/patron-web
```
EdwinGuzman, line 113: This also comes up `docker: Error response from daemon: create ./config: "./config" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.`.

README.md
kristojorg: Good point. This was actually copied over from `docker-patron-web`. I am going to remove the `--rm` option, since that would remove the container when it stops. I think, for production at least, the recommended way to run the container would be that it restarts if it exits with an error.

kristojorg:
Okay I've updated the README to include more instructions for `docker compose`, as well as updated the scripts to run the container properly. I also added documentation around what the commands were doing, and gave an example to show how to use a `LIBRARY_REGISTRY` or `SIMPLIFIED_CATALOG_BASE`. 

In terms of your second point, yes currently this is only for production deployment. We could edit this so that it can be used for development as well, but that would be a different dockerfile wherein we mount the local app files to the container. If we want to do this, I would recommend waiting until after the switch to Next.js. 

It is not assuming the production deployment will only use a CONFIG_FILE. You can use any of the three options. I tried to clarify that a bit in the README. I _think_ you could run the container with a local CM by simply binding a second port to the container, but I'm not sure, and I don't think that would be super useful unless we were using this for development.

EdwinGuzman:
Cool, the updated documentation looks good, thanks. I'm used to creating docker containers for both development and production instances, mounting the working directory and seeing updates through hot-reloading. It makes sense that this would only be used for production instances since that's the use case that is needed right now. It'd be nice to be able to run it in a development mode but it's not needed right now.

EdwinGuzman:
There are a lot of commits from this to `beta`. Does this PR need to be updated before it gets merged in?

kristojorg:
I'll rebase on beta and then we can probably use "squash and merge"

kristojorg:
Something got messed up with the git history here. Closing this in favor of https://github.com/NYPL-Simplified/circulation-patron-web/pull/80/

PR #81, opened 2020-04-22
CI/CD via github actions

kristojorg:
These workflow files will implement the CI/CD steps we discussed in our phone call. I think having the docker image building live in github makes more sense than having it live in Docker Hub, because it allows us to see what is going on with that right here in github. And since I was building the image and creating the release here, I figured might as well add the CI checks too, so everything is in one place / through one service.

Specifics...

### PRs
This will run a series of checks on all PRs. We will want to set up the repository so that PRs go to the development (beta) branch by default, and that both master and beta are protected from direct commits/pushes. The checks this runs on PR:
1. Lint
2. Typecheck
3. Run tests

### Beta branch
When a PR is merged into the beta branch, this will build a new docker image and push it to Docker Hub with the tags `beta` and `beta.short_sha` where short_sha is a reference to the merge commit into beta (like `beta.baa1ca5`). So people using the image can always install the most recent beta with the tag `beta` or a specific beta by using the `beta.short_sha`.

We can name the development branch `dev` or something other than `beta` if you like. 

### Master branch
When a PR is merged into the master branch, this will 
1. Create a new tag using the `package.json` version number.
2. Create a new release for that tag
3. Build the image and push it to Docker Hub with the tags `latest` and `2.10` or whatever the version number is.

The tag message and the release body will be the body of the pull request merge commit. This way we can discuss a release in a PR, come up with the release notes, then merge the PR with the release notes in the PR body. 

### Other things we can set up for the future
1. Slack notifications of releases or new beta builds (and failures)
1. Comment the published docker image on the PR for reference.
2. Automatic version number changes via PR label. One label of `major`, `minor`, `patch` or `no-version-change` would be required as a check on the PR.
3. Require that the changelog is updated as a PR check.

Once we merge this I will add docker secrets to the repo that will give push permission to `nypl/patron-web`, and we will have to configure the branch protections.

### DIFF
--- a/.github/workflows/publish-beta.yml
+++ b/.github/workflows/publish-beta.yml
@@ -26,6 +26,6 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           repository: nypl/patron-web
-          # tag the iamge with both beta and beta.short_sha
+          # tag the image with both beta and beta.short_sha
           tags: beta, beta.${{ steps.get_sha.outputs.short_sha }}
           add_git_labels: true
--- a/.github/workflows/publish-beta.yml
+++ b/.github/workflows/publish-beta.yml
@@ -1,0 +1,31 @@
+# This workflow will publish the beta image to docker hub
+name: Publish beta
+
+# action will only run when a pull request to the beta branch is closed.
+on:
+  pull_request:
+    branches: [beta]
+    types: [closed]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  publish_beta:
+    name: Publish beta image to Docker Hub
+    # only run if the PR is merged.
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get short SHA
+        id: get_sha
+        run: echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
+      - name: Publish to Docker Hub
+        id: publish_docker_hub
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: nypl/patron-web
+          # tag the iamge with both beta and beta.short_sha
+          tags: beta, beta.${{ steps.get_sha.outputs.short_sha }}
+          add_git_labels: true
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -1,0 +1,49 @@
+# This workflow will tag and release the master branch
+name: Production Release
+
+# workflow will only run when a pull request to the master branch is closed.
+on:
+  pull_request:
+    branches: [master]
+    types: [closed]
+
+jobs:
+  release_master:
+    name: Tag, Release and Publish master
+    # only run if PR is merged (not just closed)
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        # the tag will match the package.json version (eg. v1.0.0)
+      - name: Tag
+        id: autotagger
+        uses: butlerlogic/action-autotag@stable
+        with:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          strategy: package
+          prefix: v
+          # use the body of the PR commit as the tag message
+          tag_message: ${{ github.event.pull_request.body }}
+      - name: Release
+        id: create_release
+        if: steps.autotagger.outputs.tagname != ''
+        uses: actions/create-release@v1.0.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.autotagger.outputs.tagname }}
+          release_name: ${{ steps.autotagger.outputs.tagname }}
+          # use the body of the PR commit as the release body
+          body: ${{ github.event.pull_request.body }}
+          draft: false
+          prerelease: false
+
+      - name: Publish to Docker Hub
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: nypl/patron-web
+          tags: ${{steps.autotagger.outputs.version}}, latest
+          add_git_labels: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,60 @@
+# For all PRs, this workflow will
+#     - Install deps
+#     - Lint
+#     - Typecheck
+#     - Test
+name: CI
+
+on:
+  # will run on all PRs that are opened or updated (synchronized)
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    env:
+      CI: true
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 13.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 13.x
+      - name: Cache and install node modules
+        uses: bahmutov/npm-install@v1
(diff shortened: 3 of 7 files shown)


kristojorg:
I am going to make this a draft for now unless you'd like to include this in 2.0.0

kristojorg:
Note that you can see the PR workflow that ran on this pr because it included the workflow files under the passed checks.

kristojorg:
@leonardr I put you as a reviewer here since we discussed how this should work on the phone call. Let me know if you have any changes/input

leonardr:
I really like this technique; it keeps all the code and instrumentation in one place.

.github/workflows/publish-beta.yml
```diff
@@ -0,0 +1,31 @@
...
+        id: get_sha
+        run: echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
+      - name: Publish to Docker Hub
+        id: publish_docker_hub
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: nypl/patron-web
+          # tag the iamge with both beta and beta.short_sha
```
EdwinGuzman, line 29: small typo

.github/workflows/release-master.yml
```diff
@@ -0,0 +1,49 @@
+# This workflow will tag and release the master branch
+name: Production Release
+
+# workflow will only run when a pull request to the master branch is closed.
```
EdwinGuzman, line 4: "closed" is the same thing as merged?

.github/workflows/release-master.yml
```diff
@@ -0,0 +1,49 @@
...
+        # the tag will match the package.json version (eg. v1.0.0)
+      - name: Tag
+        id: autotagger
+        uses: butlerlogic/action-autotag@stable
+        with:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          strategy: package
+          prefix: v
+          # use the body of the PR commit as the tag message
+          tag_message: ${{ github.event.pull_request.body }}
```
EdwinGuzman, line 27: If there are multiple commits in the PR, does it take the last commit or are you expecting all the commits to be squashed?

EdwinGuzman:
From this, both Travis and Github will be doing similar building and linting/testing, right? Does it make sense to remove Travis then if Github actions is doing pretty much the same thing? Would it also make sense to let Travis build the image and deploy it to Docker Hub? (We have a similar workflow for deploying apps to AWS EB).

Is it really necessary to have the same workflow for a `beta` or `development` branch? If a new feature gets updated into `beta`, wouldn't you want to right away merge that into `master`? It seems fine but just an extra branch added to the workflow - unless if someone really wants to use the latest build but we won't release the actual work until later on (but I don't think we have this use case).

Besides that, this looks good but should have an updated version and readme.

.github/workflows/release-master.yml
kristojorg: Yep as far as github events go. This is same as on their webhook api

.github/workflows/release-master.yml
kristojorg: It's taking the body of the merge commit only, not the body of all the commits in the PR. So we would want to use the body of the merge commit as the place to write release notes

kristojorg:
That's correct that Travis becomes unnecessary with this. I wasn't aware that travis could build and deploy to docker hub. In my opinion, this is simpler because all of the output and tooling is right in github where the rest of the workflow lives, but if you'd rather configure Travis to build and deploy that's fine. I also like that github actions will show which job failed right in the PR, and I think we may want to expand on these workflows in the future to do things like check that a PR bumps the version and changelog.

In my opinion it is useful to have separate staging and production environments. This would allow for someone to make sure that the beta version works for them before those changes make their way to master. 

Ultimately, the workflow I would like would be to have a set of testing libraries / CMs with different configurations we can use, and then have not just a docker hub deployment for beta, but an actual live instance we can all look at and use to make sure our changes are working before releasing them to master. As this moves in to a production environment, I would be very nervous about deploying changes to production without having actually seen/used the live app with those changes.

kristojorg:
Yea I'm not crazy about the name `beta` for that branch either. I have seen it typically as `dev` because it's the default branch that development gets merged in to. I think we should look at what SimpleE is doing, because @leonardr said it's pretty similar to this. Maybe it makes sense to keep the names in sync between the two projects. In any case, whatever we choose for the `dev`/`beta`/`qa` branch, I'll update the workflow to have that as the title.

I'll also include some README info surrounding this.

And yes, to make a production release, you would make a PR to master, and we should protect master from allowing direct commits.

kristojorg:
@EdwinGuzman it looks like the iOS repo is using a branch called "develop". I think that the name `qa` for a branch might be somewhat confusing, because it seems like something separate from the development process - ie a branch that is specifically for the qa team or something. It wouldn't be a huge problem, but in my opinion `dev` or `canary` or `beta` or something is more descriptive. Either way, just let me know what you want to name the branch and I'll set it up and update the github actions to match.

EdwinGuzman:
Let's just stick to `beta` for now then. The mobile team and the web team don't have a unified workflow regarding git branch naming convention. For the majority of non-Library Simplified project at NYPL, the convention is `development`, `qa`, `master` since those are the environments being deployed to on AWS. It's just a context switch for me but it's not a general problem so `beta` is fine. As long as it's in the README for the next developer to learn about it, it's good to go.

kristojorg:
Okay, sounds good. It is in the README, so I think we should be ready to merge this. Things we'll need to do are 

1. Add the `DOCKER_USERNAME` and `DOCKER_PASSWORD` in github secrets
2. Re-create the beta branch
3. Add branch protections for beta and master (see screenshot below)
4. Make beta the default branch so that when you click "create PR", it goes to beta instead of master by default

EdwinGuzman:
Great. I did 1. and I'm guessing the rest need to be done after this is merged in?

kristojorg:
Yep I think that makes sense, that way beta starts out with the same history as master.

EdwinGuzman:
Sorry, I keep forgetting you don't have access to merge PRs in. We should change that.

EdwinGuzman:
All the steps are now complete. I guess it'll be tested with the next update and I'm excited to see that.
Thanks!

kristojorg:
Sweet! Looks like it created a release and successfully published to docker hub:
image

One thing though: the release seems to have pulled in the PR opening message not the body of your merge commit. That might be a mistake on my part in the script, but can I ask did you type anything in to the merge message when you make the merge? Maybe it used the opening commend as default if you didn't.

EdwinGuzman:
Hmm I'm not sure. I just pressed the merge PR button but didn't add a merge commit.

Also, the builds on docker hub are failing with :
```
Cloning into '.'...
Warning: Permanently added the RSA host key for IP address '....' to the list of known hosts.
Switched to a new branch '2.0.1'
Build path does not exist: ./docker
```

kristojorg:
That was the old way it was build, which I've now removed. Basically it used to work that we configure docker hub to auto-build the releases, but now instead github builds the image and then just pushes it to docker hub, so I removed the docker hub config and that error should be handled

EdwinGuzman:
That makes sense but those failing builds are for `master` and `2.0.1` just 12 hours ago. Should they be built again? to have a stable version?

kristojorg:
The version is out and stable, it was only the docker hub build that was failing, which we aren't using since we pushed `2.0.1` directly to docker hub from github actions. It was docker trying to build `2.0.1` itself that failed

EdwinGuzman:
I was confused but it's clearer now. So now we can just check the action logs and builds in github instead of the builds in docker hub, right? What if someone checks on the dockerhub builds only and sees those failing builds but don't know that the build is actually correct and in github?

This is minor and not a big deal but, I see it publishes under "edwinguzman" even though you did the work since it's using my credentials. It'd be nice to have the generic "nypl" username but I don't have credentials for that account.

leonardr:
nypl isn't a user on Docker Hub, it's an organization, so I don't think that will work. We have a machine user on GitHub, https://github.com/simplified-ci, who could be used for this purpose. Edwin, I can send you this user's password if you don't have it handy.

kristojorg:
We would need a machine user on docker hub, which would work. I disabled the autobuild on docker hub, so now it’s only building on github and we shouldn’t get any more errored builds

PR #83, opened 2020-05-06
add initial error page for server-side errors

M0nica:
This PR adds an initial error page to render for server-side errors. The page displays the statusCode, error title and error details when available.


src/pages/_error.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <NavButton sx={{ mt: 3 }} href="/">
+          Return Home
+        </NavButton>
+      </p>
+    </Layout>
+  );
+}
+
+Error.getInitialProps = ({ res, err }) => {
+  const statusCode = res ? res.statusCode : err ? err.statusCode : 404;
```
M0nica, line 24: The `circulation-manager` returns `status` for the errors that it returns but generally NextJS expects a `statusCode`. Should the shape be updated to expect either shape?

kristojorg:
Ok so I think this is a good start! I have some comments, and they mainly come from my own growing understanding of the types of errors we need to deal with. I hadn't had a chance to put in too much thought to those Jira tickets before you came onboard. So that's my bad. In any case, here is a summary of the types of errors in the app and what this PR should try to accomplish:

### Errors this PR should handle
1. Nextjs-thrown errors during server render. These should cause the custom error page to be shown with the status code and error message, similar to the default Next.js one. 
2. 404s from invalid path. This should return and HTTP 404 and show the 404 page you made eg if I visit `localhost:3000/hi` with a `SIMPLIFIED_CATALOG_BASE`. 
3. 404s when the `libraryDataCache` returns `null` or otherwise fails to get the prescribed library (this happens when using a `CONFIG_FILE` or `LIBRARY_REGISTRY`). This should return HTTP 404 and the custom 404 page you made.

I don't really know how to test # 1, but I think since the error page is in the right place we should be good to go there.

# 2 I don't get a 404 when I visit `localhost:3000/hi` with a `SIMPLIFIED_CATALOG_BASE`. Maybe you can check which page in the `/pages` directory is actually being rendered. It is the `/[library]/index` page, then we have a bit more work to do to figure this out. 

# 3 in order for you to work on this, you are going to need to run the application with a `CONFIG_FILE`. When run with that setting, multiple libraries are accessible at different url roots. So if my file looks like this:
```
brazoria|https://simplye.amigos.org/txbrazoria
xyzlib|http://simplye-dev-cm.amigos.org/xyzlib
```
Then `localhost:3000/xyzlib` will take me to the home page of `xyzlib` and corresponding for `/brazoria`. This is where those routes nested under `/[library]/...` come in to play. When using this setup, however, if you visit the `localhost:3000/`, you should get a `404`. Same with if you visit `localhost:3000/blah`. Currently, the `_app.getInitialProps` will think `blah` is the library you're trying to access, and will return `null` for the library data. This is the case we need to catch and properly return a 404 result. We can chat about this tomorrow, I know it might be a bit confusing (was for me initially).

I made an initial pass at handling this with the `is404` test in `_app`. That isn't really working correctly, however, so we need to update that and replace the `<Error>` component I render with your `Custom404`, and make sure the http response has the 404 status code. 

### Other Errors
There are two other types of errors I can think of, which I think we need to make a strategy for and some corresponding UI:

1. Errors thrown in the app client-side that are not caught by an `ErrorBoundary`. 
2. Errors from data fetching - these include the problem document. 

I'll handle coming up with a strategy for both of these in a separate PR, since you're still becoming familiar with the codebase. 

However, right now I have a top level `ErrorBoundary` in the `_app`. This is a carry-over from pre-nextjs. We shouldn't need it now, since Nextjs should capture unhandled errors in the `_error` page. We will want to make sure the `_error` page can show details of an error thrown client-side. Eg if you `throw new Error("Something broke!)"`, the error page should show "Something broke!". Right now it just says an error occurred.


In summary, the things left to do in this PR are:
1. Make sure 404s are working for both cases outlined above
2. Get rid of the top-level `ErrorBoundary`, and make sure the `_error` page you created can handle and display client-side errors. 
3. Handle the other code comments I'll leave above

src/pages/__tests__/404.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+  expect(node.getByText("404: Page Not Found")).toBeInTheDocument();
+});
+
+test("renders 'Return Home' link", () => {
+  const node = render(<Custom404 />);
+  expect(node.getByText("Return Home")).toBeInTheDocument();
+  expect((node.getByText("Return Home") as HTMLLinkElement).href).toBe(
+    "http://test-domain.com/"
+  );
+});
```
kristojorg, line 16: We can't have `__tests__` inside of the `/pages` directory, because that would make `localhost:3000/__tests__/...` viable routes. What we do instead is place the page component inside of `/components` and test it there, and then import it and render it simply inside of `/pages`.

src/pages/__tests__/_error.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    node.getByText("All licenses for this book are loaned out.")
+  ).toBeInTheDocument();
+});
+
+test("renders 'Return Home' link", () => {
+  const node = render(<Error />);
+  expect(node.getByText("Return Home")).toBeInTheDocument();
+  expect((node.getByText("Return Home") as HTMLLinkElement).href).toBe(
+    "http://test-domain.com/"
+  );
```
kristojorg, line 34: This works but just fyi you can also use `.toHaveAttribute("href", "http://...")`

src/pages/_app.tsx
```diff
@@ -32,12 +39,16 @@ type InitialData = {
...
+type MyAppProps = InitialData | NotFoundProps | ErrorProps;
 
 function is404(props: MyAppProps): props is NotFoundProps {
   return !!(props as NotFoundProps).statusCode;
 }
 
+function isError(props: MyAppProps): props is ErrorProps {
+  return !!(props as ErrorProps).status;
+}
+
```
kristojorg, line 51: I assume this is for handling errors that are created in `getInitialProps`. I think that errors thrown and not caught in `getInitialProps` will already render the `_error` page, so I think think this is necessary.

Source: https://github.com/zeit/next.js/issues/7127

src/pages/_error.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { NavButton } from "../components/Button";
+
+function Error({
+  statusCode,
+  title,
+  detail
+}: {
+  statusCode?: number;
+  title?: string;
+  detail?: string;
```
kristojorg, line 13: For the two cases this should handle (server-side error generated by nextjs, and uncaught client side error), I'm not sure if `title` and `details` will be present. You'll have to do some research to see what you'll be given in each of those cases. I presume that somehow you'd have access to the `Error` that was thrown and not caught on the client side.

src/pages/_error.tsx
```diff
@@ -0,0 +1,40 @@
...
+        </NavButton>
+      </p>
+    </Layout>
+  );
+}
+
+Error.getInitialProps = ({ res, err }) => {
+  const statusCode = res ? res.statusCode : err ? err.statusCode : 404;
+  return { statusCode };
+};
```
kristojorg, line 38: I'm not sure I fully understand this. Why do we need to do this?

M0nica: I believe this is required in order for NextJS to pass the `statusCode` to the Error page for server-side errors. It was included in their example Error pages.

kristojorg: Hmm I see. I know that if we were using `getInitialProps` inside of a regular page, we would have to call that pages `getInitialProps` inside of the `_app.getInitialProps`. So I guess I'm wondering if this ever actually gets called without modifying `_app.getInitialProps`

M0nica:
In the case of going to http://localhost:3000/ it is returning the default NextJS 404 page. When going to localhost:3000/blah it is throwing an error page with the stack trace which is only visible in development. My assumption is in production that would be replaced with the custom error page. The console is showing the error status as 404 in both cases.

kristojorg:
> In the case of going to http://localhost:3000/ it is returning the default NextJS 404 page. 

Is this when running with `CONFIG_FILE`?

> When going to localhost:3000/blah it is throwing an error page with the stack trace which is only visible in development. My assumption is in production that would be replaced with the custom error page. The console is showing the error status as 404 in both cases.

With `CONFIG_FILE` or `SIMPLIFIED_CATALOG_BASE`? Do you mind testing with a production build? `npm run build` then `npm run start`?

M0nica:
> > In the case of going to http://localhost:3000/ it is returning the default NextJS 404 page.
> 
> Is this when running with `CONFIG_FILE`?
> 
> > When going to localhost:3000/blah it is throwing an error page with the stack trace which is only visible in development. My assumption is in production that would be replaced with the custom error page. The console is showing the error status as 404 in both cases.
> 
> With `CONFIG_FILE` or `SIMPLIFIED_CATALOG_BASE`? Do you mind testing with a production build? `npm run build` then `npm run start`

That 404 appears on `/` and `/blah` (routes that didn't match a library) when using the CONFIG_FILE since there are multiple libraries and neither should be returned on the root. 

When using `SIMPLIFIED_CATALOG_BASE` instead of the `CONFIG_FILE` the 404 behavior is still off with the latest changes since for some reason it looks like `IS_MULTI_LIBRARY` seems to be `false` server-side and changes to `true` client-side so regardless of configuration `IS_MULTI_LIBRARY` seems to indicate it should support multiple libraries 🤔

src/components/MultiLibraryPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+import Error from "./Error";
+import Layout from "./Layout";
+
+const MultiLibraryPage: React.FC<{ showFormatFilter?: boolean }> = ({
+  children,
+  showFormatFilter = false
+}) => {
+  const errorStatusCode = 404;
+  return IS_MULTI_LIBRARY ? (
+    <Layout children={children} showFormatFilter={showFormatFilter} />
```
kristojorg, line 12: I'd rather keep these components separate from the `Layout` as they serve pretty different functions. So I think we should just render the children and move the `Layout` back out to the individual pages.

src/components/SingleLibraryPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const SingleLibraryPage: React.FC<{ showFormatFilter?: boolean }> = ({
+  children,
+  showFormatFilter = false
+}) => {
+  const errorStatusCode = 404;
+  return !IS_MULTI_LIBRARY ? (
+    <Layout children={children} showFormatFilter={showFormatFilter} />
+  ) : (
+    <Error statusCode={errorStatusCode} />
```
kristojorg, line 14: Is there a reason you're using the `Error` component instead of the `404` component? I actually think it is a good idea to combine them since they should have the same layout in the end. Does it make sense to you to get rid of `404` and rely on `Error` all the time?

kristojorg: If you were to do this, I would do something like what Next.js did in [their error page](https://github.com/zeit/next.js/blob/4a8b896122a58f07f57d64479293b8843d0a132b/packages/next/pages/_error.tsx#L5) and make a little map from status code to short default messages.

M0nica: I updated to use the map although it seems like NextJS recommends the 404 and 500 page being separate since the 500 page isn't statically optimized but the 404 page is. https://nextjs.org/docs/advanced-features/custom-error-page

kristojorg: Ah I just meant using the same component in `/components` for the `404` page and `_error` page in `/pages`, as well as anywhere else we need to programmatically render a 404/error page

src/pages/_error.tsx
kristojorg: Could you check that this gets called and/or is sending the proper status code from the server?

src/pages/404.tsx
```diff
@@ -0,0 +1,5 @@
+import NoMatch from "../components/404";
```
kristojorg, line 1: I think we should be importing the `Error` component and using that here. It will still be statically optimized because that is on the page level, not the component level

src/components/MultiLibraryPage.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react";
+import { IS_MULTI_LIBRARY } from "../utils/env";
+import NotFoundPage from "../pages/404";
+
+const MultiLibraryPage: React.FC<{}> = ({ children }) => {
+  const errorStatusCode = 404;
```
kristojorg, line 6: Is this used?

src/components/MultiLibraryPage.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react";
+import { IS_MULTI_LIBRARY } from "../utils/env";
+import NotFoundPage from "../pages/404";
```
kristojorg, line 3: Let's use the `Error` component here as well, and in `SingleLibraryPage`

M0nica:
https://github.com/NYPL-Simplified/circulation-patron-web/pull/83#discussion_r423118791
@kristojorg it didn't appear to be called. I ended up updating `_app.tsx`'s `getInitalProps` to ```if (!libraryData) {
    ctx.res.statusCode = 404;
    return { statusCode: 404 };
  }``` which resulted in a 404 page returning a 404 response instead of a 200 response. Do I need to look into any other client-side status odes or will they all be server-side?

kristojorg:
Okay that looks great! Much better than putting a `getInitialProps` in every page. So long as it's properly sending 404 status codes when using `SIMPLIFIED_CATALOG_BASE` and trying to access a multi library route and vice-versa, I think we're all good.

Let me know when you've made the other requested changes and I think we're ready to merge this

kristojorg:
Actually, if you're checking for the lack of library data and sending the 404 from `_app.tsx`, then do we even need the `MultiLibraryPage` and `SingleLibraryPage` components? I'm thinking it will catch the 404 in `_app` and render the 404 page then, in which case those guards will never be hit/necessary? What do you think?

PR #84, opened 2020-05-08
remove travis file

kristojorg:
The travis CI workflow is redundant now that we have the github actions version, so this removes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,0 @@
-language: node_js
-
-node_js:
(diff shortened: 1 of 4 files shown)


EdwinGuzman:
This should have a version bump and update in the readme, no? Or is the plan to just merge into `beta` and then bump changes when it will go into `master`?

kristojorg:
No you're right, just forgot

PR #85, opened 2020-05-15
Nextjs

kristojorg:
This PR is a refactor of the build system, server code, routing, and some other smaller bits to move the app over to Next.js. The goal here was to reduce code surface area/complexity, and to set us up for future improvements based on Next.js. Things to note:

1. We now use file-system based routing instead of react router. The routes are all the components within `/pages`. The one complication is that both multi-library routes and single-library routes are always available no matter which setting we run the app with. @M0nica added checks to make sure the app returned 404 in the case we try to access a multi library route when running with a single library, or vice versa. 
2. Links work slightly different with Next.js. [Read here](https://nextjs.org/docs/api-reference/next/link). The basics are that Next.js doesn't ship a route manifest like react router does. This means that we have to tell the link both where in the `/pages` directory to route, and what url to display in the browser. To make things simpler, I refactored the `pathFor` function into `LinkUtils`, which has three functions to help with build route for books and collections. We can potentially remove the `PathForProvider` when re-examining the relationship with opds-web-client.

In the future, we will be moving data fetching to take advantage of the Next.js APIs, which will greatly improve performance and continue to simplify the app.


EdwinGuzman:
I'm sort of swamped with my current task so this might take some time, if that's okay. I'll start reviewing over the weekend.

kristojorg:
@EdwinGuzman no problem. I'll just base future work off this branch

.gitignore
```diff
@@ -1,10 +1,35 @@
-lib/*
-dist/*
-node_modules/*
-typings/*
-*~
-npm-debug.log
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
```
EdwinGuzman, line 4: What's this for?

.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
```
EdwinGuzman, line 1: Did you mean to add this? If so, how would I use it in vscode?

package.json
```diff
@@ -9,84 +9,17 @@
...
+    "dev:axe": "REACT_AXE=true npm run dev",
+    "build": "next build",
+    "start": "next start",
+    "test": "jest --watch",
+    "test:ci": "jest --ci",
+    "lint": "eslint '*/**/*.{ts,tsx}'",
+    "lint:fix": "eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
+    "generate-icons": "npx @svgr/cli --template src/icons/_template.js --typescript --icon --ext tsx -d src/icons src/icons",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook"
```
EdwinGuzman, line 22: I don't think you meant to add these, right?

next-env.d.ts
```diff
@@ -0,0 +1,2 @@
+/// <reference types="next" />
```
EdwinGuzman, line 1: Is this for setting up the type for next to use typescript?

.gitignore
kristojorg: Not sure actually. This was generated by next.js originally. I'll remove it because I don't believe we need it.

.vscode/tasks.json
kristojorg: Nope, not sure how this got there. Thanks for catching, have removed it.

package.json
kristojorg: Ah, right we removed storybook in the beta PR. Thanks

next-env.d.ts
kristojorg: Yep exactly. This file has to be there.

src/components/Link.tsx
```diff
@@ -1,46 +1,70 @@
...
+ * removes consumed props and returns normalized props
+ */
+const buildLinkFromProps = (props: LinkProps, linkUtils: LinkUtils) => {
+  if ("bookUrl" in props) {
+    const { bookUrl, ...rest } = props;
+    return { ...linkUtils.buildBookLink(bookUrl), ...rest };
+  }
+  if ("collectionUrl" in props) {
+    const { collectionUrl, ...rest } = props;
+    return { ...linkUtils.buildCollectionLink(props.collectionUrl), ...rest };
```
EdwinGuzman, line 37: Minor, you don't need `props.collectionUrl` since you assigned it to a variable already.

src/components/__tests__/Button.test.tsx
```diff
@@ -28,13 +27,10 @@ test("disabled style", () => {
   expect(button).toMatchSnapshot();
 });
 test("NavButton renders correct element", () => {
-  const node = render(<NavButton to="/somewhere">child</NavButton>);
+  const node = render(<NavButton href="/somewhere">child</NavButton>);
   const button = node.getByText("child");
   expect(button).toHaveAttribute("href", "/somewhere");
   expect(button).toMatchSnapshot();
-  // make sure clicking it actually tries to route you home
```
EdwinGuzman, line 35: Is there an alternative for this test or why is there no need to test this anymore?

src/components/__tests__/Error.test.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { render } from "../../test-utils";
+import Error from "../Error";
+
+test("renders error message without 404 status code if status code not provided", () => {
```
EdwinGuzman, line 5: I think this should be `renders error message WITH 404 status code...`

src/components/__tests__/Error.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const node = render(<Error />);
+  expect(node.getByText("A 404 error occurred on server")).toBeInTheDocument();
+});
+
+test("renders 404 error message when there is a 404 error", () => {
+  const node = render(<Error statusCode={404} />);
+  expect(
+    node.getByText("Error: This page could not be found")
```
EdwinGuzman, line 13: This can also go in the previous test since it does render this by default.

src/dataflow/getLibraryData.ts
```diff
@@ -0,0 +1,153 @@
...
+    "No env vars found, setting registry base to http://localhost:7000"
+  );
+  return new LibraryDataCache(
+    "http://localhost:7000",
+    CACHE_EXPIRATION_SECONDS
+  );
+}
+
+// the cache should only be created once per server instance.
+const cachePromise = setupCache();
```
EdwinGuzman, line 81: Don't you need an `await` here?

src/components/__tests__/Button.test.tsx
kristojorg: Instead of testing that the click goes to the right place, we are testing that it has the right href, which does the same thing and is a more isolated test. I was running in to issues using next.js Link and jsdom which brought this up, and it became apparent that by testing the click I was testing the internals of these packages when what I really care about is the dom structure here.

src/dataflow/getLibraryData.ts
kristojorg: No because I'm actually trying to capture the promise here, not the final result yet. I await it later in `fetchLibraryData`, but wanted to create it at the top level because it should only ever be created once.

src/components/__tests__/Button.test.tsx
kristojorg: Just FYI testing the click functionality would still be possible if it were an actual button, but it's an internal link that just looks like a button which is why we test it using the href instead.

src/pages/book/[bookUrl].tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+import Layout from "../../components/Layout";
+import BookDetails from "../../components/bookDetails";
+
+const BookPage = () => {
+  return (
+    <Layout>
+      <BookDetails />
+    </Layout>
```
Comment (EdwinGuzman) [inline src/pages/book/[bookUrl].tsx:9]: Why do these components not have the status code check to render the error page?

README.md
```diff
@@ -76,6 +76,11 @@ The following environment variables can also be set to further configure the app
 - Set `SHORTEN_URLS=false` to stop the app from removing common parts of the circulation manager URLs from the web app's URLs.
```
EdwinGuzman, line 76: The readme currently says to run the production build as `npm run build:prod` but it should be `npm run build`.

EdwinGuzman:
Code looks good but here are some errors I see when running the app:

1) Using `REGISTRY_BASE=http://localhost:7000 npm run dev`:

The URL I go to is something like `http://localhost:3000/4d391d23-7bcc-45d3-9fcb-f7239a87a16c` where it's the uuid after the domain name (this is obtained from the library registry UI). I'm not sure why it's picking up the `cm_libraries` file even though I explicitly set the registry base. Everything still works but these errors are shown in the console. 

Screen Shot 2020-05-26 at 12 15 26 PM

Let's say I decide to click on a book:
Screen Shot 2020-05-26 at 12 18 08 PM

And then click on "Edwin" in the breadcrumbs, I get this error:
Screen Shot 2020-05-26 at 12 18 24 PM


2) Using `SIMPLIFIED_CATALOG_BASE=http://localhost:6500 npm run dev`:

Here it's also picking up the config file and saying it's running with the config file but it's actually not. I'm only able to access the "Edwin" library from the CM. So this runs correctly but the message is not right.
Screen Shot 2020-05-26 at 12 21 25 PM

Maybe it's that auth is not set up but when I click on "My books", I get direct to a page that says `You need to be signed in to view this page.` but there's no way to sign in.

It could be my configuration (I think this setup didn't work for me in the past anyway), but here's what I get for almost every book I click on (which wasn't a problem in the setup described in 1) ):

Screen Shot 2020-05-26 at 12 27 29 PM

3) Using `CONFIG_FILE=config_file.txt npm run`:

This starts up fine - 
Screen Shot 2020-05-26 at 12 29 15 PM

Using the _any_ library in the config file to test, I get this on the homepage. This includes the `xyzlib` and `brazoria` libraries, as well as local libraries in my CM. Actually, this error pops up when I view individual books, too.

Screen Shot 2020-05-26 at 12 29 04 PM

----
### Production build

I get this error when trying to run `npm run build` (even after removing `node_modules` and installing again):
Screen Shot 2020-05-26 at 12 47 39 PM

---- 
### Overall thoughts
I get the architecture change and I can't say I think it's any simpler just because I don't know Nextjs that well. I do think that clicking and transitioning is visibly slower. It's not the loading indicator but more so that when I click on a book or on the home button, there's a clear lag until anything seems to happen. This is also on the development instance and not the production build.

I'm hesitant about having the default jest test runner have `--watch` on. It's something that can be overlooked when deploying and forgetting to update the actions or Travis or whatever other CI tool is being used.

src/pages/book/[bookUrl].tsx
Comment (kristojorg) [reply src/pages/book/[bookUrl].tsx:9]: This of how the routing works in Next.js when there are two different routing modes that are laid on top of each other (single library and multi library). If it is running as a single library and you try to visit `/something`, the app will think you are trying to see a multi library page, and we need to check if the app is running in multi library mode or not in order to 404 you appropriately. This doesn't happen on the single library routes, so we don't need the check there.

I do hope in the future there can be a different solution for multiple libraries without adding tons of devops overhead. I have a few ideas. Hopefully one of them can remove all of these edge cases and extra code from having two separate routing modes

README.md
kristojorg: Thanks!

kristojorg:
Are 1+2 with the most recent code? Any way you can check after pulling? If you have a `.env` file or `.env.local` file set with eg a `CONFIG_FILE` and then add a `SIMPLIFIED_CATALOG_BASE` when running in the terminal, you'll end up having both set. At one point I had accidentally included a `CONFIG_FILE` setting in the `.env` file, which should be gone now.

For 1, I haven't been able to test at all with a library registry because I don't have one set up locally. Is there one deployed somewhere I can use?

If auth is not set up on your library you're testing with, the app isn't going to work right in general. With libraries I have auth set up in, the login modal does come up.

I fixed that `join` error - that was from having still not strict enough of typescript settings (we'll get there 😄 ) - aka I made a type error with an `any` type. 

On the production build - could you tell me what version of node you're using and possibly try once more with the most recent code?

**Overall thoughts**

The development instance does indeed have some lag when navigating. This is because it compiles pages lazily (when you request them), which apparently makes things like hot refresh etc faster and more stable. I find this also a bit annoying, but there is a little loading indicator in the bottom right, and after the first time you navigate to a page it should be pretty fast. It's basically just spreading out the long startup time we had before into chunks for each page. 

I think the ways that it'll be simpler haven't really been exposed yet. This just moves us over to Next.js without yet having taken advantage of really the core APIs it offers around server-rendering. That'll be down the line this summer. The two ways this specific PR does make things simpler is that it offloads the management of (a) the bundler, (b) the server, (c) the dev server/environment. It'll really start to shine later on though.

In terms of `jest --watch`, I did this because I would recommend always having it on while developing. I can definitely change it though if you prefer having `npm run jest` and `npm run jest:watch`

EdwinGuzman:
Re: `.env` file: I do have one but the environment variables are all commented out. I decided to run `npm run dev` with the env variable before it to make sure I knew what was running.

> For 1, I haven't been able to test at all with a library registry because I don't have one set up locally. Is there one deployed somewhere I can use?

I'm not sure there is one that is open.

I pulled the latest update and:
1. Using `REGISTRY_BASE=http://localhost:7000 npm run dev`:
same error as before

2. Using `SIMPLIFIED_CATALOG_BASE=http://localhost:6500 npm run dev`:
I'm not sure why my config is wrong. I get CORS errors even though my local CM should be open to any site. What I don't get is that the same "Edwin" library runs fine using step 1 from above, e.g. I cannot click on the ebook or audiobook icons using this step 2 but those buttons work fine using step 1.

3. Using `CONFIG_FILE=config_file.txt npm run`:
This is fine now.
The error I described in 1 also happens here (it was more apparent in _production_ mode) when I was on a book in the "Short Fiction" lane and then tried to click on "XYZ Public Library" in the breadcrumbs:
Screen Shot 2020-05-27 at 10 44 38 AM

This also showed up in the terminal when running in production mode but not sure how or why (maybe when I clicked on a book or the breadcrumbs):
Screen Shot 2020-05-27 at 10 47 09 AM

Notice the mismatch is happening in a missing slash in the second URL - `/xyzlib/book/URI/https:/standard ...`.

Production build:
I'm running it with node v10.19.0. This now works after the latest update.

kristojorg:
Okay I'll try to debug this this week. To summarize the errors, they are:
- LIBRARY_REGISTRY setting causes a url/as mismatch when clicking the breadcrumbs in your Edwin library
- eBook/audiobook icons can't be clicked on when running with SIMPLIFIED_CATALOG_BASE set to your local CM? Is that right?
- You get CORS errors using your local CM setup
- You get the same url mismatch in (more apparent in production) when running with CONFIG_FILE

Is this right? Sorry I got a bit confused by the description

EdwinGuzman:
Yes, that's right. Some more details:
* the LIBRARY_REGISTRY option breadcrumb issue happens with any library
* I think the second and third bullet points are related and probably just for my set up. @adriana-alter  if you have time and can test this PR branch, that'd be great.
* To be specific for the CORS issue, it doesn't happen in the CONFIG_FILE option, only the SIMPLIFIED_CATALOG_BASE option
* Yes.

adriana-alter:
Maybe it's my setup, but this has made it so I can't run the app at all 🙃

`REGISTRY_BASE` method produces this error (I tried it with several different libraries):
Screen Shot 2020-05-27 at 1 00 38 PM
 
`SIMPLIFIED_CATALOG_BASE` produces this error:
Screen Shot 2020-05-27 at 1 16 59 PM

`CONFIG_FILE` method doesn't run at all; just crashes in the terminal and seems to think the command is invalid.

kristojorg:
@adriana-alter have you removed `node_modules` and reinstalled? Could you tell me what version of node you are using?

@EdwinGuzman could you also tell me your node version? thanks

EdwinGuzman:
Node version is v10.19.0.

EdwinGuzman:
Wow, well the actions are pretty cool (in the latest commit):
Screen Shot 2020-05-29 at 9 46 22 AM

kristojorg:
Okay so here is what I have:
1. The href/as mismatch: I have fixed this error. It was a bug that sometimes we would get a collectionUrl from the backend which would just be the base url for the app, which I hadn't handled properly. I added a test for this in `Link.test.tsx`.
1. One of the mismatches you were experience - the one caused by a missing slash - I am not able to reproduce. Can you give me steps or something? Maybe even try it again now that I've fixed the first mismatch issue?
1. I am able to toggle with the audiobook/ebook filters in all of the libraries and configurations I have access to. Could you provide an example of a library where this doesn't work? How does it not work? Is there an error?
1. I'm also not able to reproduce the CORS issues. Is that specific to this PR?

EdwinGuzman:
All using the _same_ library:

1. Using `REGISTRY_BASE=http://localhost:7000 npm run dev`:
- Yes, the URL mismatch error is fixed now. 
- The message is the terminal about running with a library registry and a config file is also now fixed.

2. Using `SIMPLIFIED_CATALOG_BASE=http://localhost:6500 npm run dev`:
- The same library tested above returns 404 when attempting to view a book. It could be a CORS issue on my side but it's odd it happens in this config but not the `REGISTRY_BASE` config.
- The first error below is attempting to view a book.
- The next errors are when attempting to click on the audiobook/ebook tabs.

Screen Shot 2020-06-01 at 11 22 28 AM

3. Using `CONFIG_FILE=config_file.txt npm run dev`:
- This is working fine now with any library.
- The mismatch with the missing slash doesn't appear anymore.

------------------------------

- The CORS issue seems to just be my library.
- I noticed that when you build and run in production mode, the environment variables need to be set when building and not running. Can you make that more explicit in the readme? I should use the `.env` file but right now I'm just explicitly adding them before running any of the commands. Even when using an `.env` file, I think it makes sense to let the dev know when those environment variables will affect the app in the build process, i.e. if you run `CONFIG_FILE=config_file.txt npm run build` and then run `SIMPLIFIED_CATALOG_BASE=http://localhost:6500 npm run start`, the `SIMPLIFIED_CATALOG_BASE` environment variable will be ignored and only `CONFIG_FILE` will be used.

adriana-alter:
> @adriana-alter have you removed `node_modules` and reinstalled? Could you tell me what version of node you are using?

I did.  But, given that this is working for Edwin, I'm willing to believe that the issue is with my local setup rather than with your branch!  Node v11.12.0.

kristojorg:
- CORS Issue: This is interesting. I'm not sure what could be causing it to happen just with SIMPLIFIED_CATALOG_BASE. I'm also not sure what would cause that initial 404 error you got. Is this possibly because you should be running with `SIMPLIFIED_CATALOG_BASE=http://localhost:6500/edwin/groups` instead of just `localhost:6500`? Or, if that was just a typo, are you accidentally running it with an extra trailing slash in there?
- I will make a note about the env vars being set at build time in production

kristojorg:
Very odd, I'm not sure what I could have changed to surface that issue. I'll include it in the readme however. Where in the CM do you find that `/edwin/groups` url so I can point someone to look for the url of their specific library not the CM as a whole?

PR #86, opened 2020-06-09
Load embedded EPubs in webpub-viewer

M0nica:
This PR updates the NextJS app to load the `webpub-viewer` to render Exploded EPub files.

When a link to a specific `container.xml` file for an Exploded EPub is passed to the viewer as a query param it dynamically displays the rendered e-book. 

The app should redirect to the reader for a specific book when the book type is `"application/vnd.librarysimplified.axisnow+json"`  (or `link.type === "application/vnd.librarysimplified.web-epub"` this is currently commented out as it won't be a valid type in production but is required for the current dev test server for this feature).

To test: 

1: Run the app with `SIMPLIFIED_CATALOG_BASE={ENTER TEST OPDS FEED HERE} SHORTEN_URLS=false npm run dev` using the test OPDS feed.
2. Go to http://localhost:3000/
3. Click "plain text" (unecrypted) e-book which has type `link.type === "application/vnd.librarysimplified.web-epub"` (make sure that check isn't commented out in FulfillmentCard.tsx)
4. Click download link
5. Redirected to `webpub-viewer` with the desired e-book rendered (and updated URL)

Webpub-viewer loaded from c-p-w: Screen Shot 2020-07-15 at 3 25 02 PM

Note: The `webpub-viewer` currently only supports rendering unencrypted books. There is current working being done for `webpub-viewer` to support decrypted supported e-books.


src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import reader from "utils/reader";
+
+const BookPage = () => {
+  if (typeof window !== "undefined") {
+    reader();
+  }
+
+  return <div id="viewer" />;
```
Comment (M0nica) [inline src/pages/[library]/read/[bookUrl].tsx:9]: I updated so that the `reader.js` appends to the `viewer` div. I ended up having to copy a lot of files over directly from `webpub-viewer` as the single output build file I was using before was a feature of AMD  as `Only 'amd' and 'system' modules are supported alongside --outFile.`. I generated the files by changing the `webpub-viewer` tsconfig from `  "module": "amd",` to `  "module": "commonjs",` and removing `"outFile": "dist/webpub-viewer.js",`

src/pages/_app.tsx
```diff
@@ -22,6 +22,9 @@ import Head from "next/head";
 import Error from "../pages/_error";
 import { ParsedUrlQuery } from "querystring";
 import enableAxe from "utils/axe";
+import "../css/core.css";
```
M0nica, line 25: Need to scope this CSS to the reader only. Next was complaining that since it's global CSS it needed to be imported in _app.

src/utils/reader.ts
```diff
@@ -0,0 +1,114 @@
...
+    NightTheme,
+    ColumnsPaginatedBookView,
+    ScrollingBookView,
+    LocalAnnotator,
+    BookSettings
+  ) {
+    const element = document.getElementById("viewer");
+    const webpubManifestUrl = new URL(
+      /* this manifest url should be dynamic */
+      "/alice/manifest.json",
```
M0nica, line 34: This is where we point to the Alice manifest. The goal is to be just swap in remote manifests as needed in order to display different books in the e-reader.

M0nica:
Screen Shot 2020-06-10 at 7 07 34 PM
Just wrapping the e-reader in the circulation patron web layout resulted in a funky output. I still need to look into the design requirements for the e-reader in this app.

src/pages/_app.tsx
kristojorg: I'd look in to how Next.js handles css by default, because I believe there is a way to scope this that isn't too difficult

kristojorg:
Okay it needs some design massaging, but that's fine and this is great! I'm assuming the header of the reader is on top of the header of CPW because it's position: fixed ? I think all of that can be solved, we just need to check how this is meant to look with NYPL (also taking in to consideration the new design I'm working on now).

src/utils/reader.ts
```diff
@@ -0,0 +1,114 @@
+import SepiaTheme from "../../public/webpub-viewer/SepiaTheme";
```
kristojorg, line 1: I would put these in a folder within `/src`. We shouldn't need to keep this all in the public directory. This way also it will get compiled and optimized by Next

kristojorg: I'm referring to everything currently in `/public` btw

src/utils/reader.ts
```diff
@@ -0,0 +1,114 @@
...
+    SansFont,
+    DayTheme,
+    SepiaTheme,
+    NightTheme,
+    ColumnsPaginatedBookView,
+    ScrollingBookView,
+    LocalAnnotator,
+    BookSettings
+  ) {
+    const element = document.getElementById("viewer");
```
kristojorg, line 31: Ultimately it would be good to use a react ref to get the element here. Do that later though, once the other stuff is figured out.

src/utils/reader.ts
kristojorg: You should be able to move the manifest into `/src` and then look in to how Next imports json files. If I'm not wrong, you can get a url to the file by importing it. But yes ultimately this will need to be dynamic anyways, so until we have a strategy for generating them you can just leave it as is.

src/utils/reader.ts
```diff
@@ -0,0 +1,114 @@
...
+    const webpubManifestUrl = new URL(
+      /* this manifest url should be dynamic */
+      "/alice/manifest.json",
+      window.location.href
+    );
+    const store = new LocalStorageStore({
+      prefix: webpubManifestUrl.href
+    });
+    const cacher = new ServiceWorkerCacher({
```
kristojorg, line 40: I would be curious to know if this is needed anymore. If it's just a standard service worker, it probably isn't necessary since Next will provide most of that functionality already. 

If removing this would require heavy changes to the reader JS then we can leave it as is, but it would be nice ultimately to keep stripping away this kind of stuff so it's running closer and closer to just a bare react component inside Next.js. 

I suppose it partially depends on how NYPL is going to support this. IE if they just want to support the webpub-viewer repo and expect this app to use it as is, or if they're happy to have this app customize it some and support it directly here. We can discuss on the call today

src/utils/reader.ts
```diff
@@ -0,0 +1,114 @@
...
+import DayTheme from "../../public/webpub-viewer/DayTheme";
+import NightTheme from "../../public/webpub-viewer/NightTheme";
+import BookSettings from "../../public/webpub-viewer/BookSettings";
+import LocalAnnotator from "../../public/webpub-viewer/LocalAnnotator";
+import ServiceWorkerCacher from "../../public/webpub-viewer/ServiceWorkerCacher";
+import ColumnsPaginatedBookView from "../../public/webpub-viewer/ColumnsPaginatedBookView";
+import ScrollingBookView from "../../public/webpub-viewer/ScrollingBookView";
+// ( after
+export default function () {
+  (function (
```
kristojorg, line 16: For simplicity sake, I would unwrap this so that it's not a closure. I can't really think of why it needs to be one anymore, so I would split out this function, and then call it with the required args from within the default function. I'd also give the default function a name, like `initReader` or something

src/utils/reader.ts
```diff
@@ -0,0 +1,114 @@
...
+      ariaLabel: "Go back to the Github repository"
+    };
+    BookSettings.create({
+      store: settingsStore,
+      bookFonts: [publisher, serif, sans],
+      fontSizesInPixels: fontSizes,
+      //      FontSizeInPixels: FontSize,
+      bookThemes: [day, sepia, night],
+      bookViews: [paginator, scroller]
+    }).then(function (settings) {
```
kristojorg, line 79: Try using async/await here if possible

src/pages/[library]/read/[bookUrl].tsx
Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:9]: Yea eventually we will either need to change how webpub viewer is built so we can import it, or move it directly into this repo if it's only to be used here.

src/utils/reader.ts
kristojorg: Do you know how we will generate these manifest files eventually? Does the webpub viewer do that?

M0nica: The backend should be returning the manifest links. So eventually they’ll be external HTTPS URLs.

hokei:
@kristojorg 
`Which one we choose largely depends on how webpub-viewer is currently being used and how NYPL plans to use it in the future. Looking in to that code, there does seem to be quite a bit that could be done to make it simpler and easier to integrate into an existing app.` 

I'm interested in what  you think the steps might be to do this.  I was always operating under the assumption that to make it into a simpler app would pretty much require us to re-write it from ground up, because of 

1) the lack of current framework means that the code is not written in such a way that it would allow for easily being able to break it up into components.  
2) The fact that the styles are all injected, at different points in the app, means that they're pretty brittle and I assumed it would be a good chunk of work to separate them out and figure out how they're interacting / what's overriding what.  

Have I been overthinking this?  Or is there an intermediate step that can get us to componentization/injection that doesn't require us to re-write the entire app?

hokei:
> So, we need some information from NYPL on what the plans are for this, as well as what the design requirements are and then we can decide how to move forward here.

There's a hope that this can be "our webreader" going forward.  That means, we can continue to build on top of it, perhaps add pdf functionality, bookmarking, other features (These aren't planned, just the kind of things that NYPL /might/ do /sometime/ in the future).  But for this product, I think it only needs to work as it is now, so if there's refactoring that we could do to make both this work easier and also get us to a place where it's possible to add new functionality, that would be amazing!  But if there's a way for us to get the webreader as it is, into CPW, we'll take that too.

src/utils/reader.ts
M0nica: Yeah, we can look into using `ref` if/when `reader.ts` is converted to React. Right now it doesn't have access to the virtual DOM.

kristojorg:
So yea I agree that if you want to start making significant upgrades to to the viewer, it will probably be easier to port it to React, although it may be possible to do that incrementally. I think it's actually a relatively simple app, it just looks somewhat complicated because of how it is built. I might be missing the complexity somewhere in there though. 

I think let's make the changes that are required to get it integrated into CPW within a react component, and then we can figure out afterwards what upgrades would want to be made and how it makes sense to do those. My guess is that yes it will require a major refactoring if you want to get it to a place where it can be easily upgraded

src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -0,0 +1,976 @@
...
+  const bookManifestUrl = `/${bookUrl}/manifest.json`;
+
+  if (typeof window !== "undefined") {
+    reader(bookManifestUrl);
+  }
+
+  return (
+    <>
+      <div id="viewer" />{" "}
+      <style global jsx>{`
```
Comment (M0nica) [inline src/pages/[library]/read/[bookUrl].tsx:19]: It seems like to get these styles into the page level component they needed to be added like this. I've had to copy/paste the styles in /read/[bookUrl] and /[library]/read/[bookUrl]. I set them as global since we don't actually have access to the DOM that injected by the viewer to use classes to make the css modular.

Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:19]: Does this not alter the global `:root` `body` and other elements styling? I know the `<style jsx>` tag usually isolates styles to a component, so does the `global` prop override that?

Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:19]: Also, what made it necessary to copy paste them here instead of importing the css file itself?

Comment (M0nica) [reply src/pages/[library]/read/[bookUrl].tsx:19]: If they're imported they need to be imported into `_app` or else the global CSS error is thrown https://github.com/vercel/next.js/blob/master/errors/css-global.md#global-css-must-be-in-your-custom-app.

Setting `global` within  vercel/styled-jsx doesn't throw that warning. I'm not sure how global those styles are. If I remove the ` <style global jsx>`  from one of the pages then the other one is unstyled, so I'd have to see if those styles are affecting other components or not since they aren't being applied to the other viewer page.

Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:19]: I think that error being thrown is because some of the css is global (ie affects root tag, body tags, etc). The error is correctly pointing out that if we import (or otherwise load) this css, it will affect the whole app when it is loaded. 

What we want is css that will _not_ affect the whole app, but is scoped to the webpub viewer component and its children. This will likely require changing some of the css in that file eg from
```css
h1, h2 {
  fontFamily: serif;
}
```
to 
```css
.reader> h1,h2 {
  fontFamily: serif;
}
```

I'm not sure if that's the exact syntax, but the idea is to scope the css to only beneath one class or id or something (the div the reader is rendered into).

I would also suggest reading over the [NextJS CSS Docs](https://nextjs.org/docs/basic-features/built-in-css-support) to see how they support scoped css imports.

Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:19]: It may be that if you import the file correctly (it looks like it needs to use `.module.css` extension), then NextJS will scope even the global css rules in there properly to the component. Or, you may have to change the import and also change the global styles so NextJS can properly scope it.

M0nica:
This is currently blocked by updates that are being made to webpub-viewer (https://github.com/NYPL-Simplified/webpub-viewer/tree/SFR-develop) to support statically rendering remote EPub files without CORs issues. I switching this to a draft and will rebase and update once it's can be integrated with the version of webpub-viewer that supports statically rendering remote EPub files without CORs issues.

src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -0,0 +1,978 @@
...
+  const library = useLibraryContext();
+  const router = useRouter();
+  const { bookUrl } = router.query;
+  const bookManifestUrl = `${bookUrl}/META-INF/container.xml`;
+
+  const { catalogName } = library;
+
+  if (typeof window !== "undefined") {
+    reader(bookManifestUrl, catalogName);
+  }
```
Comment (kristojorg) [inline src/pages/[library]/read/[bookUrl].tsx:16]: I'm not certain, but it seems like this would need to be done inside a react `useEffect` hook otherwise it will be called every time the component re-renders, whereas we really only need it to be called on mount.

tsconfig.json
```diff
@@ -22,15 +18,16 @@
...
+  "exclude": ["node_modules", "config/srv"],
   "include": [
     "next-env.d.ts",
     "**/*.ts",
     ".eslintrc.js",
     "**/*.tsx",
     "next.config.js",
+    "src/utils/require.js",
+    "src/utils/fetch.js",
+    "src/webpub-viewer"
```
kristojorg, line 31: Isn't everything in `src/` included by default via the `"**/*.tsx"` glob?

kristojorg: I'm guessing this is left over from before bc I don't see these files

tsconfig.json
```diff
@@ -22,15 +18,16 @@
     "importHelpers": true,
     "baseUrl": "src"
   },
-  "exclude": [
-    "node_modules",
-    "config/srv",
-  ],
+  "excludedFiles": "*.test.js",
```
kristojorg, line 21: I'm not sure we should do this. I rely on typescript quite heavily in test files

src/pages/read/[bookUrl].tsx
```diff
@@ -0,0 +1,978 @@
+import React from "react";
```
Comment (kristojorg) [inline src/pages/read/[bookUrl].tsx:1]: If this file is the same as `/pages/[library]/read/[bookUrl]` can we just have one component that is rendered in both?

src/utils/reader.ts
```diff
@@ -0,0 +1,111 @@
...
+            rootfile || ""
+          );
+
+          return resolve(rootfile ? new URL(url) : webpubBookUrl);
+        })
+        .catch(err => reject(err));
+    });
+  }
+
+  webpubManifestUrl().then(url => initBookSettings(element, url, catalogName));
```
kristojorg, line 46: Is there a reason you need to declare this function before calling it? I think this might be simpler to read:
```
const response = await fetch(containerHref);
const text = await response.text();
const xml = await new window.DOMParser().parseFromString(text, "text/html");
const rootfile = xml.getElementsByTagName("rootfile")[0]
  ? xml.getElementsByTagName("rootfile")[0].getAttribute("full-path")
  : "";
const url = containerHref.replace("META-INF/container.xml", rootfile || "");

const finalUrl = rootfile ? new URL(url) : webpubBookUrl;

initBookSettings(element, url, catalogName);
```

tsconfig.json
M0nica: this was only for .js* but I will remove since I don't think there are any test.js files 🤔

src/pages/[library]/read/[bookUrl].tsx
Comment (M0nica) [reply src/pages/[library]/read/[bookUrl].tsx]: I was running into Next specific issues when trying to import the CSS separately since it seems like out of the box Next only supports separate CSS file that's imported into _app which is not what we want. https://nextjs.org/docs/basic-features/built-in-css-support. I don't recall the error but there was an error when I tried to import the styles from a separate file.

.gitignore
```diff
@@ -31,4 +31,6 @@ yalc.lock
 wallaby.js
 config/
 .vscode/
-.vercel
\ No newline at end of file
+lib/
```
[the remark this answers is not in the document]
M0nica: I still had some old `lib/` files around but it seems like they are no longer necessary so updated the `.gitignore` accordingly

src/pages/[library]/read/[bookUrl].tsx
Comment (M0nica) [reply src/pages/[library]/read/[bookUrl].tsx]: Good call! I updated to use `useEffect()` to avoid unecessary calls

src/utils/reader.ts
M0nica: I updated the syntax to look more-so like you suggested but was running into some `Promise` related issues when trying to avoid calling the function after it was declared 🤔

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -257,9 +257,29 @@ const DownloadCard: React.FC<{
...
+
+              if (
+                /* web-epub is currently used in test-server.
+                 to-do: remove the below commented out check */
+                // link.type === "application/vnd.librarysimplified.web-epub" ||
+                link.type === "application/vnd.librarysimplified.axisnow+json"
+              ) {
+                hasReaderLink = true;
+                link.url = `/read/${encodeURIComponent(link.url)}`;
+              }
```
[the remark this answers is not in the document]
M0nica: I can update `useDownloadButton` in `opds-web-client` to return "Read Online" instead of "Download" for resources that will be rendered in the `webpub-viewer` (based on type). However, I have mixed feelings about hard-coding the `/read/` part of the url  into `opds-web-client` as if we decide to change the route in Next then we will also have to update `opds-web-client`. I can move the route logic there if we are okay with that trade-off, since c-p-w is the only consumer that cares about       `link.type === "application/vnd.librarysimplified.web-epub" ||
      link.type === "application/vnd.librarysimplified.axisnow+json"` at the moment.

src/utils/reader.ts
EdwinGuzman: This works for me but the last line should be `initBookSettings(element, finalUrl, catalogName);`

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -272,16 +274,42 @@ const DownloadButton: React.FC<{
...
+
+  let hasReaderLink = false;
+
+  if (
+    /* web-epub is currently used in test-server.
+                 to-do: remove the below commented out check */
+    // link.type === "application/vnd.librarysimplified.web-epub" ||
+    link.type === "application/vnd.librarysimplified.axisnow+json"
+  ) {
+    hasReaderLink = true;
```
EdwinGuzman, line 286: I think you can eventually refactor this so it doesn't use `hasReaderLink` and just directly returns the `NavButton` here but it's not a big deal.

src/components/WebpubViewer.tsx
```diff
@@ -0,0 +1,980 @@
...
+          color: #5b5852;
+          padding: 0.35rem 0.35rem 0.2rem;
+          margin: 0;
+          -webkit-appearance: none;
+        }
+
+        a {
+          color: #5b5852;
+          text-decoration: none;
+        }
```
kristojorg, line 54: I would still rather this be imported and not global if possible? Are you seeing an error when trying to do that?

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -495,6 +495,29 @@ describe("available to download", () => {
...
+    availability: {
+      status: "available",
+      until: "2020-06-18"
+    }
+  });
+
+  test("constructs link to viewer for OpenAxis Books", () => {
+    const utils = render(<FulfillmentCard book={viewableAxisNowBook} />);
+    const readerLink = utils.getByRole("link", {
+      name: /download/i
```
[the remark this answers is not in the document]
M0nica: I just bumped the opds-web-client version to 0.4.7 so now it will says "Read Online".

src/components/WebpubViewer.tsx
M0nica: For the CSS it seems the only way to get the Reader to render as expected is by putting the styles in-line with the reader component. I am not importing them as global styles in the app.tsx because the styles should only affect the reader. 

I tried importing the styles directly from the webpub-viewer NPM package but since it is using generic selectors like :root it also threw the "Syntax error: Selector ":root" is not pure (pure selectors must contain at least one local class or id)" error. So it seems like right now importing CSS is not an option.

I also tried using CSS modules but the styles in the viewer ended up appearing broken (some styles applied and some didn't) and was not usable.

kristojorg: Can you give me a `SIMPLIFIED_CATALOG_BASE` to run this with so I can try and debug?

kristojorg:
I also think we should rename the `DownloadCard` to `AccessCard` or something similar since now it is both for downloading and reading online

kristojorg:
After debugging the style issues locally, I can see that it won't be simple to properly encapsulate the styles. At the moment the strategy only works because the viewer is fullscreen, so you don't see how these styles are changing the rest of the application (and they get removed on unmount). 

In my opinion, this method of integrating the webpub-viewer isn't a great long-term solution because it will be difficult for future developers to understand and it can't really be changed / upgraded. I also don't know how changing styles within CPW will affect the webpub viewer and if it does break something, I think fixing it won't be easy. IMO, the three best options for integrating this with long term support are: 

1. Encapsulate styles and everything into webpub-viewer. Make sure styles are properly scoped to just that package, and then render it as we are now within a react component.
2. Rebuild the webpub-viewer with react. Especially if it's only meant for use here in CPW. I understand this probably would take too long right now.
3. We could make the `/reader` route an API route that serves HTML. This would separate it from the rest of the application as moving between `/reader` and another route would require a full navigation refresh. 

All that said, I know that putting more work on this might not be feasible. I'm fine approving if @EdwinGuzman and @leonardr are comfortable with the tradeoffs of having a section of the app be pretty fragile and difficult to maintain.

M0nica:
> Couple more things:
> 
>     1. I think we should rename the `DownloadCard` to `AccessCard` or something similar since now it is both for downloading and reading online
> 
>     2. Is there any way to get out of the viewer once it is entered? Would be nice to have a close or a back button.
> 
> 
>  It looks like the color theme css isn't quite working?
1. I agree it makes sense to rename the `DownloadCard`
2. Screen Shot 2020-07-21 at 12 16 18 PM

To allow users to easily exit the viewer, we can have some copy like BACK TO LIBRARY NAME in the top navigation that when clicked would take the user back to the previous page/book details page. Digital Research Books currently uses that UI/UX to allow readers to exit the viewer (see above).

I will look into the issues with the styles not being applied properly.

leonardr:
After talking with Crystal I'm OK with these tradeoffs. We know that we have plenty of work still to do on webpub-viewer.

hokei:
> Is there any way to get out of the viewer once it is entered? Would be nice to have a close or a back button.
https://github.com/NYPL-Simplified/webpub-viewer/issues/98 😆 
But yeah, what Monica posted is the way we would do that now.

hokei:
Yeah this works for me!  I'll leave it to Kristo to give final sign-off

kristojorg:
@M0nica Sounds good, but I'm not seeing the back button you're referring to from the DRB screenshot. Or are you saying you'll add a back button?

M0nica:
> @M0nica Sounds good, but I'm not seeing the back button you're referring to from the DRB screenshot. Or are you saying you'll add a back button?

I created a separate PR (off of this branch) to make the back link clickable, it did require a small change to webpub-viewer that went out with `0.1.3`  https://github.com/NYPL-Simplified/circulation-patron-web/pull/100

kristojorg:
Okay nice thank you. I think with that back button, and the style fixes and the renaming of `DownloadCard` we should be ready to merge this : )

src/components/WebpubViewer.tsx
```diff
@@ -0,0 +1,986 @@
+import React from "react";
+import reader from "utils/reader";
+import { useRouter } from "next/router";
+import useLibraryContext from "./context/LibraryContext";
+
+const BookPage = () => {
+  const library = useLibraryContext();
+  const router = useRouter();
+  const { bookUrl } = router.query;
+  const bookManifestUrl = `${bookUrl}/META-INF/container.xml`;
```
hokei, line 10: Is this always guaranteed?  Are they always going to be container xmls and never content opfs?

M0nica: I think for now they'll always be `container.xml`s but that's a good point we should make it work for xmls and opfs as well

M0nica:
Currently the ci-check is failing due to an error in a file that wasn't changed in this PR 
image

kristojorg:
@M0nica you did update some of the `@types` dependencies however. I would try going back to the old dependencies and see if that fixes it. If so, then it's a change there that caused this.

kristojorg:
Also, see here: https://github.com/reduxjs/redux/issues/3466

You can try updating react, redux etc and their types to the latest and see if that fixes it as well

M0nica:
It seems like all of the checks are passing now, I had to update the syntax in some of the tests and try different versions of the libraries. I still have some tests that fail locally because the date is off by 1 on my computer but it passes in CI. Is there a specific timezone we should set in the test env?

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -156,8 +156,8 @@ describe("available to borrow", () => {
       })
     });
 
-    const borrowButton = await utils.findByRole("button", {
-      name: /Borrowing.../
+    const borrowButton = await utils.getByRole("button", {
+      name: /Borrow/i
```
kristojorg, line 159: I'm curious about these changes. I think we should still have this borrowing state and still be testing for it

kristojorg:
I'm not sure there is a specific timezone. We should probably just mock out the Date object so that it's returning something standard no matter what environment or computer we run this on

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -272,16 +276,39 @@ const DownloadButton: React.FC<{
...
-    >
-      {downloadLabel}
-    </Button>
-  );
+
+  /* web-epub is currently used in test-server.
+                 to-do: remove the below commented out check */
+  const hasReaderLink =
+    // link.type === "application/vnd.librarysimplified.web-epub" ||
+    link.type === "application/vnd.librarysimplified.axisnow+json";
```
kristojorg, line 284: Is this supposed to be here still or no?

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -272,16 +276,39 @@ const DownloadButton: React.FC<{
...
+
+  if (hasReaderLink) {
+    link.url = `/read/${encodeURIComponent(link.url)}`;
```
kristojorg, line 287: I think it would be safer to create a new variable here instead of modifying `link.url`

src/components/bookDetails/ReportProblem.tsx
```diff
@@ -12,19 +12,22 @@ import { ComplaintData } from "../../interfaces";
...
 import { H1 } from "components/Text";
+import { getReportUrl } from "utils/libraryLinks";
 
 const getDisplayType = (type: string) =>
   type
     .replace("http://librarysimplified.org/terms/problem/", "")
     .replace(/-/g, " ")
     .split(" ")
-    .map(t => t[0].toUpperCase() + t.slice(1))
+    .map(t => (t ? t[0].toUpperCase() + t.slice(1) : ""))
```
kristojorg, line 22: Is this necessary? Isn't `t` always of type string?

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -155,8 +155,8 @@ describe("available to borrow", () => {
       })
     });
 
-    const borrowButton = await utils.findByRole("button", {
-      name: "Borrowing..."
+    const borrowButton = await utils.getByRole("button", {
+      name: /Borrow/i
```
kristojorg, line 159: Why did this need to be changed? These tests are meant to test the loading state when the button should say "Borrowing..." instead of "Borrow". Also, using `await` with `getByRole` won't do anything. 

I think this is a case where the test was correctly failing and we need to fix the code instead of changing the test if possible

src/utils/axe.ts
```diff
@@ -7,5 +7,5 @@ import * as React from "react";
 export default async function enableAxe() {
   const ReactDOM = await import("react-dom");
   const axe = await import("react-axe");
-  axe(React, ReactDOM, 1000);
+  axe.default(React, ReactDOM, 1000);
```
kristojorg, line 10: Have you made sure that running with AXE still works?

src/components/bookDetails/FulfillmentCard.tsx
M0nica: We don't have a test server with AxisNow documents so for now it would be helpful to more easily test with the .web-epub which isn't supported in the `opds-web-client` types.

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -107,8 +107,8 @@ describe("available to borrow book", () => {
     userEvent.click(utils.getByText("Borrow"));
     expect(updateBookSpy).toHaveBeenCalledTimes(1);
     expect(updateBookSpy).toHaveBeenCalledWith("borrow url");
-    const borrowButton = await utils.findByRole("button", {
-      name: "Borrowing..."
+    const borrowButton = await utils.getByRole("button", {
+      name: /Borrowing.../i
```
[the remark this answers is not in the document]
M0nica: I don't think the non-test changes I made touched BookListItem so I was assuming there was an issue with the version of React testing library. doing something like `title: 'Borrowing'` passed the tests but failed the types, `name: 'Borrowing...'` failed the test but `name: /Borrowing.../i`didn't. 

If you have a button like  `<button type="submit">Submit</button>` then the [testing playground](https://testing-playground.com/ ) recommends ```screen.getByRole('button', {
  name: /submit/i
})``` as the query to test  and if I change the query to  ```screen.getByRole('button', {
  name: 'submit'
})``` it doesn't work either which is the same thing I was experiencing in the circulation-patron-web codebase.

src/components/bookDetails/ReportProblem.tsx
M0nica: sometimes `t` was undefined but I think it was updated elsewhere to not call that function if t is undefined

src/utils/axe.ts
M0nica: I was able to run the app with the AXE variable set without any compile errors. I saw a warning like `It's recommended to have at least one tabbable element inside dialog. The dialog element has been automatically focused.` which I'm assuming is coming from AXE.

kristojorg: Ah no that is from the sign in dialog. It's because there isn't anything inside the sign in dialog at the first moment of rendering, but it shows up after once the redux state is properly set. So it's kind of an annoying warning that doesn't actually affect us

src/components/bookDetails/FulfillmentCard.tsx
kristojorg: Ah so the commented out line is just for development purposes?

M0nica:
@kristojorg can you re-review this? I think all of the outstanding comments have been addressed.

src/utils/axe.ts
EdwinGuzman: You'll see something like this if `react-axe` is running:

Screen Shot 2020-08-10 at 11 54 11 AM

EdwinGuzman: `await import ....` doesn't work for the nextjs app I'm working on but this does:
```
  // eslint-disable-next-line @typescript-eslint/no-var-requires
  const ReactDOM = require("react-dom");
  // eslint-disable-next-line @typescript-eslint/no-var-requires
  const axe = require("react-axe");
```

EdwinGuzman:
As of this morning, the `crummy` url doesn't work for me for the unencrypted ebook. There's no "read online" button and the "download" button downloads an HTML file.

EdwinGuzman:
Okay, so I forgot about uncommenting the test media type to get this to work in dev. But the end-to-end flow is working well for me. There are some minor css issues which could be fixed now:

1. When loading, the frame seems to appear twice.
Screen Shot 2020-08-10 at 12 33 17 PM
Screen Shot 2020-08-10 at 12 33 22 PM

2. The previous button interferes with selecting a chapter and it could be hidden when that setting is open.
Screen Shot 2020-08-10 at 12 35 09 PM

EdwinGuzman:
Right, I don't have all the context or background knowledge to really say much about this, but the code looks fine and can be optimized/refactored later. Monica made some tickets already but if there are other code issues, new tickets can be made. This will need Kristo's approval to keep moving forward, however.

M0nica:
> Right, I don't have all the context or background knowledge to really say much about this, but the code looks fine and can be optimized/refactored later. Monica made some tickets already but if there are other code issues, new tickets can be made. This will need Kristo's approval to keep moving forward, however.


I made follow up tickets for the following:
- Resolving any issues with React-Axe
- Remove references to   “application/vnd.librarysimplified.web-epub” (awaiting the merging of https://github.com/NYPL-Simplified/circulation-patron-web/pull/93)
- Bug with EReader frame appearing twice during loading
- Previous button overlapping with table of contents

PR #87, opened 2020-06-15
Babel config changes

hokei:
These are changes I've made to make the webreader work as a module.  

Do these changes square with the way the application is supposed to work? (`.babelrc` is suspiciously empty to me haha)


src/utils/reader.ts
```diff
@@ -1,16 +1,16 @@
...
-import LocalStorageStore from "../webpub-viewer/LocalStorageStore";
-import IFrameNavigator from "../webpub-viewer/IFrameNavigator";
-import DayTheme from "../webpub-viewer/DayTheme";
-import NightTheme from "../webpub-viewer/NightTheme";
-import BookSettings from "../webpub-viewer/BookSettings";
-import LocalAnnotator from "../webpub-viewer/LocalAnnotator";
-import ServiceWorkerCacher from "../webpub-viewer/ServiceWorkerCacher";
-import ColumnsPaginatedBookView from "../webpub-viewer/ColumnsPaginatedBookView";
-import ScrollingBookView from "../webpub-viewer/ScrollingBookView";
+import { SepiaTheme,
```
M0nica, line 1: This looks good to me to merge into the EReader branch. Do we want that other branch to handle removing the `.../webpub-viewer` directory?

M0nica:
I;m not sure why the linting is failing. It looks like it wasn't able to install node_modules. So maybe re-running will resolve the issue?

kristojorg:
Can you explain what the motivation is behind adding these plugins to the babel config? Next.js should provide everything necessary to compile code with the `next/babel` preset. You can read more about the default Next.js preset [here](https://nextjs.org/docs/advanced-features/customizing-babel-config). 

My general thoughts behind this are that we adopted Next.js in order to offload the effort of creating and maintaining a build configuration as well as server code etc. Basically, we chose convention over configuration. We can extend the babel config from there, but the closer we can stay to the Next.js defaults, the less we have to maintain and the more we can enjoy the optimization and DX work the community has put in to Next.js. So yea, maybe let me know what prompted you to do this and let's investigate whether it's really necessary

kristojorg:
@M0nica it looks like a `yarn.lock` file was committed. AFAIK, NYPL uses npm only. I'm open to switching, but for now we shouldn't have this file in the repo and removing it should fix the failing dependency install in CI.

hokei:
Cool yeah, I've never used NextJS before, but based on how bare bones the configuration looked, I suspected maybe something like this was the case ... =( 

So this is the export configs on the viewer side:  https://github.com/NYPL-Simplified/webpub-viewer/pull/160/commits/3ecd164be0649697835224aa5bbe02d69540c7b5#diff-e5e546dd2eb0351f813d63d1b39dbc48

When that gets pulled in, though, there's a disconnect between the es6 syntax and the commonjs syntax - that's what `@babel/preset-env` is supposed to fix

And `plugin-transform-runtime` is because I think either `async` or generators are being used in the dependency, and `preset-env` uses `plugin-transform-runtime` to backfill that, I think.  

But I feel like of all of the dependencies in CPW there must be someone else exporting es6 modules, so did nextjs handle them in a different way?  Can we handle this in the same way?

src/utils/reader.ts
hokei: Yeah I think that'd be good.

hokei:
@kristojorg Aaah I see what you mean: https://nextjs.org/docs/advanced-features/customizing-babel-config 
these should already be there, I'll take a look and see if there's configs I'm missing.

kristojorg:
@hokei exactly, I think es6 module support should be included. Unless I'm wrong, isn't the whole app using es6 modules for import/export already? What's the error that you're getting when importing these?

hokei:
@kristojorg it looked a lot like https://stackoverflow.com/questions/55329589/babel-nodejs-es6-syntaxerror-unexpected-token-export

I've removed some of this stuff and it still seems to work, which feels similar to on Friday when it looked like it was working for me and then when Monica pulled it it didn't work.  I'm wondering if there's internal cacheing going on or something.

M0nica:
> @M0nica it looks like a `yarn.lock` file was committed. AFAIK, NYPL uses npm only. I'm open to switching, but for now we shouldn't have this file in the repo and removing it should fix the failing dependency install in CI.
> 
> I think the `yarn.lock` is in the base branch this is being PR'd to, so you'll have to remove it there and then you might need to rebase this PR.

I just removed the yarn.lock from the base branch

kristojorg:
Interesting @hokei . I've seen something similar to unexpected token export but with type exports. Maybe it was something like that? Next.js does cache in development but I haven't found it to be very buggy myself. Have never had to delete the cache, but if you do, you can delete `.next` and possibly `node_modules` and reinstall and `npm run dev` again.

kristojorg:
@hokei you'll have to rebase this branch on top of the base branch again in order to fix the CI errors

hokei:
@kristojorg ... it works now, but definitely didn't this morning.  I did clear out everything, but we'll go with it for now haha

@M0nica Without the babel/package changes the only thing that is changed is the changes to `reader.ts` so feel free to merge or not merge this PR into your branch!

M0nica:
@hokei I'll merge this into the EReader branch now. Thanks!

PR #88, opened 2020-06-29
Redesign

kristojorg:
This PR is a complete redesign of CPW, but it maintains much of the underlying logic, so it is effectively a re-skinning of the application. 

For the time being, it does not take theme inputs from the circulation manager, as we have not decided how to proceed with migrating old values to new values. For the time being, the theme is static, but that will change in the next PR.

This design of the application is based on the NYPL Design System, though it only uses one component from the DS at the moment (the Card for BookListItem). All of the other internal components were updated to be styled in a fashion acceptable in the DS, and can be later swapped out as the Design System matures with components imported from there.

Deployment of this PR will require changes to how libraries are using their web settings in the CM. For example, it will not work great to have many header links set, and that option maybe should be changed or removed from the CM. There will likely be some other unforseen issues relating to the various configurations possible in the CM.

### Edit

I also added in the ability to see a list of libraries when you run the app with a config file and visit the root url. It's currently pretty utilitarian, but a big improvement for developers at least when we load the root url of an app running with a config. Just thought I'd let you know in case you see that code and wonder what I was doing that for.

### To do

- [ ] Open a ticket in Jira about updating breadcrumb generating logic
- [x] Make some basic updates to get the app usable on mobile
- [x] Make a note for UX about what in a book card should hold the link to the book
- [ ] Make a ticket to address breadcrumbs clickability indicator (maybe make them permanent underlined)
- [ ] Ask about proper way to address horizontal scroll bar in lanes


package.json
```diff
@@ -17,12 +17,13 @@
...
     "lint": "eslint '*/**/*.{ts,tsx}'",
     "lint:fix": "eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
-    "generate-icons": "npx @svgr/cli --template src/icons/_template.js --typescript --icon --ext tsx -d src/icons src/icons"
+    "generate-icons": "npx @svgr/cli --template src/icons/_template.js --typescript --icon --ext tsx --svg-props 'fill=currentColor' -d src/icons src/icons"
   },
   "dependencies": {
     "@csstools/normalize.css": "^10.1.0",
     "@emotion/core": "^10.0.28",
     "@emotion/styled": "^10.0.27",
+    "@nypl/design-system-react-components": "github:NYPL/nypl-design-system#add-he",
```
hokei, line 26: this doesn't seem to be a real branch?

src/components/BasicAuthForm.tsx
```diff
@@ -69,7 +69,10 @@ const BasicAuthForm: React.FC<AuthFormProps<BasicAuthMethod>> = ({
           `Your ${provider.method.labels.password} is required.`
         }
       />
-      <Button type="submit" sx={{ alignSelf: "flex-end", m: 2 }}>
+      <Button
```
hokei, line 72: Is there a reason to not be using Design System buttons/links/headings?

src/components/Button/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    color = defaultColor,
+    size = defaultSize,
+    variant = defaultVariant,
+    newTab,
+    iconLeft,
+    iconRight,
+    ...rest
+  } = props;
+  return (
+    <a
```
hokei, line 124: what's the difference between the places you use `Link` and the places you use `a` ?

package.json
kristojorg: this is a great point

src/components/BasicAuthForm.tsx
kristojorg: I initially just didn't need to switch, but also was avoiding it because I implemented some logic to render button-looking things that ended up being something different under the hood, eg an anchor tag or a Next.js `Link` that just looked like a button

kristojorg: It shouldn't be hard to switch it out in the future, but I didn't have to change my pre existing buttons much to match the styles needed, so I just kept them

src/components/Button/index.tsx
kristojorg: `Link` is a Next.js Link for navigation within the app, similar to a react-router `Link`, or links from other routing libraries.

package.json
kristojorg: fixed

src/components/BasicAuthForm.tsx
kristojorg: Actually, when I think about it, the real reason is that I wasn't able to get the DS working in this branch until the last week or so, at which point I had already done all of this.

hokei: OK!  Just if it is anything re: there are features you'd like that aren't available in design system, we'd really like to know about those even if we can't get those in.

kristojorg: One thing that I think you'll need to consider eventually is the way something that looks like a button often times needs to have different components backing it, like a button that's an `a` tag or a button that's a `button` or a button that's a `Link` etc. It seems like it would be as easy as passing in a component to render as the "base" of the button, but that turns out to be difficult with typescript when each base component takes different props

hokei:
Some general things clicking through the app -- this might be part of the design, but I thought I'd point it out anyway as stuff that sticks out to me  

On the lanes, clicking on the book cover will take you to the page, but clicking on the book title won't 
None of the authors are clickable
None of the categories are clickable (It seems like you should at least be able to search "historical" if you click them? ) 
On the breadcrumbs, the links and non-links have no visual difference.

hokei:
Looks like it's really breaking on small enough screens (chrome/Mac) 

Screen Shot 2020-06-29 at 3 57 32 PM

hokei:
Something seems off with the breadcrumbs -- 
Whenever you click the breadcrumbs from All Books/XYZlibrary/Fiction (or All Books/XYZlibrary/Book_title section, clicking "All Books" and "XYZ Library" both go to the same page

kristojorg:
> On the lanes, clicking on the book cover will take you to the page, but clicking on the book title won't

I did this because having two identical links right next to each other was an accessibility issue previously. I'm honestly not sure what the right approach is though, and am open to suggestions. 

> None of the authors are clickable
None of the categories are clickable (It seems like you should at least be able to search "historical" if you click them? )

I left it out because we don't have "author" or "category" pages right now, though I think we could add that in the future. But you're right we could make clicking them perform a search.

> On the breadcrumbs, the links and non-links have no visual difference.

That's a good point. What visual difference do you give the "current location" breadcrumb in the DS? It should be that only the links get underlined on hover/focus, but that might not be enough. 

> Looks like it's really breaking on small enough screens (chrome/Mac)

That's true. Mobile was out of scope of this version, but I could definitely do some simple things to fix it up a bit. I'll do that after Edwin submits his review (so I don't make him lose his place).

kristojorg:
> Something seems off with the breadcrumbs --
Whenever you click the breadcrumbs from All Books/XYZlibrary/Fiction (or All Books/XYZlibrary/Book_title section, clicking "All Books" and "XYZ Library" both go to the same page

Yea this isn't a CPW issue, but is an issue with the CM. We need to look at how breadcrumbs are generated in the CM and make some updates. Those two breadcrumbs just have the same href in the data I get from the server. We should make sure we open a jira ticket to address that. I'll make a todo list in this PR description

hokei:
> I did this because having two identical links right next to each other was an accessibility issue previously. I'm honestly not sure what the right approach is though, and am open to suggestions.

Ah yes that's a good point - I don't know the answer, but I'd make a note for UX to address this when they do their review (after you get it deployed on a test environment)

 > That's a good point. What visual difference do you give the "current location" breadcrumb in the DS? It should be that only the links get underlined on hover/focus, but that might not be enough.

I don't actually think DS handles this properly either.  Digital Research Books doesn't show the current page in the breadcrumbs, only parent pages, so everything is clickable.  (But it's still missing underlines, which we use to differentiate links from other things)

hokei:
> Mobile was out of scope of this version
Wow, considering this, mobile looks pretty ... useable!  haha

hokei:
The lane horizontal scrollbar is still visible when it doesn't scroll 
Screen Shot 2020-06-29 at 4 20 30 PM

It's also not possible to click and drag the horizontal scrollbars (though I'm having trouble parsing the w3 UI guidelines, maybe another thing to ask UX if it isn't easily fixable)

kristojorg:
@hokei I just pushed some style changes that should make it WAY more useable on mobile,  LMK what you think : )

> The lane horizontal scrollbar is still visible when it doesn't scroll

This I think is only when using an external mouse (at least in OSX). I'm also not sure the proper way to handle it, although mine are clickable. I think if you don't see them on a horizontal scrolling site, it's bc that site is hiding them with css wizardry, but I don't really think that's accessible. I'm not sure tbh what the best option is here. Made a todo to bring it up with UX.

src/components/BasicAuthForm.tsx
hokei: In DS we have this:  https://nypl.github.io/nypl-design-system/storybook/storybook-static/react/?path=/story/link--button-link is this what you're talking about?  
So you're not thinking about it as a button, you're thinking about it as a link and would pass props accordingly

src/components/BookMetaDetail.tsx
```diff
@@ -4,10 +4,11 @@ import * as React from "react";
 
 const DetailField: React.FC<{ heading: string; details?: string }> = ({
   heading,
-  details
+  details,
+  ...rest
```
hokei, line 8: I can kind of understand the usage of `...rest` in super fundamental elements such as `Button` or `Link` but is the variety of props that can be passed here really that unpredictable?  Is there a way to make this more specific?

src/components/Stack.tsx
```diff
@@ -0,0 +1,36 @@
+/** @jsx jsx */
+import { jsx } from "theme-ui";
+import * as React from "react";
+
+type StackProps = {
+  direction?: "row" | "column";
+  spacing?: number | string;
+};
+const Stack: React.FC<StackProps> = ({
```
hokei, line 9: Is the purpose of the `Stack` (as well as the `Text`) components just to impart styles to things?

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -0,0 +1,308 @@
...
+import { State } from "opds-web-client/lib/state";
+import userEvent from "@testing-library/user-event";
+import { FetchErrorData } from "opds-web-client/lib/interfaces";
+
+function expectViewDetails(utils: ReturnType<typeof render>) {
+  const button = utils.getByRole("link", { name: "View Book Details" });
+  expect(button).toBeInTheDocument();
+  expect(button).toHaveAttribute("href", "/book/test-book-url");
+}
+test("OPEN_ACCESS", () => {
```
hokei, line 14: can we get the test names to all be in the [operation + expected result] format?  
(So here, something like "Renders open access book" or something?

src/components/BookMetaDetail.tsx
kristojorg: Yea this should be taken out, your're right

src/components/Stack.tsx
kristojorg: Yep exactly. The stack just styles children. I have a lot of components for these sorts of uses. Could say this is what css classes are for, but there are some benefits to having the styles live in JS as well

src/components/__tests__/BookListItem.test.tsx
kristojorg: Yep I can do that

hokei:
This looks pretty good to me, but I'll let Edwin give a second opinion since I know nothing about next.

EdwinGuzman:
I haven't looked at the code yet but doing a superficial walkthrough of the app, here are a few things I noticed:

* In the Brazoria library some of the text is cut off for long book titles or multiple authors:
Screen Shot 2020-07-01 at 10 16 37 PM
Screen Shot 2020-07-01 at 10 16 32 PM

This is in one of my libraries:
Screen Shot 2020-07-01 at 10 10 09 PM

* This may not be part of this PR but when I click on "borrow" on an ebook or audiobook on a book page, the login modal pops up. If I exit the modal, through the escape button or by clicking outside the modal, I get redirected to the homepage and not the book page I was just one.

* The footer columns could some some spacing:
Screen Shot 2020-07-01 at 10 10 48 PM

* I noticed this before reading the conversation you and Crystal had about this same section. Is it possible to simply remove the left/right arrow buttons when the lane has too few items to actually scroll? Though not a stopper for this PR.

Screen Shot 2020-07-01 at 10 10 42 PM

*  The header nav links on mobile seem unusable. This is a simulation on iPhone X and the links are close together and can be problematic for users who have a hard time using their thumbs on small links. But I think that's something Crystal can discuss with the UX team about.
Screen Shot 2020-07-01 at 10 17 56 PM


I'll review the code soon and add more feedback soon.

kristojorg:
> In the Brazoria library some of the text is cut off for long book titles or multiple authors:

This is something I don't quite know how to address. The cards have fixed height, and I truncate the title and authors at some length. If both are at the limit of what's allowed, the card will overflow as you see. If I set the truncation much lower, almost every title will be truncated (and usually needlessly). My only thought is to make some logic that actually measures these nodes and makes sure they conform to some height, but that seems... like a lot of code for a simple problem. Do you have any other ideas?

> This may not be part of this PR but when I click on "borrow" on an ebook or audiobook on a book page, the login modal pops up. If I exit the modal, through the escape button or by clicking outside the modal, I get redirected to the homepage and not the book page I was just one.

This is because of how OWC works, I believe. I will be changing it in a subsequent PR at some point.

> The footer columns could some some spacing:

Agreed, screwed that up when I made a mobile optimization, will fix.

> I noticed this before reading the conversation you and Crystal had about this same section. Is it possible to simply remove the left/right arrow buttons when the lane has too few items to actually scroll? Though not a stopper for this PR.

This also would require measuring the viewport and the list of items. It's doable, but I'd rather make it a separate fix after this PR. I also would consider that if we can detect touch devices, the arrow buttons shouldn't be shown and the list should flow to the 16px page margin.

> The header nav links on mobile seem unusable. 

Yea so perfect mobile responsiveness wasn't really part of the scope here, but this PR should still be an improvement on the existing UX. Though if you think it's just a matter of adding some vertical spacing, I can definitely do that in this PR. Would that fix the issue you're referring to?

EdwinGuzman:
Cool, so some issues will be covered later.
I'm not too sure what to do with the card and long titles or multiple authors, but maybe the font can be percentage height to the height of the card. Not sure if that will help actually. It doesn't happen often but when it does it makes it look very off.
I'll start looking over the code.

src/utils/string.ts
```diff
@@ -3,11 +3,11 @@
  * if you select useWordBoundary, it will only
  * truncate at whole words
  */
-const truncateString = (
+export function truncateString(
   str: string,
   length: number,
   useWordBoundary = true
-) => {
```
EdwinGuzman, line 10: Why the switch from this function declaration style?

hokei:
> I'm not too sure what to do with the card and long titles or multiple authors
The authors right now aren't clickable, right?  It might be OK to truncate them.  

How's progress on getting this on a server? It'll really help to present that to UX and flag these issues so they can come up with some better guidance.

src/utils/string.ts
kristojorg: I'm just increasingly of the opinion that it helps with grepping a file to see functions written explicitly instead of as arrow functions. But honestly I go back and forth.

kristojorg:
@hokei You can show this to UX: https://ebooksbeta.lyrasistechnology.org/lyrasis

Keep in mind though that this PR probably won't make the app _perfect_, but it should certainly not make it worse than the current version at any point.

kristojorg:
> The authors right now aren't clickable, right? It might be OK to truncate them.

Ah yes I thought I had truncated the authors. But even if I do that, the problem still sort of remains. If I truncate both authors and title to a length short enough that if both are at their longest, it will still fit in the card, then often one of them will be needlessly truncated because the other one is short. I guess I'll truncate title to be two lines max and authors to be one.

src/components/Text.tsx
```diff
@@ -0,0 +1,41 @@
+/** @jsx jsx */
+import { jsx } from "theme-ui";
+import * as React from "react";
+
+type NativeComponent<
+  T extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>,
+  P = {}
```
EdwinGuzman, line 7: Why does `P` need to be an empty object?

src/components/TextInput.tsx
```diff
@@ -8,12 +8,18 @@ import * as React from "react";
...
+  border: "solid",
+  px: 2,
+  py: 1,
+  fontSize: "0",
+  variant: "text.body.regular",
+  width: "100%",
+  "&::placeholder": {
+    fontStyle: "italic",
+    color: "ui.gray.dark",
+    fontSize: "-1"
```
EdwinGuzman, line 21: Why is a negative value?

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+        available: 0
+      },
+      holds: {
+        total: 23,
+        position: 5
+      }
+    });
+    const node = render(<BookListItem book={reservedBookWithQueue} />);
+    expect(
+      node.getByText("You have this book on hold. Position: 5")
```
EdwinGuzman, line 269: The mobile apps tell you the current position out of the total, so it would't be a bad idea to include her, too. Are you getting the copy from anyone or just writing it yourself?

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+    availability: {
+      status: "available",
+      until: "2020-06-18"
+    }
+  });
+
+  test("displays correct title and subtitle and view details", () => {
+    const node = render(<BookListItem book={downloadableBook} />);
+    expect(
+      node.getByText("You have this book on loan until Thu Jun 18 2020.")
```
EdwinGuzman, line 298: If you could spell out the day and month, that'd be great.

src/components/TextInput.tsx
EdwinGuzman: Does this correspond to https://github.com/NYPL-Simplified/circulation-patron-web/pull/88/files#diff-5ff813e32ef312fe271fd3cf4023a1abR9 ?

EdwinGuzman:
Is there a ticket outlining the proposed changes you implemented in this PR?
Just wondering because the grid/list view changed into just a list view (which I think is simple and looks great) but I am curious if that got looked at and approved from stakeholders or anyone else.

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -1,247 +1,295 @@
...
 import useBorrow from "../../hooks/useBorrow";
+import Stack from "components/Stack";
+import { Text } from "components/Text";
+import { MediumIcon } from "components/MediumIndicator";
+import SvgDownload from "icons/Download";
+import SvgPhone from "icons/Phone";
 
-/**
- * This card will handle logic to download books. There are a few cases
- *  1.  It is open access, in which case show download button for different
```
EdwinGuzman, line 27: Is the logic different this time around or do you not need a comment explanation for this anymore?

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -1,247 +1,295 @@
...
+      );
+    }
+
+    case "RESERVED":
+      return <Reserved book={book} />;
+
+    case "READY_TO_BORROW": {
+      const availableUntil = book.availability?.until
+        ? new Date(book.availability.until).toDateString()
+        : "NaN";
```
EdwinGuzman, line 104: I think using the `"NaN"` value is a bit odd since this is not a number variable but a string. Maybe just undefined or null?

src/components/FormatFilter.tsx
```diff
@@ -29,40 +36,40 @@ const FormatFilter: React.FC = () => {
...
+
+  const value = [allFacet, ebookFacet, audiobookFacet].find(
+    facet => facet?.active
+  )?.href;
   return (
-    <Group
-      role="navigation"
-      aria-label="Format filters"
-      sx={{ display: "flex", py: 0 }}
+    <FormField
```
EdwinGuzman, line 58: Looking at the implementation of this component, it seems unnecessary to create a component for this if it just renders a div with styling. Or are there future cases this will cover?

src/components/Header.tsx
```diff
@@ -19,128 +20,122 @@ export interface HeaderContext extends NavigateContext {
...
-    <header
-      sx={{
-        display: "flex",
-        flexDirection: ["column", "column", "row"],
-        alignItems: ["stretch", "stretch", "flex-end"]
-      }}
-      className={className}
-    >
-      <Link
+    <>
```
EdwinGuzman, line 27: Why is this wrapped in a fragment?

src/components/bookDetails/Recommendations.tsx
```diff
@@ -42,86 +37,26 @@ const Recommendations: React.FC<{ book: BookData }> = ({ book }) => {
...
-              </Link>
-            )
-        )}
-      </div>
-    </div>
+        Recommendations{" "}
+        {isFetching && <LoadingIndicator size="1.75rem" color="ui.gray.dark" />}
+      </H2>
+      {!isFetching &&
+        lanes.map(lane => <Lane key={lane.title} lane={lane} titleTag={H3} />)}
```
EdwinGuzman, line 58: Each `Lane` renders a div but in this case this should be a list of items `<ul><li>lane...</li>....</ul>`.

src/components/BookList.tsx
```diff
@@ -106,82 +59,248 @@ export const ListView: React.FC<{
...
-        >
-          {label}
-        </Button>
-      </div>
+      </DS.Card>
     </li>
   );
 };
 
+const BookListCTA: React.FC<{ book: BookData }> = ({ book }) => {
```
EdwinGuzman, line 117: What does `CTA` stand for?

src/components/Text.tsx
kristojorg: This is just a default argument for the type, so you don't have to pass in extra props to the component, but you can if you want.

src/components/TextInput.tsx
kristojorg: This is following the NYPL design system scale, which ranges from -3 up

src/components/__tests__/BookListItem.test.tsx
kristojorg: I'm just writing it myself. I thought about showing that, but didn't see how it really helped patrons at all. Why do they care how many people are behind them in line?

src/components/__tests__/BookListItem.test.tsx
kristojorg: I was trying to avoid including moment.js in the bundle for this small use case since it's a very very heavy library. This is just the native way JS parses dates to localeString. If it's a major issue to write it this way, I can probably find a smaller date parser though

EdwinGuzman: Oh yes, definitely let's avoid momentjs.

kristojorg:
> Is there a ticket outlining the proposed changes you implemented in this PR?
Just wondering because the grid/list view changed into just a list view (which I think is simple and looks great) but I am curious if that got looked at and approved from stakeholders or anyone else.

No there isn't anywhere that got decided publicly. It was in a conversation between myself and James English. Basically, the list and the grid view are functionally the same, and I was trying to cut down on design + dev time so we could get this into production in early July. I couldn't really foresee any reason both would be required, but you're right it probably should have been a public decision. I am down to add it back in if we decide we want it in a follow on PR, but I wanted to get this at least into the beta branch so Lyrasis can deploy it ASAP

src/components/BookList.tsx
kristojorg: Oh Call to Action, it's language I inherited from the design system

src/components/bookDetails/Recommendations.tsx
kristojorg: Oh interesting. I hadn't considered that. Can you have a list within a list? Because the lane is a list of books, but you're right that the recommendations is a list of lanes...

src/components/BookList.tsx
EdwinGuzman: Gotcha, I haven't seen the DS in a while so forgot parts of it.

EdwinGuzman:
Yes, please add removing the grid view as something to discuss publicly. Since CPW will be used for Open Ebooks, that project may still want that view. 

CC @killerpuppytails

src/components/bookDetails/FulfillmentCard.tsx
kristojorg: Yea so this is because if the `new Date(...).toDateString()` fails, it will return `"NaN"`, so if I made this undefined, then I have to later check for both undefined and `"NaN"`, so I decided it would be `"NaN"` in any failure case and I could just check for that. I'm down to change it though if that makes it confusing to read. Can see how that would be the case.

src/components/bookDetails/FulfillmentCard.tsx
kristojorg: The logic is slightly different, and now these book states are encoded into `getFulfillmentState`. I also updated the documentation on the wiki to contain this info, so I figured it wasn't necessary anymore. But maybe I should add a comment above `getFulfillmentState` linking to that documentation.

EdwinGuzman: Yes, please link to any documentation so we can have more cross-documentation. thanks.

src/components/bookDetails/FulfillmentCard.tsx
EdwinGuzman: Cool, let's keep "NaN" then.

src/components/Header.tsx
kristojorg: Good Q. Probably the remnant an intermediary refactor that I tried somewhere along the road. Will remove

src/components/FormatFilter.tsx
kristojorg: Yea this was done mainly because I thought I would maybe add some logic there in the future, but also because it's my preferred abstraction for repetitive styles. Instead of using css classes that pollute the global scope and can become difficult to maintain, I just extract the common style into a component I can reuse

kristojorg:
@EdwinGuzman would bringing it up on slack in the web-development channel be the way to start that discussion, or is there something else I should do to bubble these decisions up?

src/theme/colors.ts
```diff
@@ -18,17 +18,46 @@ primaries.light = primaries[0];
 
 const grey = "#B3B9BE";
 
+const ui = {
+  black: "#000000",
+  white: "#ffffff",
+  error: "#97272c",
+  link: {
+    primary: "#0071ce",
```
scooooooooby, line 26: The designer on our team switched this to `#0576d3` so it would pass accessibility on white and black backgrounds—you might consider making the same update.

EdwinGuzman:
> @EdwinGuzman would bringing it up on slack in the web-development channel be the way to start that discussion, or is there something else I should do to bubble these decisions up?

Yes, but I think a larger audience, especially those who help make decisions, will be better and not just web developers. It can be a start for issues/features moving forward but it'll be good to check with Risa to see if that grid layout was an expected feature of the Ebooks development.

killerpuppytails:
Thanks for the tag, @EdwinGuzman . I want to be sure that I understand what precisely is meant by you and @kristojorg re: list view vs grid view, since I can't find another reference in this PR. Could you point me to mockups or give a quick description?

kristojorg:
Hi @killerpuppytails, below is a mockup of the new "list view" page which is included in this PR. In the past, this repo had both a list view and a "grid view". I've attached screenshots of those below. I decided to remove the grid view for a couple reasons:

1. We were short on time in this redesign cycle and both of these screens provided the same functionality, just with a different UI. James English and I thought the list view could serve both roles for now. 
2. We figured that if the grid view is desired, it could be added in a follow on PR later down the road. 
3. With or without the grid view, the new CPW designs should be far more useable than the old design.


The new list view:

List View Page

The old list view:

Screen Shot 2020-07-06 at 4 52 45 PM

The old grid view:

Screen Shot 2020-07-06 at 4 49 13 PM

killerpuppytails:
Thanks so much, @kristojorg ! One follow-up comment and a few follow-up questions:

Comment: We will absolutely need a more visual layout for Open eBooks, as the text-heavy smaller font doesn't suit young children's use at all. However, I don't believe the grid view as shown is that solution.

Questions:
1. Does the CPW as you're building it support a lane view the way we have it on catalog.openebooks.us, or is that not in your model right now due to DS considerations?
2. If lane views aren't part of the model, what would be involved in adding it?
3. This particular grid view was designed by Amigos, and if NYPL wanted something similar to a block layout we probably would not want it to look like that. Do you have an estimate of how much time is needed to assess a new design, so we can figure out timeframes and impact?

kristojorg:
@killerpuppytails below is the lane view of the updated version:

Boise Lanes Page


By assess a new design, do you mean to design and implement a new grid/block view? It's difficult to say without knowing what the desired UX is and what the process is for approval (from what I gather it is somewhat more stringent that I am used to at Lyrasis). 

I'm hoping that maybe we can get this version merged in without the grid view as I believe it is in almost every way better than the current version pictured above, even missing the grid view. And as NYPL prepares to put it into production via openEbooks or otherwise, maybe NYPL can add a grid view later? For the time being, OpenEBooks can stay on the version it is using now.

kristojorg:
@scooooooooby 

> * I notice that there's something causing the page to hang, so the tab has a "loading" circle perpetually. It looks like everything is working on the page that I can tell, however

Can you tell me which page and also what circulation manager/config setup you're using so I can debug this? If you're running it in dev mode, it may be the react hot loader connection. In production, it could well be simply because the CM sometimes serves many megabytes of JSON for various collections. But also could be something wrong, so I'd like to check.

> * On the cards, only the image is clickable, but not the title, which feels odd. I would recommend a few different paths based on what is feasible for you:
> 
> a) Make the title a link, and make the author a link, because author (being red) looks like a link
> b) If authors don't have links, make the entire card clickable, because there's only one place the card can take you.

This is an issue I've been trying to figure out, since you can't have multiple links going to the same place as siblings for accessibility reasons. But you're right that if the authors don't have links (they don't, as far as I know), we could make the whole card a link. I'll do that for now. 

> * When images fail to load there's an error message, but it's pretty noisy. I would consider having a book icon fallback instead of the error message. We have one in the [Figma file here](https://www.figma.com/file/74YfQYZ9zEYaYCjZG3bYJ3/CPW-v2?node-id=10968%3A1918) (first icon, top row).

Good point. I think I might try to transition the BookCard component to use the DS Image component. Does that one have a fallback? 

> * Section headers and the "See More >" links don't share a baseline, so the link looks like it's floating in space arbitrarily. If they're sharing the same flex container, I'd do a `align-items: baseline` and decrease the `margin-right` on the header to something ~half the size that it currently is.

Good idea : )

> * The first two columns in the footer could use more of a gutter `> * (::not-last-child) margin-right: 1rem`, or something to that effect.


> * Same note about the author—being red, it looks clickable, but it's not! I'd recommend making it clickable, or removing the red styling.

Will do, I think you're right.

> * The download icon is looking a little fuzzy for some reason—it'd be great to even just know where the issue is coming from. Is it the SVG file itself? Is it the way it's being implemented?

Hmmm good question. I'll look in to this.

> * Lyrasis should perhaps consider their CTAs for when a user is ready to read a book here—I think calling out "Open this book on your phone or tablet with the SimplyE app" would give users the next action they need to do, especially on all these open source books. Right now the secondary CTAs are all the same size and have a direct action associated with them, so they feel like the next thing you should do. Moving them out of the main gray box might help, potentially.

Are you not seeing a message to read it on SimplyE, or are you suggesting a change to that language? I agree the secondary CTAs are still too prominent. Maybe we should have made a link that says "I'd rather read on my computer" which then shows the download options either in a modal or directly on the page. Sadly I think it's a bit too late to make that big of a change right now, but it could be a follow up PR.

src/components/TextInput.tsx
kristojorg: And yes it does correspond to that theme value

kristojorg:
@killerpuppytails sounds good! My guess is that if it uses cards relatively similar to the ones in the lane view, it would be very simple to design and implement. The changes required are:

- Add a new `GridView` component that renders the books as a grid of cards. Cards are already built for the lane, so this shouldn't be difficult.
- Add some way to toggle between the two. Probably a state variable in the Grid/List view container and a button in the header next to the filter dropdowns. 
- Where it may get complicated is if OpenEbooks wants the grid view to be default, while others want the list view to be default. That would require some changes to the CM presumably to add a new setting.

scooooooooby:
> Can you tell me which page and also what circulation manager/config setup you're using so I can debug this? If you're running it in dev mode, it may be the react hot loader connection. In production, it could well be simply because the CM sometimes serves many megabytes of JSON for various collections. But also could be something wrong, so I'd like to check.

I can't seem to replicate this problem anymore! Uh, problem fixed??? 🤨😂 (It was something I was seeing originally on what I assume is a QA environment [here](https://ebooksbeta.lyrasistechnology.org/lyrasis/collection/groups)).

> > * When images fail to load there's an error message, but it's pretty noisy. I would consider having a book icon fallback instead of the error message. We have one in the [Figma file here](https://www.figma.com/file/74YfQYZ9zEYaYCjZG3bYJ3/CPW-v2?node-id=10968%3A1918) (first icon, top row).
> 
> Good point. I think I might try to transition the BookCard component to use the DS Image component. Does that one have a fallback?

The `Image` component doesn't have a fallback, but that's a great idea, so I'll file a ticket for it. The icon I supplied isn't in the React library yet, so you'd have to add it yourself here. 

> > * Lyrasis should perhaps consider their CTAs for when a user is ready to read a book here—I think calling out "Open this book on your phone or tablet with the SimplyE app" would give users the next action they need to do, especially on all these open source books. Right now the secondary CTAs are all the same size and have a direct action associated with them, so they feel like the next thing you should do. Moving them out of the main gray box might help, potentially.
> 
> Are you not seeing a message to read it on SimplyE, or are you suggesting a change to that language? I agree the secondary CTAs are still too prominent. Maybe we should have made a link that says "I'd rather read on my computer" which then shows the download options either in a modal or directly on the page. Sadly I think it's a bit too late to make that big of a change right now, but it could be a follow up PR.

I was suggesting a change to the language and moving the secondary CTAs, but I understand if that's out of scope for this redesign—like with all the other feedback, it's totally up to you what you decide to incorporate (or, rather, what you feel you have time to incorporate!)

kristojorg:
@scooooooooby sounds good!

I did look at the download icon, and I think it's not blurry, just really thin. I could be wrong though. What do you think?

image

scooooooooby:
@kristojorg weird... this is what I'm seeing in Chrome on my computer:

Screen Shot 2020-07-06 at 3 53 44 PM

It's not a big deal. Could be that the icon itself isn't pixel-snapped.

kristojorg:
@scooooooooby  That's also what I see when zoomed out, but it gets clearer when I zoomed in, which is why I though it's because it's too thin maybe?

scooooooooby:
@kristojorg yeah, I think you're right... I'll log an issue, I think this icon just isn't pixel-snapped so you can't display it that small

src/components/FormatFilter.tsx
kristojorg: Removed this

src/components/bookDetails/Recommendations.tsx
kristojorg: Make the lists `li` elements

kristojorg:
Okay all, I've implemented the following changes according to feedback:

- Make whole card a link
- Make See More links share a baseline
- Fix gutter in footer
- Make authors black not red
- Change link primary to `0576d3`
- Make the lane itself an LI Element within a list of lanes
- Remove Fragment from Header
- Look at `FormField` component
- Link to documentation in `getFulfillmentState`

Let me know if there is anything else you find @EdwinGuzman

EdwinGuzman:
What did you mean by  "Look at `FormField` component"?

Everything else looks good. Can you just update the readme and package.json with a new version before I merge?

kristojorg:
Okay I bumped the version number to 2.2.0 and added to the changelog. I think this is ready when you are

EdwinGuzman:
Sweet, thanks!

PR #89, opened 2020-07-07
Open access bug

kristojorg:
This PR will update CPW to not display open access links until a book has been loaned, unless the library in question does not have auth enabled.

It is a draft because it will require a corresponding change to opds-web-client and possible circulation-web. 

It also was branched off of the `design` branch, so once that PR is merged this one will be much cleaner.

### DIFF
--- a/src/utils/__tests__/book.test.ts
+++ b/src/utils/__tests__/book.test.ts
@@ -46,113 +46,145 @@ describe("get authors", () => {
 });
 
 describe("getFulfillmentState", () => {
-  test("returns AVAILABLE_OPEN_ACCESS when open access links present and status is 'available'", () => {
-    expect(getFulfillmentState(bookFixture)).toBe("AVAILABLE_OPEN_ACCESS");
+  test("returns AVAILABLE_OPEN_ACCESS when open access links present and status is 'available' and book is borrowed", () => {
+    expect(getFulfillmentState(bookFixture, true)).toBe(
+      "AVAILABLE_OPEN_ACCESS"
+    );
+  });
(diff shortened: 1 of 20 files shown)


kristojorg:
Because of how I chose to solve this, it no longer requires a change in opds-web-client. 

I have updated the logic in the `FulfillmentCard` and `getFulfillmentStatus` to only show open access links when a book is confirmed to be borrowed by checking against a user's loans. 

I also updated some tests surrounding this functionality that were flaky by splitting them into multiple tests and mocking out `download.js` and `window.open` to eliminate test errors.

src/utils/__tests__/book.test.ts
```diff
@@ -46,105 +46,145 @@ describe("get authors", () => {
...
+  test("returns AVAILABLE_OPEN_ACCESS when open access links present and status is 'available' and book is borrowed", () => {
+    expect(getFulfillmentState(bookFixture, true)).toBe(
+      "AVAILABLE_OPEN_ACCESS"
+    );
   });
-  test("available to access", () => {
+  test("returns AVAILABLE_TO_BORROW when open access links present and status is 'available' and book is not borrowed", () => {
+    expect(getFulfillmentState(bookFixture, false)).toBe("AVAILABLE_TO_BORROW");
+  });
+  test("does not return AVAILABLE_OPEN_ACCESS is status is not 'available'", () => {
```
EdwinGuzman, line 57: I think you meant `... IF status is ...`.

kristojorg: true

PR #90, opened 2020-07-10
no audiobook downloads

kristojorg:
This PR removes download options for audiobooks since they cannot be played outside of SimplyE. Instead, we only show the "You are now ready to read this book in SimplyE!" banner.

### DIFF
--- a/src/utils/book.tsx
+++ b/src/utils/book.tsx
@@ -8,6 +8,7 @@ import {
 import { BookFulfillmentState } from "interfaces";
 
 import { Book, Headset } from "../icons";
+import { getMedium } from "opds-web-client/lib/utils/book";
 
 export function getAuthors(book: BookData, lim?: number): string[] {
   // select contributors if the authors array is undefined or empty.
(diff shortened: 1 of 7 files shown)


src/utils/book.tsx
```diff
@@ -107,6 +108,13 @@ export function getErrorMsg(error: FetchErrorData | null): string | null {
   return null;
 }
 
+export function bookIsAudiobook(book: BookData): boolean {
+  if (getMedium(book) === "http://bib.schema.org/Audiobook") {
```
EdwinGuzman, line 112: Any way that the string can be a constant variable coming from somewhere rather than the string?

EdwinGuzman: You also just return `!!(getMedium(.....) ...)`

kristojorg: The return type of `getMedium` is a `MediaType` enum, which is a union of all those strings, so if that request is for DX and error checking, that should be handled by typescript. It will autocomplete the strings for you since there are limited options. 

And yea that's true, I can simplify the function to that form if you prefer it. I thought it good to add the function so that when developers are working on the app they don't have to think about these media types much, can just use the util function.

EdwinGuzman: Cool, that makes sense. Just keep it as the function then.

EdwinGuzman:
It just needs the tests to pass.

PR #91, opened 2020-07-10
replace node with utils in tests

kristojorg:
This PR makes a simple language refactor across all of our test files and the readme. The result of `render` from `react-testing-library` is not actually a node. It is a set of utilities to query the dom by, as well as other utils we pass in via `test-utils/index.tsx`. I had started to transition slowly to using `utils` instead of `node`, but this should avoid confusion in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix: Only show download options for open-access books once they have been borrowed and are present in a user's loans. Download options should still be shown for open-access books in libraries that do not have any auth enabled.
 - Fix: Don't perform state update on unmounted `BorrowCard`. 
+- Refactor: Rename `node` in tests to `utils`, as it is more accurate.
 
 ## 2.2.0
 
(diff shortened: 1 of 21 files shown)


EdwinGuzman:
This is the last of four PRs for CPW. Do you want to merge them all into beta and combine them into one release? I don't think any of the PRs have an updated changelog or package.json version. I tend to have a PR `feature-branch` where I push multiple PRs into for one big release rather than have multiple small releases, but it's up to you how you want to release this.

kristojorg:
Well it won't get released into production until we merge `beta` into `master`, and I thought you would appreciate having them split up a bit more so that I'm not making one big PR like I did on the redesign. 

I didn't update changelog/version because I wasn't sure which would go in first. I sort of think we should start updating those things only when merging from `beta` into `master`. That way we can keep a line of 

```
### New Changes

- Some new change
```

In the changelog and add to it as we merge into beta, and then when it comes time to release, we bump the version number, edit that heading in the changelog, and merge into master. What do you think?

EdwinGuzman:
Gotcha, yes. I forgot that `beta` is like a `development` branch and not a beta branch that will be released a beta version. I do appreciate the split up PRs, it's easier to see what the changes are and to review.

PR #92, opened 2020-07-10
Fix set state on unmounted component

kristojorg:
This PR fixes a simple error we got in the borrow card where we would call `setLoading(false)` after the component had already mounted. We now track whether the component is mounted or not and only update the sate if it is still mounted (which would happen if there was an error in the borrow api call).

### DIFF
--- a/src/hooks/useBorrow.ts
+++ b/src/hooks/useBorrow.ts
@@ -5,6 +5,7 @@ import { getErrorMsg } from "utils/book";
 import { useActions } from "opds-web-client/lib/components/context/ActionsContext";
 
 export default function useBorrow(book: BookData) {
+  const isUnmounted = React.useRef(false);
   const [isLoading, setLoading] = React.useState(false);
   const bookError = useTypedSelector(state => state.book?.error);
   const errorMsg = getErrorMsg(bookError);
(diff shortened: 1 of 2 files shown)


src/hooks/useBorrow.ts
```diff
@@ -25,6 +26,13 @@ export default function useBorrow(book: BookData) {
     }
   };
 
+  React.useEffect(
```
EdwinGuzman, line 29: I haven't used `useEffect` much and not for a function only (as in, not for non-component functions) before. `useEffect` is called for whatever component calls `useBorrow` right? So the `useEffect` is essentially for each component that calls `useBorrow` rather than for `useBorrow` directly, right? I'm just not seeing how `isUnmounted` works when this is not actually a component that gets mounted.

kristojorg: Ah yes, basically the custom hook can only ever be used in the body of a function component, just like `useEffect` and other built in hooks. So therefore, the hook is associated with a specific component instance, and it runs on every render. It's essentially the same as if this were inlined into the component in every place the `useBorrow` is used. So `useEffect` works the same as if it were written directly in the component. 

So yea, you're right. It's tracking the mount status of the component that calls `useBorrow`

PR #93, opened 2020-07-10
Loads an encrypted ebook

hokei:
To be looked at in conjunction with 
https://github.com/NYPL-Simplified/webpub-viewer/pull/164 and 
https://github.com/NYPL-Simplified/axisnow-access-control-web/pull/1

To run this, please use 
`NEXT_PUBLIC_COMPANION_APP=openebooks NEXT_PUBLIC_AXIS_NOW_DECRYPT=true SIMPLIFIED_CATALOG_BASE=https://qa-circulation.openebooks.us/USOEI/groups SHORTEN_URLS=false npm run dev:https`


localhost-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
hokei, line 1: These are needed to run cpw in https in localhost.

src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -7,7 +7,7 @@ const BookPage = () => {
   const { bookUrl } = router.query;
 
   /*This can be updated to point to an external manifest.json */
-  const bookManifestUrl = `/${bookUrl}/OEBPS/package.opf`;
+  const bookManifestUrl = `https://node.axisnow.com/content/stream/9781599904993/META-INF/container.xml`;
```
Comment (leonardr) [inline src/pages/[library]/read/[bookUrl].tsx:10]: By the time you merge this, this URL should be derived from bookUrl (maybe you even use bookUrl as is). Starting with container.xml rather than package.opf is the right move.

leonardr:
This is definitely the right approach.

src/pages/[library]/read/[bookUrl].tsx
Comment (hokei) [reply src/pages/[library]/read/[bookUrl].tsx:10]: Is it ok to call 
baseUrl/library/reader?url=https://node.axisnow.com/content/stream/9781599904993/META-INF/container.xml or similar?  (@M0nica would it look like this or is there a fun next thing we can do to make it read)

Comment (leonardr) [reply src/pages/[library]/read/[bookUrl].tsx:10]: Yeah, AxisNow does almost exactly that so it should be fine.

.env
```diff
@@ -1,5 +1,5 @@
 # Can only set one of the following variables
-# CONFIG_FILE=config/cm_libraries.txt
+CONFIG_FILE=config/config.txt
```
kristojorg, line 2: Just an FYI you can define an `.env.local` which will override the `.env` file and will be ignored by git. I set up the `.env` file initially to just be examples. We shouldn't change the `.env` file that is in github

kristojorg: Maybe I should make this more clear in the README

hokei: aaaah it's there, I just missed it orzorz

package.json
```diff
@@ -9,7 +9,8 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "scripts": {
-    "dev": "next dev",
+    "postinstall": "if test \"$AXIS_NOW_DECRYPT\" = \"true\" ; then cd axisnow-access-control-web && npm install ; fi ",
+    "dev": "CONFIG_FILE=config.txt node server.js",
```
M0nica, line 13: Should there be two `dev` scrips configured? one for `next dev` nd one for `node server.js`?

hokei: With this change it doesn't really make sense to run in dev without https so it does need to call  
https://nextjs.org/docs/advanced-features/custom-server

I can't figure out the exact command that `next dev` is running but my guess is that it was calling the default `server.js` so it's not necessary now

kristojorg: If this gets merged I'd like to try and find a way to do it without the custom server, and also we definitely shouldn't set the config file env var here. But yes, `node server.js` takes the place of `next dev` in this case

hokei: I think (something like) this is going to be merged eventually.  I guess the hope is that it can be run in http if the `AXIS_NOW_DECRYPT=false` but is there a way to make it run https without a custom server?

Yeah the config.txt doesn't need to be here...

kristojorg: You're referring specifically to localhost right, because HTTPS should work in production by default. If it's only development we need this for, then I think it's totally fine. Let's just add a comment to the top of the file and maybe rename it `dev-server.ts` so that it's clear it isn't the production server?

src/utils/reader.ts
```diff
@@ -68,30 +69,42 @@ function initBookSettings(element, webpubManifestUrl, catalogName) {
   const night = new NightTheme();
   const paginator = new ColumnsPaginatedBookView();
   const scroller = new ScrollingBookView();
-  BookSettings.create({
+
+  //TODO: Check that the book is of type application/vnd.librarysimplified.axisnow+json
```
M0nica, line 73: The type for the encrypted epub from @leonardr's test server is actually `"application/vnd.librarysimplified.web-epub;profile=axis-encryption"` instead of "application/vnd.librarysimplified.axisnow+json". If this is the format of actual encrypted docs then `FullfillmentCard.tsx` anf the types in `opds-web-client` will need to be updated to support loading this type in the online viewer.

leonardr: The format you'll be seeing on real (non-test) servers is always `application/vnd.librarysimplified.axisnow+json`.

hokei:
This isn't going to be able to be built until https://github.com/NYPL-Simplified/axisnow-access-control-web/pull/1 is merged.  (I also still haven't completely figured out the github actions piece, but will do it after the above PR is merged.)

hokei:
It would be super useful to have some help running this app with everyone's different configurations to make sure it still works.  (And also ideas on how to test that this works with all the different configurations)

.github/workflows/ci.yml
```diff
@@ -38,6 +38,14 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 13.x
+      - name: clone submodule
```
hokei, line 41: this hasn't been tested.

package.json
```diff
@@ -33,8 +33,8 @@
     "font-awesome": "^4.7.0",
     "history": "^4.10.1",
     "jsdom": "^16.2.2",
-    "library-simplified-webpub-viewer": "0.1.3",
-    "lodash": "^4.17.15",
+    "library-simplified-webpub-viewer": "git://github.com/NYPL-Simplified/webpub-viewer#accessControl",
```
M0nica, line 36: Reminder to update this once https://github.com/NYPL-Simplified/webpub-viewer/pull/164 is merged.

src/components/BookList.tsx
```diff
@@ -137,7 +138,33 @@ export const BookListItem: React.FC<{
 const BookListCTA: React.FC<{ book: BookWithUrl }> = ({ book }) => {
   const isBorrowed = useIsBorrowed(book);
   const fulfillmentState = getFulfillmentState(book, isBorrowed);
-  const { borrowOrReserve, isLoading, errorMsg } = useBorrow(book);
+  const { borrowOrReserve, isLoading, errorMsg } = useBorrow();
+  const getButtons = (buttonType: BorrowButtonType) => {
+    const isBorrow = buttonType === "Borrow";
+    const loadingText = isBorrow ? "Borrowing..." : "Reserving...";
+    return book.allBorrowLinks?.map(link => {
+      let label: string;
```
M0nica, line 146: Can this be rewritten a conditional without `let`?

src/components/BookList.tsx
```diff
@@ -162,14 +189,8 @@ const BookListCTA: React.FC<{ book: BookWithUrl }> = ({ book }) => {
...
         <>
-          <Button
-            onClick={borrowOrReserve}
-            color="ui.black"
-            loading={isLoading}
-            loadingText="Borrowing..."
-          >
-            Borrow
-          </Button>
+          {getButtons("Borrow")}
```
M0nica, line 192: Can "Borrow" but put into a constant since this `BorrowButtonType` is referenced in multiple places? To avoid issues if it is later renamed / reduce typos.

src/components/BookList.tsx
```diff
@@ -193,14 +214,8 @@ const BookListCTA: React.FC<{ book: BookWithUrl }> = ({ book }) => {
...
         <>
-          <Button
-            onClick={borrowOrReserve}
-            color="ui.black"
-            loading={isLoading}
-            loadingText="Reserving..."
-          >
-            Reserve
-          </Button>
+          {getButtons("Reserve")}
```
M0nica, line 217: Can "Reserve" also be converted to a constant?

Maybe `const BUTTONTYPES = {RESERVE: "Reserve", BORROW: "Borrow}`

src/components/WebpubViewer.tsx
```diff
@@ -1,19 +1,56 @@
...
+  return await reader(entryUrl, catalogName, decryptorParams);
+};
 
 const BookPage = () => {
   const library = useLibraryContext();
   const router = useRouter();
   const { bookUrl } = router.query;
-  const bookManifestUrl = `${bookUrl}/META-INF/container.xml`;
+
+  const bookManifestUrl = `${bookUrl}`;
```
M0nica, line 41: At some point are we going to rename manifestUrl in `circulation-patron-web` and `webpub-viewer` since the url now is a generic startUrl where the actual manifest url is extracted by webpub-viewer's internal logic?

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -211,6 +213,74 @@ describe("ready to borrow book", () => {
...
+          books: []
+        }
+      })
+    });
+    // click borrow
+    userEvent.click(utils.getByText("Borrow to read on a mobile device"));
+    expect(updateBookSpy).toHaveBeenCalledTimes(1);
+    expect(updateBookSpy).toHaveBeenCalledWith("/adobe-borrow-link");
+    const borrowButtons = await utils.findAllByRole("button");
+    // Both buttons have the "Borrowing..." text and are disabled
```
M0nica, line 274: I think this behavior should potentially be revisited (in future work) as I think as a user if there are two buttons and I click one button, that only one would changed to loading/borrowing even if both are disabled.

src/components/__tests__/MyBooks.test.tsx
```diff
@@ -133,7 +133,6 @@ test("sorts books", () => {
   const bookNames = utils.queryAllByText(/Book Title/);
   expect(bookNames[0]).toHaveTextContent("Book Title 11");
   expect(bookNames[1]).toHaveTextContent("Book Title 10");
```
M0nica, line 135: Why was` expect(bookNames[2]).toHaveTextContent("Book Title 0");` removed from the tests?

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -262,6 +291,12 @@ const AccessCard: React.FC<{
           </Text>
           <Stack sx={{ justifyContent: "center", flexWrap: "wrap" }}>
             {dedupedLinks.map(link => {
+              if (
+                NEXT_PUBLIC_COMPANION_APP === "openebooks" &&
+                link.type === "application/vnd.librarysimplified.axisnow+json"
```
M0nica, line 296: Should the AXIS_NOW environment variable also be a condition for the `ReadOnlineButton` to render for axisnow resources?

kristojorg:
In terms of testing how this works with all the different configs -- that's a general problem that we ideally want to address at some point, though I'm not sure how. We need to have a clearer idea of what the server config guarantees are so we can set up some automated testing for all of the various configs. Or alternatively we could lean on something more manual and have a single default config file that has a list of demo CMs with various configs and we can have a version of CPW published with that config file for every PR, which would allow us to at least manually test.

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -319,6 +330,116 @@ describe("ready to borrow", () => {
...
+    // click borrow
+    userEvent.click(utils.getByText("Borrow to read online"));
+    expect(updateBookSpy).toHaveBeenCalledTimes(1);
+    expect(updateBookSpy).toHaveBeenCalledWith("/axis-borrow-link");
+
+    // the borrow button should be gone now
+    await waitForElementToBeRemoved(() => utils.getByText("Borrowing..."));
+  });
+
+  test("shows loading state on both buttons when borrowing", async () => {
```
kristojorg, line 388: one question that this brings up for me... what happens if a user has a book borrowed with two borrow links at the same time? do we know what this looks like in terms of response from the CM and in the UI?

src/utils/__tests__/reader.test.ts
```diff
@@ -0,0 +1,9 @@
+// describe("Get Reader with Decryption", () => {
+//   test("");
+// });
+
+// describe("Get Reader with no decryption", () => {});
+
+// describe("Throws error when Decryptor is not available but AXIS_NOW_DECRYPT is set to true", () => {
+//   test(() => {});
+// });
```
kristojorg, line 9: should this be included or uncommented?

src/utils/book.tsx
```diff
@@ -139,3 +139,5 @@ export const bookMediumMap: {
   "http://schema.org/EBook": { name: "eBook", icon: Book },
   "http://schema.org/Book": { name: "Book", icon: Book }
 };
+
+export type BorrowButtonType = "Borrow" | "Reserve";
```
kristojorg, line 143: I think in my suggested refactor we wouldn't need this type, but if we still need it i think it should be moved to `interfaces`

src/utils/reader.ts
```diff
@@ -14,7 +16,11 @@ import {
   ScrollingBookView
 } from "library-simplified-webpub-viewer";
 
-export default async function (bookUrl: string, catalogName: string) {
+export default async function (
+  bookUrl: string,
+  catalogName: string,
+  decryptorParams?: any
```
kristojorg, line 22: does this need to be `any`? i don't fully understand how this works but what is the `initBookSettings` expecting?

src/utils/reader.ts
```diff
@@ -68,30 +79,45 @@ function initBookSettings(element, webpubManifestUrl, catalogName) {
...
+
+  const Decryptor = AXIS_NOW_DECRYPT
+    ? await import("../../axisnow-access-control-web/src/decryptor")
+    : undefined;
+  let decryptor;
+  if (Decryptor) {
+    decryptor = Decryptor
+      ? await Decryptor.default.createDecryptor(decryptorParams)
+      : undefined;
+  }
```
kristojorg, line 91: I don't think you need the wrapping `if (Decryptor)` statement here, so the `decryptor` can be a `const`.

kristojorg (changes requested):
okay this looks great! I made a few comments but otherwise looks good to me

.env
```diff
@@ -11,4 +11,7 @@ CACHE_EXPIRATION_SECONDS=86400
 # turn this on to enable react-axe in development
 # process.env values are strings, so setting this to "false" will
 # unexpectedly turn it on.
-# REACT_AXE=true
\ No newline at end of file
+# REACT_AXE=true
+
+# turn this on to enable axisNow decryption
+# AXIS_NOW_DECRYPT=true
```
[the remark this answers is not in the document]
hokei: Gecause the Decryptor is private, there's no guarantee that every person who wants to use CPW has access to Decryptor.  As a good OSS project, we want to make sure that CPW still works if you want to use it with a circulation manager on an unencrypted feed and don't have access to Decryptor.

src/utils/__tests__/reader.test.ts
hokei: Yeah I still need to write these tests, 😬

src/components/WebpubViewer.tsx
M0nica: Oh yeah now that the `bookManifestUrl` is the same value as `bookUrl` it probably makes more sense to pass it directly.

M0nica:
> The label "Borrow to read on a mobile device" seems very long for a button. I can't run this right now because I'm on a loaner computer, so I don't know how it looks, but maybe that would be better as a label above the button? 

@kristojorg for reference the last time I ran this branch the buttons on the book details page looked like this:

Screen Shot 2020-08-24 at 11 03 12 AM

src/components/WebpubViewer.tsx
```diff
@@ -1,19 +1,56 @@
...
 
   const { catalogName } = library;
+  const { fetcher } = useActions();
 
   React.useEffect(() => {
     if (typeof window !== "undefined") {
```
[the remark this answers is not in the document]
M0nica: +1 ...useEffect runs initially after the first render so the window check can be removed.

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -160,6 +166,9 @@ describe("available to borrow", () => {
     const borrowButton = await utils.findByRole("button", {
       name: /Borrowing.../i
     });
+
+    expect(_useBorrowSpy).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
M0nica: In the UI it also looked to me like it was trying to borrow once for each version (atleast on the book search page)

src/components/BookList.tsx
hokei: This is how it looks:  

Screen Shot 2020-08-24 at 11 48 34 AM

Screen Shot 2020-08-24 at 11 49 10 AM


I'll take a look again at this logic, it's currently pretty confusing. 

>     5. By "Read on a mobile device" do you mean "Read with SimplyE/OpenEbooks"? I think we should be specific about that since on a mobile device is ambiguous.

So that's not exactly what's happening.  Books can either be served in `application/vnd.librarysimplified.axisnow+json` or other formats such as Adobe ACM, etc.  
If you borrow a book in a format, you must also read it in that format.  Before, when only one borrow and read link was ever shown, this wasn't a problem.  Now, though, we want to give people the choice of formats.  

However, showing them "Borrow `application/vnd.librarysimplified.axisnow+json` format" is clearly not useful.  "Read Online/Read on Mobile" seems to be the most useful distinction because that's what people care about?  SimplyE/Open Ebooks probably won't work here because the entire product is Open Ebooks, right?  

Talking through this, I'm also realizing that if a book has more than one mobile link it'll show up twice ... gah, I need to put in logic to handle that.  I might separate that out into a separate ticket depending on how this refactor work goes ...

src/components/bookDetails/FulfillmentCard.tsx
hokei: I asked Risa about this -- SimplyE will want to be able to access encryption/decryptions options eventually, but not for like another year (re: current plans)  

Personally, I think of the AXIS_NOW env variable as a "Do I have access to the axis-now-access-control-web repo" 
Is there a better way to represent this?

.gitignore
```diff
@@ -31,4 +31,8 @@ yalc.lock
 wallaby.js
 config/
 .vscode/
+
+lib/
+
+config.txt
```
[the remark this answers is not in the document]
hokei: aaah I completely missed that.  I'll put my config into /config

.gitignore
```diff
@@ -31,4 +31,8 @@ yalc.lock
 wallaby.js
 config/
 .vscode/
+
+lib/
```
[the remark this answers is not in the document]
hokei: This was coming from something I was trying earlier, I need to remove it.

hokei:
Note, currently, loading an ebook for the first time is incredibly slow.  Like, walk away and eat a snack slow.  It also will show you "there is an error loading this page"  while it is loading.  In this PR I want to replace that error page with a loading page, but speeding up ebook loading is going to happen in a different ticket

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
kristojorg: Okay well at least it will give an appropriate error

kristojorg:
@M0nica thanks! I think the ui there should be updated rethought, what do you all think? It seems like too much for the purpose of that card. Maybe we put the options into a modal and just leave a single "Borrow" button on the card? We could reuse that UI for the book details page then. Thoughts?

hokei:
@kristojorg I think that would be great!  I also think, let's get this ticket to a "this is a reasonable thing to use" state and separate that out into a separate ticket as this ticket is ostensibly a "load encryption" ticket and I'm pretty excited to get this to a state where people who aren't us can see that encrypted ebook reading works

.gitignore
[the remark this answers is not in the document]
hokei: So when I put it into `/config/config.txt` something in the code is still looking for `config.txt` in the main file.  What am I doing wrong?
Screen Shot 2020-08-26 at 3 40 14 PM

src/utils/reader.ts
hokei: Currently it might need to be, but theoretically, decryptor implementation details shouldn't be exposed in this public repo, so `any` is a way to hide that.

src/utils/__tests__/reader.test.ts
hokei: How would y'all feel if I moved this and the webpubviewer.test.tsx to a different ticket to try and unblock work?

hokei:
Screen Shot 2020-08-26 at 5 07 30 PM

Since the two pages are now sharing the same button, this is what the book list page looks like 

And 

Screen Shot 2020-08-26 at 5 08 03 PM

The book detail page

kristojorg:
@hokei I understand wanting to keep this from bloating too much. Is it possible instead of the full blown modal solution to just add space between the buttons so they're visually separated? You can use a `Stack` component to do that easily.

.config/config.txt
```diff
@@ -0,0 +1,3 @@
+brazoria|https://simplye.amigos.org/txbrazoria
```
kristojorg, line 1: This should be in `/config` instead of `/.config`, and then it will be gitignored

config.txt
```diff
@@ -0,0 +1,3 @@
+brazoria|https://simplye.amigos.org/txbrazoria
```
kristojorg, line 1: This file should be removed

src/dataflow/getLibraryData.ts
```diff
@@ -28,12 +28,11 @@ async function setupCache() {
     );
   }
   if (REGISTRY_BASE) {
-    console.log("Running with Library Registry at: ", REGISTRY_BASE);
     return new LibraryDataCache(REGISTRY_BASE, CACHE_EXPIRATION_SECONDS);
   }
 
   if (CONFIG_FILE) {
-    console.log("Running with Config file at: ", CONFIG_FILE);
+    console.log("CONFIG_FILE", CONFIG_FILE);
```
kristojorg, line 35: Sorry if my previous comment wasn't clear... I think all these console logs should be left as they were personally.

src/utils/book.tsx
```diff
@@ -3,12 +3,17 @@ import {
   BookData,
   FetchErrorData,
   BookMedium,
-  MediaLink
+  MediaLink,
+  FulfillmentLink
```
kristojorg, line 7: Check the lint warnings in this file

src/utils/reader.ts
kristojorg: Okay, we can leave it if it needs to be

hokei:
@kristojorg Oh  yes I can definitely do that!

src/components/WebpubViewer.tsx
```diff
@@ -1,19 +1,56 @@
...
 import useLibraryContext from "./context/LibraryContext";
+import { useActions } from "opds-web-client/lib/components/context/ActionsContext";
+
+const initializeReader = async (
+  entryUrl,
+  catalogName,
+  useDecryptor: boolean,
+  fetcher?
+) => {
+  let decryptorParams = undefined;
```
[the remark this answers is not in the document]
kristojorg: Still think it would be more readable with an early return

src/components/WebpubViewer.tsx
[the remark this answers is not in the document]
kristojorg: Should remove this window check

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import { jsx } from "theme-ui";
+import { jsx, Link } from "theme-ui";
```
kristojorg, line 2: We don't use this `Link` component

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -154,12 +173,29 @@ const FulfillmentContent: React.FC<{
...
+    </Button>
+  );
+};
+
+const CTAPanel: React.FC<{
+  title: string;
+  subtitle: React.ReactNode;
+  buttons?: JSX.Element[];
+}> = ({ title, subtitle, buttons }) => {
+  const { errorMsg } = useBorrow();
```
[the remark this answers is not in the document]
kristojorg: I think each button should render the error message separately (this goes with their separate `useBorrow` instances

kristojorg (changes requested):
Okay just a couple more things and then I think this is ready : )

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -131,7 +133,7 @@ describe("available to borrow book", () => {
     });
     expect(
       utils.queryByText("10 out of 13 copies available.")
-    ).not.toBeInTheDocument();
+    ).toBeInTheDocument();
```
hokei, line 136: this (and L209) happened because of the moving the error checking to be per-button, which means it's no longer possible to check for the error outside the button.  FulfillmentCard doesn't check for errors before showing the availability string or borrow status, so, I think, actually represents more consistent behavior.

src/utils/__tests__/reader.test.ts
M0nica: I think I'd be okay with that assuming the test tickets will get picked up in this sprint?

src/components/__tests__/BookListItem.test.tsx
M0nica: If this test is passing it looks like it shows both the availability and the error message in cases where there is an error? If so, I think this is fine unless we end up making UX changes in the future.

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -204,13 +206,82 @@ describe("ready to borrow book", () => {
     });
     expect(
       utils.queryByText("You can now borrow this book!")
```
M0nica, line 208: This seems wonky to me that when the response is a 403 error it is able to find the text "You can now borrow this book!". I think that it should _not_ appear and it seems to be a breaking change if it does.

src/dataflow/getLibraryData.ts
```diff
@@ -28,7 +28,6 @@ async function setupCache() {
     );
   }
   if (REGISTRY_BASE) {
-    console.log("Running with Library Registry at: ", REGISTRY_BASE);
```
M0nica, line 31: I think these `console.logs` would be helpful to retain.

src/components/__tests__/BookListItem.test.tsx
kristojorg: I think it's somewhat normal actually. Consider a "Submit" button on a form. If the form gives you an error, it'll display the error but still allow you to click submit again, which will again show you the error. So I think it's okay that you can still try and borrow the book after an error occurred, so long as the error message is also displayed. But idk, both perspectives seem valid to be

src/components/__tests__/BookListItem.test.tsx
kristojorg: This seems okay with me. One thing I wonder is that with multiple borrow buttons wouldn't there actually be _multiple availabilities_? 🤯  I don't know if that's the case or not, but if so then we should open a ticket to fix that once the data fetching is refactored and the internal data model more closely represents the actual OPDS Feed

hokei:
@kristojorg I am having an issue with the `useBorrow` error message logic.  It seems like, `state.book.error` isn't specific to any loan, but to all loans.  This means that, when you have an error for one book, it shows underneath all books.  This is what it looks like on the `beta` branch 
Screen Shot 2020-08-27 at 7 41 27 PM

Is there a ticket already for this?  Since it's in the beta branch, should I try address it here?  

I think @M0nica 's outstanding question can't really be addressed until this is addressed because I think this means that there currently isn't a way to query "is there a problem with this exact book loan"?

src/dataflow/getLibraryData.ts
hokei: aaaaaaaahhh I went overly delete-happy again.  I'm sorry.

src/components/__tests__/BookListItem.test.tsx
hokei: I think this UI needs a real good re-think.  There isn't multiple availabilities (because if you borrow a book to read online, you can't borrow the same book to read on your device.  Is this intended?  I don't know?)  But, I think a better UI would be one where you first select the format you want and then get presented with a borrow button, so you never see more than one borrow button on the same book.

kristojorg:
Yea I think we should make a ticket for this. It's a symptom of how redux is used in opds-web-client. I'm currently working on further updates that will separate CPW from opds-web-client completely and remove redux, so this should no longer happen once my updates are finished. Let's just leave it

src/components/__tests__/BookListItem.test.tsx
kristojorg: Yea that makes sense, but also the single borrow button we used before led to multiple formats available for download. So yea I think we need to make a separate ticket for this

hokei:
> led to multiple formats available for download. So yea I think we need to make a separate ticket for this

Oh no I guess we're gonna need to look into the response logic 

I filed 
https://jira.nypl.org/browse/SIMPLY-3021 and https://jira.nypl.org/browse/SIMPLY-3020 for the other issues

kristojorg:
> Oh no I think that means the response logic isn't the same for downloads vs not-downloads aaahhhh

It's worse even because one borrow button can currently lead to multiple download + read online options, and then the other one leads to something else I guess? So thinking about this more carefully, I don't think the labels "Borrow to read online" and "Borrow to read on a mobile device" are really accurate... I don't know what they should be because I don't know what the two different borrow urls are actually meant to represent

CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
 ## CHANGELOG
 
+### Unreleased Changes
+
+- Feature: Add support for Clever auth providers.
+- Redesign: in Auth Modal display AuthProvider logos as selectable buttons when there are 2-4 AuthProviders instead of Combobox
+
```
M0nica: Should this mention anything about the AxisNow functionality/dependency being added?

kristojorg:
Okay I got this to run, but can anyone give me a login so I can try the reading logic?

.env
```diff
@@ -11,4 +11,10 @@ CACHE_EXPIRATION_SECONDS=86400
...
```
kristojorg, line 17: My one thought here is that since we are using the `NEXT_PUBLIC_` prefix with the `NEXT_PUBLIC_COMPANION_APP` variable, I think we should do it with this one as well. It means we don't have to declare it in `next.config.js` as next.js will automatically expose it on the frontend (without `NEXT_PUBLIC_` or declaring it in `next.config.js` it is only available on the backend for security). 

What do you think? Should we switch this to that format or should we switch the companion app to not use it? I can see arguments for both. In the past we never had it, so the old env vars don't use it. It looks a little weird to someone not in the know who is setting those env vars, but at the same time it removes a source of bugs if we forget to declare an env var in `next.config.js`...

src/hooks/useBorrow.ts
```diff
@@ -15,10 +15,10 @@ export default function useBorrow(book: BookData) {
       : undefined;
   const { actions, dispatch } = useActions();
 
-  const borrowOrReserve = async () => {
-    if (book.borrowUrl) {
+  const borrowOrReserve = async (url: string | undefined) => {
```
kristojorg, line 18: This doesn't need to go in to this PR, but we can make this function only take a `string` and then you don't have to throw the error if there is no url. You just deal with it higher up and probably don't display the button

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -8,23 +8,19 @@ import {
...
 import Stack from "components/Stack";
 import { Text } from "components/Text";
 import { MediumIcon } from "components/MediumIndicator";
 import SvgExternalLink from "icons/ExternalOpen";
 import SvgDownload from "icons/Download";
 import SvgPhone from "icons/Phone";
 import useIsBorrowed from "hooks/useIsBorrowed";
-import { NEXT_PUBLIC_COMPANION_APP } from "../../utils/env";
+import { NEXT_PUBLIC_COMPANION_APP, AXIS_NOW_DECRYPT } from "../../utils/env";
+import { BorrowOrReserve } from "components/BookList";
```
kristojorg, line 23: This is a small thing but this component shouldn't be in the `BookList` anymore if we are using it this way. Could you move it out into `/components`?

src/components/BookList.tsx
```diff
@@ -327,3 +316,31 @@ export const LanesView: React.FC<{ lanes: LaneData[] }> = ({ lanes }) => {
...
+  const buttonLabel = isBorrow
+    ? borrowLink.indirectType ===
+      "application/vnd.librarysimplified.axisnow+json"
+      ? "Borrow to read online"
+      : "Borrow to read on a mobile device"
+    : "Reserve";
+  return (
+    <>
+      <Button
+        size="lg"
```
kristojorg, line 336: These buttons look pretty large to me on the page, is there a reason they need to be so big?

src/components/BookList.tsx
```diff
@@ -327,3 +316,31 @@ export const LanesView: React.FC<{ lanes: LaneData[] }> = ({ lanes }) => {
...
+  borrowLink: FulfillmentLink;
+}> = ({ book, isBorrow, borrowLink }) => {
+  const { isLoading, borrowOrReserve, errorMsg } = useBorrow(book);
+  const loadingText = isBorrow ? "Borrowing..." : "Reserving...";
+  const buttonLabel = isBorrow
+    ? borrowLink.indirectType ===
+      "application/vnd.librarysimplified.axisnow+json"
+      ? "Borrow to read online"
+      : "Borrow to read on a mobile device"
+    : "Reserve";
```
kristojorg, line 332: Again a small thing for now, but I think that this should be moved in to `useBorrow` as it's logic that is necessary any time we are letting people borrow.

kristojorg (commented):
Few comments, none of them blocking because everything does work, but I think these would be good to get in to a follow on if possible.

kristojorg:
I think the main thing that needs to be fixed in a follow on PR is the logic around "Borrow to read online" vs "Borrow to read on mobile device". We need to move to a solution that we discussed with Leonard before this goes in to production, because that language is confusing and isn't accurate

.env
hokei: I'd go with the one that removes bugs, so I'll prefix NEXT_PUBLIC to this.  I think generally the person doing the configuration just rolls with stuff.

src/components/BookList.tsx
hokei: To do this I'd have to pass `isBorrow` and `borrowLink` into `useBorrow` -- is that standard? I thought hooks were supposed to pretty straightforward?

src/utils/env.ts
```diff
@@ -19,7 +19,8 @@ export const CACHE_EXPIRATION_SECONDS = parseInt(
   process.env.CACHE_EXPIRATION_SECONDS ?? "10",
   10
 );
-export const AXIS_NOW_DECRYPT = process.env.AXIS_NOW_DECRYPT;
+export const NEXT_PUBLIC_AXIS_NOW_DECRYPT =
+  process.env.NEXT_PUBLIC_AXIS_NOW_DECRYPT;
```
kristojorg, line 23: I would take the opportunity to make this a boolean right here, that way you don't have to check `=== "true"` or `!!NEXT_PUBLIC_AXIS_NOW_DECRYPT`, which I don't think works the way you would expect it to (I think it would pass if it was `"false"`...

src/components/WebpubViewer.tsx
```diff
@@ -52,7 +52,12 @@ const BookPage = () => {
   const { fetcher } = useActions();
 
   React.useEffect(() => {
-    initializeReader(`${bookUrl}`, catalogName, !!AXIS_NOW_DECRYPT, fetcher);
+    initializeReader(
+      `${bookUrl}`,
+      catalogName,
+      !!NEXT_PUBLIC_AXIS_NOW_DECRYPT,
```
kristojorg, line 58: This check will currently pass if it is `"false"`, so I would make the variable a boolean in `/utils/env`

kristojorg (changes requested):
Okay just the one change on the env variable that I think is actually important because it won't work the way we expect currently (I think?). Other than that this looks ready to me!

src/components/BookList.tsx
kristojorg: Yea that's standard! you can def pass variables in to hooks

kristojorg (approved):
I think this is ready for an approval provided we can fix the CI issues and that we can have a follow on PR refining the "Borrow to read on Mobile" and "Borrow to read online" logic based on our conversation with Leonard. 

Please update the CHANGELOG before merging as well.

PR #96, opened 2020-07-17
Add env variable to toggle SimplyE branding in footer

M0nica:
Currently in circulation-patron-web the default branding logic sprinkles Simply-E references throughout the app. This branding should be controlled by an environment variable of the type companion-app = "simplye" | "openebooks".

Requirements (to be confirmed)

When environment variable is "simplye"

 -   Footer (display references to download SimplyE)

When environment variable is "openebooks"

-    Footer (remove eferences to download SimplyE)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Set one of the following environment variables when running the application:
 
 The following environment variables can also be set to further configure the application.
 
-- Set `COMPANION_APP=openebooks` to display OpenE Books Branding (default displays SimplyE Branding).
+- Set `NEXT_PUBLIC_COMPANION_APP=openebooks` to display OpenE Books Branding (default displays SimplyE Branding).
 - Set `SHORTEN_URLS=false` to stop the app from removing common parts of the circulation manager URLs from the web app's URLs.
 - Set `CACHE_EXPIRATION_SECONDS` to control how often the app will check for changes to registry entries and circ manager authentication documents.
 - Set `AXE_TEST=true` to run the application with `react-axe` enabled (only works when `NODE_ENV` is "development").
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Set one of the following environment variables when running the application:
 
 The following environment variables can also be set to further configure the application.
 
+- Set `COMPANION_APP=openebooks` to display OpenE Books Branding (default displays SimplyE Branding).
 - Set `SHORTEN_URLS=false` to stop the app from removing common parts of the circulation manager URLs from the web app's URLs.
 - Set `CACHE_EXPIRATION_SECONDS` to control how often the app will check for changes to registry entries and circ manager authentication documents.
 - Set `AXE_TEST=true` to run the application with `react-axe` enabled (only works when `NODE_ENV` is "development").
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,5 @@
 const config = {
   env: {
-    NEXT_PUBLIC_COMPANION_APP: process.env.NEXT_PUBLIC_COMPANION_APP,
     SIMPLIFIED_CATALOG_BASE: process.env.SIMPLIFIED_CATALOG_BASE,
     SHORTEN_URLS: process.env.SHORTEN_URLS,
     CONFIG_FILE: process.env.CONFIG_FILE,
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,6 @@
 const config = {
   env: {
-    COMPANION_APP: process.env.COMPANION_APP || "simplye",
+    COMPANION_APP:  process.env.COMPANION_APP === "openebooks" ? "openebooks" : "simplye";
     SIMPLIFIED_CATALOG_BASE: process.env.SIMPLIFIED_CATALOG_BASE,
     SHORTEN_URLS: process.env.SHORTEN_URLS,
     CONFIG_FILE: process.env.CONFIG_FILE,
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,6 @@
 const config = {
   env: {
-    COMPANION_APP: process.env.COMPANION_APP,
+    COMPANION_APP: process.env.COMPANION_APP || "simplye",
     SIMPLIFIED_CATALOG_BASE: process.env.SIMPLIFIED_CATALOG_BASE,
     SHORTEN_URLS: process.env.SHORTEN_URLS,
     CONFIG_FILE: process.env.CONFIG_FILE,
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,6 @@
 const config = {
   env: {
-    COMPANION_APP:  process.env.COMPANION_APP === "openebooks" ? "openebooks" : "simplye";
+    COMPANION_APP:  process.env.COMPANION_APP === "openebooks" ? "openebooks" : "simplye",
     SIMPLIFIED_CATALOG_BASE: process.env.SIMPLIFIED_CATALOG_BASE,
     SHORTEN_URLS: process.env.SHORTEN_URLS,
     CONFIG_FILE: process.env.CONFIG_FILE,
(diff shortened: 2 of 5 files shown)


src/components/Footer.tsx
```diff
@@ -97,24 +97,28 @@ const Footer: React.FC<{ className?: string }> = ({ className }) => {
...
-        <Text>
-          Our mobile app lets you browse, borrow and read from our whole
-          collection of eBooks and Audiobooks right on your phone!
-        </Text>
-        <div sx={{ width: "75%", overflow: "hidden", ml: -3 }}>
-          <IosBadge sx={{ p: 3, pb: 0 }} />
-          <GooglePlayBadge />
-        </div>
-      </div>
+      {process.env.COMPANION_APP === "simplye" && <DownloadSimplyECallout />}
```
M0nica, line 100: The tests refused to pass unless I accessed `process.env` directly in the boolean check. Even moving it to a constant in the file caused the tests to fail despite the behavior in the browser is the same. I'm not sure if this is a quirk/bug/feature of how Next handles environment variables: https://nextjs.org/docs/api-reference/next.config.js/environment-variables

src/utils/env.ts
```diff
@@ -2,6 +2,7 @@
 /**
  * Simply exporting processed env vars
  */
+
```
M0nica, line 5: Setting and importing the process.env from this file worked in the browser but not for unit tests when manually setting .env _or_ mocking this file (since conflicting mocks are required) 🤔

hokei:
I'm super stumped on the test thing, being unfamiliar both with Jest and with Next.  @EdwinGuzman do you have thoughts?

hokei:
The rest of it looks good to me, just that baffling thing about the tests ...

README.md
```diff
@@ -74,6 +74,7 @@ Set one of the following environment variables when running the application:
 
 The following environment variables can also be set to further configure the application.
 
+- Set `COMPANION_APP=openebooks` to display OpenE Books Branding (default displays SimplyE Branding).
```
EdwinGuzman, line 77: I was going to suggest that this be something like "... to display a branding, such as the OpenE Books branding ..." to allow multiple brandings. And in `next.config.js`, you'd just have `COMPANION_APP:  process.env.COMPANION_APP`. But you'd also set the variable in the `.env` file. 

But then I noticed that thread in slack and that you _don't_ need to think about more available branding options right now so this is not a big deal.

M0nica: Yeah, I think for now since we only want to support two modes more explicit is better but it should be straightforward to refactor for expansion in the future since it isn't set as a boolean value even though there are only two options.

next.config.js
```diff
@@ -1,5 +1,6 @@
 const config = {
   env: {
+    COMPANION_APP:  process.env.COMPANION_APP === "openebooks" ? "openebooks" : "simplye";
```
EdwinGuzman, line 3: The end should be a comma and not a semicolon.

src/components/Footer.tsx
EdwinGuzman: It seems to be how Next handles environment variables. I vaguely remember having to use the whole `process.env` prefix when I wanted an environment variable using webpack and its define plugin. The page says the same so I think it's fine to leave `process.env` in there.

EdwinGuzman:
I think this is fine. I haven't used an environment variable for client-side react components, only for server-side work, and from the docs this seems to be the way to access and manipulate it in tests.

Also, if this environment variable worked like the rest, I'm not sure how you would test it since in the test you mount and test whereas in next, it bundles and transforms the variables into values.

src/utils/env.ts
kristojorg: Setting the value here should allow you to pretty simply mock this `env.ts` module in your tests rather than having to `resetModules`, but either of those mocking strategies should be fine

src/components/__tests__/Footer.test.tsx
```diff
@@ -45,39 +45,81 @@ test("shows external links when present in state w/ apropriate attributes", () =
...
-  const googleBadge = utils.getByRole("link", {
-    name: /get simplye on the google play store/i
+
+  test("does not show simplyE callout when COMPANION_APP is 'openebooks'", () => {
+    process.env.COMPANION_APP = "openebooks";
+
+    const utils = render(<Footer />);
+
+    expect(process.env.COMPANION_APP).toBe("openebooks");
+    expect(utils.queryByText(/download simplye/i)).not.toBeInTheDocument();
```
kristojorg, line 59: I've never used this exact setup and something tells me it's not actually testing whether it's in the document. Just to be sure, have you gotten this to fail? I always used `.toBeNull` with `queryBy` and I kind of remember it being because this didn't work. Mind just making sure this works?

EdwinGuzman: I'm currently introducing `react-testing-library` into an NYPL app I'm working on so I'm reading and getting deeper into it. It seems like this is the right pattern according to a few posts, https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence unless I'm missing something?

kristojorg: Yep you're right : )

src/components/__tests__/Footer.test.tsx
```diff
@@ -45,39 +46,84 @@ test("shows external links when present in state w/ apropriate attributes", () =
...
-    "https://apps.apple.com/us/app/simplye/id1046583900"
-  );
-
-  const googleBadge = utils.getByRole("link", {
-    name: /get simplye on the google play store/i
+describe("toggling SimplyE Branding", () => {
+  test("does not show simplyE callout when NEXT_PUBLIC_COMPANION_APP is 'openebooks'", () => {
+    /* eslint-disable */
+    // @ts-ignore
+    env.NEXT_PUBLIC_COMPANION_APP = "openebooks";
```
M0nica, line 53: I was running into issues when mocking `../../utils/env` since it's returning constants vs. functions that return constants and functions are a lot easier to mock multiple times in jest. I was able to mock the value of NEXT_PUBLIC_COMPANION_APP once the jest way but then it seemed impossible to clear the mock or override the value the next time it needed to be accessed. I manually changed the value that `../../utils/env`  returns which required disabling rules 🤔  I'm not sure if there was another mock approach you had in mind.

next.config.js
```diff
@@ -1,6 +1,6 @@
 const config = {
   env: {
-    COMPANION_APP:  process.env.COMPANION_APP === "openebooks" ? "openebooks" : "simplye",
+    NEXT_PUBLIC_COMPANION_APP: process.env.NEXT_PUBLIC_COMPANION_APP,
```
kristojorg, line 3: This is no longer necessary when using `NEXT_PUBLIC_` prefix. Those are automatically exposed to browser so we don't have to explicitly tell next.js to do that here.

src/components/__tests__/Footer.test.tsx
kristojorg: Hmmmm yea I think you're right it's a bit tougher with constants. This should work though:

```
beforeEach(() => jest.resetModules());

test("something", () => {
  jest.mock("./appEnv", () => ({ NEXT_PUBLIC_COMPANION_APP: "simplyE"}));
  const { NEXT_PUBLIC_COMPANION_APP } = require("env.ts");
});
```
You will have to define the whole set of env vars exported in the mock, so you'll probably want a helper function like 
```js
const fullEnv = partialEnv => ({
  /* default env vars */,
  ...partialEnv // this will override the defaults
})
// then you can just do this in the test
jest.mock("env.ts", fullEnv({NEXT_PUBLIC_COMPANION_APP: "simpleE"}));
```

kristojorg: But if the way you're doing it is working then I think you should be fine keeping it as is. Just don't ignore the ts errors (you can typecast them instead).

kristojorg:
Whoops it looks like this didn't have a version bump or changelog update in it. I'll go ahead and add one in #97 for this PR as well, but in the future let's try to include them in PRs. Was my bad for approving without catching that though.

PR #97, opened 2020-07-27
SAML Auth

kristojorg:
This PR corresponds to https://github.com/NYPL-Simplified/opds-web-client/pull/271 . It is a draft because it depends on that PR being merged first. 

I implement SAML authentication with a SAML Auth Form that simple renders a button redirecting to the SAML IdP.


kristojorg:
Okay @EdwinGuzman this is ready for review. I also added a changelog update for @M0nica's work in #96 because we forgot to include it there.

src/auth/__tests__/BasicAuthForm.test.tsx
```diff
@@ -78,11 +79,11 @@ test("sumbits", async () => {
...
 
   // act
-  const barcode = utils.getByLabelText("Barcode");
-  const pin = utils.getByLabelText("Pin");
+  const barcode = utils.getByLabelText("Barcode input");
+  const pin = utils.getByLabelText("Pin input");
   userEvent.type(barcode, "1234");
   userEvent.type(pin, "pinpin");
-  const loginButton = utils.getByText("Login");
+  const loginButton = utils.getByRole("button", { name: "Login" });
```
kristojorg, line 86: FYI if you're wondering why I update these from getByText to getByRole, it's just because the getByRole also specifies that it should be a button so it's a little more specific of a test. I'm just updating when I'm already refactoring a test.

EdwinGuzman: Yeah, I've been using `getByRole` a lot for the form I'm testing and I like its explicitness.

src/auth/SamlAuthForm.tsx
```diff
@@ -0,0 +1,37 @@
...
+      sx={{
+        p: 2,
+        pt: 4,
+        display: "flex",
+        justifyContent: "center",
+        alignItems: "center"
+      }}
+    >
+      <Button onClick={handleClick}>
+        Login with {provider.method.description ?? "Unknown IDP"}
```
EdwinGuzman, line 31: What does IDP stand for?

kristojorg: Identity Provider. This shouldn't really ever be undefined so we don't expect the fallback to be shown. I suppose I could move the check up higher and just not render anything if there is no description provided.

src/components/__tests__/ListView.test.tsx
```diff
@@ -46,3 +57,12 @@ test("truncates authors", () => {
   expect(utils.getByText("one, two & 3 more"));
   expect(utils.queryByText("one, two, three")).toBeFalsy();
 });
+
+test("displays loader", () => {
```
EdwinGuzman, line 61: Is this directly related to the PR or just adding another test?

kristojorg: Yea just added another test... I was looking at coverage and just decided to add tests for a few uncovered areas while I was at it

kristojorg:
Yep I'll send you a CM and credentials in a DM on slack

kristojorg:
That test site I sent you was actually a couple commits old. Console logs are gone now : )

PR #100, opened 2020-07-27
add link to previous page

M0nica:
Screen Shot 2020-07-27 at 5 46 47 PM

- add link to navigate back to previous page
- if there is no link to previous page then navigate to host (is there a better fallback in this case? In particular I am concerned if someone does directly to the reader via URL (as opposed to from within the app) in the multi-library context and hits the back arrow)
- I tried passing the router variables from WebpubViewer -> Reader but they seemed to be getting overridden.

### DIFF
--- a/src/utils/reader.ts
+++ b/src/utils/reader.ts
@@ -56,8 +56,8 @@ function initBookSettings(element, webpubManifestUrl, catalogName) {
     prefix: "webpub-viewer"
   });
   const upLink = {
-    url: new URL("https://github.com/NYPL-Simplified/webpub-viewer"),
-    label: catalogName,
+    url: new URL(document.referrer || window.location.origin),
+    label: `Return to ${catalogName}`,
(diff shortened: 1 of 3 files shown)


src/utils/reader.ts
```diff
@@ -56,8 +56,8 @@ function initBookSettings(element, webpubManifestUrl, catalogName) {
     prefix: "webpub-viewer"
   });
   const upLink = {
-    url: new URL("https://github.com/NYPL-Simplified/webpub-viewer"),
-    label: catalogName,
+    url: new URL(document.referrer || window.location.origin),
+    label: `Return to ${catalogName}`,
     ariaLabel: "Go back to the Github repository"
```
kristojorg, line 61: I think we should change aria label to match the label yea?

M0nica: Yes, good catch! Thanks.

M0nica:
I'm going to merge into addEReader so they can go out together.

PR #103, opened 2020-07-28
update image fallback

kristojorg:
This PR updates the BookCover component so that it shows the medium icon while the image is loading and keeps it there in the case of an error. When the book finishes loading we now fade it in to view.

### DIFF
--- a/src/css-overrides.css
+++ b/src/css-overrides.css
@@ -1,0 +1,4 @@
+/* this is to fix a problem in NYPL Design System setting global svg styles */
+svg {
+  width: initial;
+}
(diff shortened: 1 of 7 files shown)


src/css-overrides.css
```diff
@@ -0,0 +1,4 @@
+/* this is to fix a problem in NYPL Design System setting global svg styles */
```
EdwinGuzman, line 1: Can you also flag this in the Design System repo? Or have you mentioned this to Crystal?

kristojorg: I raised an issue for it here: https://github.com/NYPL/nypl-design-system/issues/295

kristojorg:
Yea not a bad idea to lazy load images, especially since they aren't always very small. We should maybe raise a ticket for that in Jira or somewhere?

PR #106, opened 2020-07-30
Implement color theme

kristojorg:
This PR reads the `web_color_scheme` from the auth document and uses that to customize the theme-ui `theme` object, thus enabling the color customization in the CM.


kristojorg:
That's correct in a way. We no longer use the `foreground` and `background` colors, and old CMs will no longer be able to set those values. However, in the case of an old CM, we do have fallback values we use, so the app will still render, it just won't be themed. 

This is actually intended because the old color values were enforced via an a11y check to contrast well with each other, whereas the new values need to both contrast well with white. So values for the old system wouldn't make sense in this design and shouldn't be used.

kristojorg:
I just went through and made sure the types indicated that the new `primary` and `secondary` values are optional so that we stay compatible with old CMs.

kristojorg:
Both the CM and CPW are backwards compatible with this change. The CM will continue emitting a `foreground` and `background` color even if updated so the interface shows `primary` and `secondary`. CPW will only accept `primary` and `secondary`, but will fall back if they are missing to defaults - which is better than using the previous `foreground` and `background` colors that don't make sense in this context.

PR #108, opened 2020-07-30
omit current book from appearing in recommendations

M0nica:
Currently in circulation patron web we are recommending other books by the same author on the book details page. On the client-side we should filter out the current book from appearing as a recommended book.

Screen Shot 2020-07-30 at 11 49 14 AM


kristojorg:
Actually it won't be under `UNRELEASED CHANGES` but under the most recent version number (it hasn't been released yet)

PR #109, opened 2020-07-31
Fetch loans only on app startup

kristojorg:
This resolves a bug in CPW caused by the 30 minute cache of loans data instituted by the server. Previously, we refetched the loans data quite often (relying on the browser cache to avoid network requests). This meant that after borrowing a book, we would then fetch loans and get stale information. To rectify this, OWC is set up to modify the redux loans state with the return value of a borrow request. 

What I've done here is to mimic the mobile apps, and only fetch loans once when the app starts, and then we simply maintain our internal state from there on out. This means we can become out of sync with the server, but will become eventually consistent (after the 30 minute cache expires). 

This PR also updates the url we fetch for loans from a manually created one to the one present in `state.loans.url`. This should resolve an issue I saw being discussed on slack @M0nica @leonardr .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,13 @@
 ## CHANGELOG
 
 # 2.3.0
+
 - Feature: Add support for SAML auth providers.
 - Feature: Add `NEXT_PUBLIC_COMPANION_APP` env var (in PR #97) to toggle the display of SimplyE branding. Value can be either `"simplye"` or `"openebooks"`. 
 - Fix: Don't show search bar if collection does not include search data.
 - Fix: Fetch loans on page load when auth credentials are detected so that if you navigate directly to a book you have checked out, it will properly show checked out state.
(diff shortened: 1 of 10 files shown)


kristojorg:
It means we will be okay without it for now, but the mobile client and the web client will become out of sync because the data is cached on the client side not server side. This will be especially problematic when someone uses the web app to borrow a book, and then is strongly directed to read that book in SimplyE, where the book may no show up for 30 minutes. 

I think we should continue discussing a solution that involves backend caching instead of frontend caching.

leonardr:
OK, I'll continue my work. It won't be too bad since we will immediately invalidate the server-side cache if a loan transaction happens.

kristojorg:
I suppose I misspoke. I'm not against client-side caching, I think that's the right solution, but I think there needs to be a mechanism for revalidation (ie a `Last-Modified` header). That way the client side can cache the data and feel free to revalidate it however often they want

PR #110, opened 2020-07-31
Filter platform incompatible links

kristojorg:
This PR filters download links to only show ones that are platform compatible. Currently the only check we are performing is to not show ACSM files for mac users, as there is no way to read them on mac.


kristojorg:
Ah I probably could write one actually

kristojorg:
Okay I added a couple tests

kristojorg:
After talking to James, it looks like there is indeed a version of Adobe Digital Editions that will work on mac. Closing.

PR #111, opened 2020-08-04
update signIn to showAuth form when "Sign In" is clicked

M0nica:
This is an attempt to resolve issue where "log in" button in the navigation seemed to be working when logging in for the first time but was not functional after logging out when attempting to log back in again.

1. Run circulation-patron-web locally with SIMPLIFIED_CATALOG_BASE=http://qa.circulation.openebooks.us/USOEI/groups/ SHORTEN_URLS=false npm run dev
2. Go to index page: http://localhost:3000/
3. Click on log in the navigation
4. Login Modal appears and Log in using First Book credentials
5. Log Out
6. Click Log in in navigation

Expected: Login modal appears and allow First Book or Clever credentials.
Actual: Nothing happens

Note: Currently if you attempt to check out a book on the book details page then you are prompted to login which supports First Book login. The nav login seems to be working when logging in for the first time but not after logging out and then attempting to log back in again.

### DIFF
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -27,21 +27,22 @@ function useAuth() {
     return state.loans.url;
   });
 
-  const authTitle = useTypedSelector(state => state.auth.title) || "";
+  const auth = useTypedSelector(state => state.auth);
 
   const noop = () => ({});
 
-  const authCallback = useTypedSelector(state => state.auth.callback) || noop;
-
-  const authCancel = useTypedSelector(state => state.auth.cancel) || noop;
-
-  const authProviders = useTypedSelector(state => state.auth.providers) || [];
+  const { title, callback, cancel, providers } = auth;
 
   const signIn = () =>
     loansUrl &&
-    authProviders &&
+    providers &&
     dispatch(
-      actions.showAuthForm(authCallback, authCancel, authProviders, authTitle)
+      actions.showAuthForm(
+        callback || noop,
+        cancel || noop,
+        providers,
+        title || ""
+      )
     ) &&
     dispatch(actions.fetchLoans(loansUrl));
 
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -23,8 +23,27 @@ function useAuth() {
     router.push(link.href, link.as);
   };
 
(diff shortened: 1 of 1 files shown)


src/hooks/useAuth.ts
```diff
@@ -23,8 +23,27 @@ function useAuth() {
...
+
+  const authTitle = useTypedSelector(state => state.auth.title) || "";
+
+  const noop = () => ({});
+
+  const authCallback = useTypedSelector(state => state.auth.callback) || noop;
+
+  const authCancel = useTypedSelector(state => state.auth.cancel) || noop;
+
+  const authProviders = useTypedSelector(state => state.auth.providers) || [];
```
EdwinGuzman, line 38: I can verify this PR works and fixes the bug, but I'm not sure if this is the best way to get all the variables you need. Can you do `const authTitle = useTypedSelector(state => state.auth);` and then destructure the values you need instead of calling `useTypeSelector` multiple times?

M0nica: I updated to destruct the values from `const auth` instead of having multiple selectors. However, when I set default fallback values the types seemed to think that the variables could be undefined even though I explicitly set fallback values. I ended up setting the fallback values within ` actions.showAuthForm()` but would prefer to set the fallbacks outside of that method if there's a way to do that without typescript complaining.

EdwinGuzman: I don't know which is the better pattern and I didn't mean for my question to be something to implement. I just wasn't sure but if the other way is better and allows you to set the fallback outside of the method, then it may be better to revert. Sorry! Either way, the bug is fixed so approving but you can update this part again.

M0nica: I will merge this now but may refactor in my current Clever auth work

kristojorg:
So this issue is actually caused by a browser cache I believe. If you check the network tabs, I _think_ you'll see that clicking "Sign in" the second time actually sent a new request to the loans endpoint which resulted in a 200. I attempted to resolve this in opds-web-client but may have introduced a regression when implementing SAML auth.

kristojorg:
This might solve the problem of showing the sign in form, but the underlying issue of having the user still be seen as "logged in" despite having logged out I believe will still remain

PR #113, opened 2020-08-07
Add Ability to Login Via Clever Auth

M0nica:
- add https certificates from @hokei's work in https://github.com/NYPL-Simplified/circulation-patron-web/pull/93. The app should now be run at https://localhost:3000 (not the https)

Expected: You are able to login using either First Book or Clever credentials when an OPDS feed's authentication document has both providers as authentication options. Previously, Clever was not supported. 

1. Run circulation-patron-web locally with SIMPLIFIED_CATALOG_BASE=[OPEN E BOOKS QA SERVER] SHORTEN_URLS=false npm run dev
2. Go to index page: https://localhost:3001/
3. Click on a book
4. Click button to check out book
5. See modal to login
 

To-Do:
- Style Clever button a bit (there will be follow up work to get the styling for all auth providers to show logo instead of text) 

Temporary styling:

Screen Shot 2020-08-12 at 2 27 22 PM
Screen Shot 2020-08-12 at 2 27 27 PM


src/components/Auth.tsx
```diff
@@ -82,6 +84,10 @@ const Auth: React.FC = ({ children }) => {
           )}
           {authProvider && authProvider.plugin.formComponent ? (
             <authProvider.plugin.formComponent provider={authProvider} />
+          ) : authProvider &&
+            authProvider.plugin.buttonComponent &&
+            authProvider.method.description === "Clever" ? (
+            <authProvider.plugin.buttonComponent provider={authProvider} />
           ) : (
```
EdwinGuzman, line 91: This is fine but I think nested ternary operations are hard to read and this could be simplified, but not a big deal.

src/auth/cleverAuthButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const currentUrl = window.location.origin + window.location.pathname;
+  let authUrl;
+
+  for (const link of props?.provider?.method.links || []) {
+    if (link.rel === "authenticate") {
+      authUrl =
+        link.href +
+        "&redirect_uri=" +
+        encodeURIComponent(encodeURIComponent(currentUrl));
```
EdwinGuzman, line 33: Why do you need to encode this twice?

src/hooks/useAuth.ts
```diff
@@ -51,19 +54,48 @@ function useAuth() {
    * query param
    */
   const { access_token: samlAccessToken } = router.query;
+
+  const cleverAccessToken =
+    typeof window !== "undefined" && CleverAuthPlugin.lookForCredentials();
+
+  const accessToken = samlAccessToken || cleverAccessToken;
```
EdwinGuzman, line 61: What happens if there's both SAML and Clever auth in place? Won't only the SAML access token be found since both have the `access_token` query in the URL and here the `SAMLAccessToken` is first? The other is also different in line 69 but I'm not sure that makes a difference - it just seems inconsistent.

Also, how is the error case being handled when `CleverAuthPlugin.lookForCredentials` returns a problem detail rather than an object with `credentials`?

src/auth/cleverAuthButton.tsx
M0nica: I don't think the second encoding is necessary. I just removed it.

src/components/Auth.tsx
```diff
@@ -80,11 +91,17 @@ const Auth: React.FC = ({ children }) => {
               </Select>
             </div>
           )}
-          {authProvider && authProvider.plugin.formComponent ? (
+
+          {authProvider && authProvider.plugin.formComponent && (
```
M0nica, line 95: @EdwinGuzman Do you think this approach is more readable than the ternaries? For some reason TypeScript complained if I used `showFormComponent` directly instead of spelling out both conditions. It would say "JSX element type 'authProvider.plugin.formComponent' does not have any construct or call signatures.ts(2604)". I am not sure if something is off with the TS configuration as I'd expect it to be able to infer.

src/hooks/useAuth.ts
M0nica: They aren't interchangeable. I _think_  the SAML URL post authentication has `access_token` whereas the Clever url has `#access_token` and`CleverAuthPlugin.lookForCredentials()` specifically looks for hash value in URL. I tried logging in via SAML and Clever. When I logged in via Clever I was able to see only the Clever access token was set (in the below useEffect) and that SAML access token was undefined. When I logged in via SAML both Clever and SAML access token were undefined...but when I checked the local data the SAML access token was being saved so some part of the SAML authentication is happening outside of the useEffect/URL change as well.

M0nica: If there's an error I believe it will set the credentials to be {credentials: "", provider: ""}...but maybe that needs to be explicit. When those values are set to empty string it will return the login screen, the user will still be logged out but it doesn't display an error. 

``` const { credentials: cleverCredentials } = cleverAccessToken || {
    credentials: { credentials: "", provider: "" }
  };
```
Should `accessToken` check to make sure that cleverAccessToken is returning credentials and not an error when. 

`const accessToken = samlAccessToken || (cleverAccessToken.credentials && cleverAccessToken);`

`const accessToken = samlAccessToken || (!cleverAccessToken.error && cleverAccessToken);`

All of the logic in the useEffect is fated by `accessToken` evaluating to true.

src/components/Auth.tsx
EdwinGuzman: That is odd... yes this is fine. I don't want to make this more complicated than it should be if Typescript is complaining.

src/hooks/useAuth.ts
EdwinGuzman:
```
const { credentials: cleverCredentials } = cleverAccessToken || {
    credentials: { credentials: "", provider: "" }
};
```

This makes me think that if `cleverAccessToken` is `{ error: ... }`, then `credentials` will be undefined. The `||` should take effect when `cleverAccessToken` is empty or undefined so I'm not sure if you'll get the empty empty with empty string values if there's an error. But double check and maybe a test can help with this.

hokei: Is there a way to make something like this clear in tests?  I realize it might be hard to mock, but having the "what is supposed to happen" encoded in some way will be useful.

src/auth/BasicAuthForm.tsx
```diff
@@ -10,15 +10,24 @@ import { generateCredentials } from "opds-web-client/lib/utils/auth";
...
+  href: string;
+}
+
 type FormData = {
   login: string;
   password: string;
 };
 
+export interface CPWBasicAuthMethod extends BasicAuthMethod {
+  links?: AuthLink[];
```
M0nica, line 24: It might make more sense to add `links?` as an optional field in BasicAuthMethod directly in opds-web-client. https://github.com/NYPL-Simplified/opds-web-client/blob/66d644158151971dddecb03c3e8f9fc57b97c36a/packages/opds-web-client/src/interfaces.ts#L214

src/hooks/__tests__/useAuth.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  afterAll(() => {
+    window.location = location;
+  });
+
+  const mockSAMLToken = "potatokey";
+  const mockRouterWithEmptyQuery = { query: {} };
+  const mockRouterWithSAMLToken = {
+    query: {
+      //eslint-disable-next-line camelcase, @typescript-eslint/camelcase
```
M0nica, line 17: Going to see if this should be ignored or if the linting rule can be ignored.

src/hooks/useAuth.ts
M0nica: I extracted the fetching of credentials into a function `getCredentials()` that has some unit tests. I don't think there  should ever be a case where both types of access tokens exists at the same time in the url since one is `#access_token` and one is `access_token` so I didn't explicitly test for that scenario.

M0nica: I extracted the fetching of credentials into a function `getCredentials()` that has some unit tests. I don't think there  should ever be a case where both types of access tokens exists at the same time in the url since one is `#access_token` and one is `access_token` so I didn't explicitly test for that scenario. Although, if for some reason both did exist how would we want the app to handle that?

src/auth/cleverAuthButton.tsx
```diff
@@ -13,7 +13,7 @@ export interface CleverAuthMethod extends AuthMethod {
   links?: AuthLink[];
 }
 
-const CleverButton: React.FC<AuthButtonProps<CleverAuthMethod>> = ({
+export const CleverButton: React.FC<AuthButtonProps<CleverAuthMethod>> = ({
```
EdwinGuzman, line 16: This component is already exported as the default. Do you need it exported like this, too?

src/hooks/__tests__/useAuth.test.ts
EdwinGuzman: I'm okay with this way of doing it.

src/hooks/useAuth.ts
```diff
@@ -7,6 +7,36 @@ import { CleverAuthPlugin } from "../auth/cleverAuthPlugin";
...
 
+const TOKEN_NOT_FOUND = "TOKEN_NOT_FOUND";
+
+export function getTokenFromUrl(
+  router
+): {
+  token:
+    | string
+    | { credentials?: string; undefined; error?: string | undefined }
+    | undefined;
```
EdwinGuzman: Can you define a variable name for this type?

src/hooks/useAuth.ts
```diff
@@ -7,6 +7,36 @@ import { CleverAuthPlugin } from "../auth/cleverAuthPlugin";
...
+  type: string;
+} {
+  const cleverAccessToken =
+    typeof window !== "undefined" && CleverAuthPlugin.lookForCredentials();
+
+  const { access_token: samlAccessToken } = router.query;
+
+  let type = TOKEN_NOT_FOUND;
+  /* there should never be a case where both types of access tokens exists at the same time */
```
EdwinGuzman, line 24: For another PR, maybe throw an error if this case happens.

src/auth/BasicAuthForm.tsx
EdwinGuzman: Hmm yeah. Can be a separate PR because this extension looks good.

src/hooks/useAuth.ts
EdwinGuzman: I didn't see the question before but I commented that the function that checks for either or can throw an error if both are found.

hokei:
Screen Shot 2020-08-12 at 2 19 43 PM

I know that the styling is temp but is this intended?

M0nica:
> Screen Shot 2020-08-12 at 2 19 43 PM
> 
> I know that the styling is temp but is this intended?

I believe that happens when the QA server returns the invalid IP address :( since it's not able to resolve the image URL that is in the authentication document.

hokei:
ok yeah let's leave that for when we figure out the ip issue

M0nica:
> Will this work if the "SHORTEN_URL" env variable is set to true?

Yes, the double encoding works in both SHORTEN_URL configurations

PR #115, opened 2020-08-17
Display login logos instead of combo box when there are 2-4 AuthProviders

M0nica:
- Display login logos instead of combo box when there are fewer than 5 AuthProviders
- When there are 2-4 AuthProviders show Auth buttons instead 

GIF of login with buttons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 ## CHANGELOG
 
+### Unreleased Changes
+
+- Feature: Add support for Clever auth providers.
+- Redesign: in Auth Modal display AuthProvider logos as selectable buttons when there are 2-4 AuthProviders instead of Combobox
+
 # 2.3.0
 
 - Feature: Add support for SAML auth providers.
 - Feature: Add confirmation modal to the sign out button.
-- Feature: Add `NEXT_PUBLIC_COMPANION_APP` env var (in PR #97) to toggle the display of SimplyE branding. Value can be either `"simplye"` or `"openebooks"`. 
+- Feature: Add `NEXT_PUBLIC_COMPANION_APP` env var (in PR #97) to toggle the display of SimplyE branding. Value can be either `"simplye"` or `"openebooks"`.
 - Fix: Don't show search bar if collection does not include search data.
 - Fix: Fetch loans on page load when auth credentials are detected so that if you navigate directly to a book you have checked out, it will properly show checked out state.
 - Fix: Make BookCover show a better image fallback on image load failure. Also show medium icon on load, and fade image in once it is done loading.
--- a/src/auth/BasicAuthButton.tsx
+++ b/src/auth/BasicAuthButton.tsx
@@ -1,17 +1,9 @@
 /** @jsx jsx */
 import { jsx } from "theme-ui";
-import * as React from "react";
 import Button from "components/Button";
 
 const BasicAuthButton = props => {
-  let imageUrl;
-
-  for (const link of props.links || []) {
-    if (link.rel === "logo") {
-      imageUrl = link.href;
-      break;
-    }
-  }
+  const imageUrl = props.links.find(link => link.rel === "logo")?.href;
 
   return (
     <Button
--- a/src/auth/BasicAuthButton.tsx
+++ b/src/auth/BasicAuthButton.tsx
@@ -23,6 +23,7 @@ const BasicAuthButton = props => {
         flex: "1 0 auto",
         width: "280px",
         height: "51px",
+        backgroundSize: `280px 51px`,
         backgroundRepeat: "no-repeat",
         backgroundPosition: "0",
         backgroundImage: `url(${imageUrl})`,
--- a/src/auth/BasicAuthButton.tsx
+++ b/src/auth/BasicAuthButton.tsx
@@ -1,0 +1,39 @@
+/** @jsx jsx */
+import { jsx } from "theme-ui";
+import * as React from "react";
+import Button from "components/Button";
+
+const BasicAuthButton = props => {
+  let imageUrl;
+
+  for (const link of props.links || []) {
+    if (link.rel === "logo") {
+      imageUrl = link.href;
+      break;
+    }
+  }
+
+  return (
+    <Button
+      type="submit"
+      sx={{
+        alignSelf: "flex-end",
+        m: 2,
+        mr: 0,
+        flex: "1 0 auto",
+        width: "280px",
+        height: "51px",
+        backgroundRepeat: "no-repeat",
+        backgroundPosition: "0",
+        backgroundImage: `url(${imageUrl})`,
+        cursor: "pointer",
+        border: "none"
+      }}
+      {...props}
+    >
+      {!imageUrl ? "Login" : ""}
+    </Button>
+  );
+};
+
+export default BasicAuthButton;
--- a/src/components/Auth.tsx
+++ b/src/components/Auth.tsx
@@ -20,14 +20,16 @@ import { AuthProvider, AuthMethod } from "opds-web-client/lib/interfaces";
  *  - uses the AuthPlugin system to render the auth form
  */
 
-function shouldShowButton(authProvider): boolean {
+function shouldShowButton(authProvider: AuthProvider<AuthMethod>): boolean {
   return Boolean(
     authProvider?.plugin?.buttonComponent &&
       authProvider?.method?.description === "Clever"
   );
 }
 
-function shouldShowFormComponent(authProvider): boolean {
+function shouldShowFormComponent(
+  authProvider: AuthProvider<AuthMethod>
+): boolean {
   return Boolean(authProvider.plugin.formComponent);
 }
 const Auth: React.FC = ({ children }) => {
@@ -64,7 +66,7 @@ const Auth: React.FC = ({ children }) => {
       setAuthProvider(providers?.[0]);
   }, [authProvider, providers]);
 
-  const handleChangeProvider = e => {
+  const handleChangeProvider: React.ChangeEventHandler<HTMLSelectElement> = e => {
     setAuthProvider(
       providers?.find(provider => provider.id === e.target.value) ||
         ([][0] as AuthProvider<AuthMethod>)
@@ -127,7 +129,7 @@ const Auth: React.FC = ({ children }) => {
                     <provider.plugin.buttonComponent
                       key={`${provider.id}${idx}`}
                       provider={provider}
-                      onClick={handleChangeProvider}
+                      onClick={handleChangeProvider as () => void}
                     />
                   )}
                 </>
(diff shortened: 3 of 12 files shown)


src/auth/BasicAuthButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const BasicAuthButton = props => {
+  let imageUrl;
+
+  for (const link of props.links || []) {
+    if (link.rel === "logo") {
+      imageUrl = link.href;
+      break;
+    }
+  }
```
kristojorg, line 14: This isn't a huge deal in this case, but for the future I try to steer clear of declaring let variables as it becomes difficult for developers to track their value when code becomes complicated. I personally would change this to:
```
const imageUrl = props.links.find(link => link.rel === "logo").href
```

src/auth/BasicAuthButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <Button
+      type="submit"
+      sx={{
+        alignSelf: "flex-end",
+        m: 2,
+        mr: 0,
```
kristojorg, line 22: I assume this is so that you can have multiple in a list? Does this work if it breaks from a horizontal to a vertical list? If not, you might look at using the `Stack` component in whatever component is consuming this one. That one is supposed to space children evenly and allow horizontal to vertical wrapping

M0nica: The AuthButton is currently only used in the `<ul>`  in the Auth modal so it should always be displayed as a stacked horizontal list. I'll look into replacing the `<ul>` with the `Stack` component.

src/components/Auth.tsx
```diff
@@ -7,22 +7,38 @@ import useAuth from "../hooks/useAuth";
...
+function shouldShowButton(authProvider): boolean {
+  return Boolean(
+    authProvider?.plugin?.buttonComponent &&
+      authProvider?.method?.description === "Clever"
+  );
+}
+
+function shouldShowFormComponent(authProvider): boolean {
+  return Boolean(authProvider.plugin.formComponent);
+}
```
kristojorg, line 33: Can you type the arguments to these functions?

src/auth/BasicAuthButton.tsx
```diff
@@ -0,0 +1,32 @@
+/** @jsx jsx */
+import { jsx } from "theme-ui";
+import Button from "components/Button";
+
+const BasicAuthButton = props => {
```
kristojorg, line 5: Can you type this component please?

src/auth/BasicAuthButton.tsx
kristojorg: Within a stack I think you shouldn't need to add these custom spacings to make them line up

src/auth/BasicAuthButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <Button
+      type="submit"
+      sx={{
+        alignSelf: "flex-end",
+        m: 2,
+        mr: 0,
+        flex: "1 0 auto",
+        width: "280px",
+        height: "51px",
+        backgroundSize: `280px 51px`,
```
kristojorg, line 18: Are we doing this so that the button is the same height as the clever background image? If so, maybe we should extract this style and apply it to all of the buttons on this form so they're all of uniform size?

src/components/Auth.tsx
```diff
@@ -7,22 +7,38 @@ import useAuth from "../hooks/useAuth";
...
 const Auth: React.FC = ({ children }) => {
   const { showForm, cancel, providers } = useAuth();
 
   const dialog = useDialogState();
   const library = useLibraryContext();
-  const [authProvider, setAuthProvider] = React.useState(providers?.[0]);
+  const [authProvider, setAuthProvider] = React.useState(
+    [][0] as AuthProvider<AuthMethod>
```
kristojorg, line 40: I don't really understand what's going on here. What is the shape of this state meant to be now?

src/components/Auth.tsx
```diff
@@ -45,25 +61,37 @@ const Auth: React.FC = ({ children }) => {
...
+    if (!authProvider && providers?.length === 1)
+      setAuthProvider(providers?.[0]);
   }, [authProvider, providers]);
 
-  const handleChangeProvider: React.ChangeEventHandler<HTMLSelectElement> = e => {
+  const handleChangeProvider = (e: { target: { value: string } }) => {
     setAuthProvider(
-      providers?.find(provider => provider.id === e.target.value)
+      providers?.find(provider => provider.id === e.target.value) ||
+        ([][0] as AuthProvider<AuthMethod>)
```
kristojorg, line 71: Why is it now necessary to cast this everywhere. Casting is generally an escape hatch or to be used with values from outside our typescript environment (like when making a network request and receiving some value).

src/components/Auth.tsx
```diff
@@ -92,14 +121,72 @@ const Auth: React.FC = ({ children }) => {
             </div>
           )}
 
+          {showProviderButtons && !authProvider && (
+            <div sx={{ mb: 2, textAlign: `center` }}>
```
kristojorg, line 125: You can apply this styling directly to the stack and get rid of the wrapping div I believe

src/components/Auth.tsx
```diff
@@ -92,14 +121,72 @@ const Auth: React.FC = ({ children }) => {
...
+              <Stack
+                direction="column"
+                sx={{
+                  padding: 0,
+                  margin: 0,
+                  listStyleType: "none"
```
kristojorg, line 131: Are these doing anything? I think the Stack has these set by default already

src/components/Auth.tsx
```diff
@@ -92,14 +121,72 @@ const Auth: React.FC = ({ children }) => {
...
+                        <BasicAuthButton
+                          links={provider.method.links || []}
+                          aria-label={`Login to ${provider.method.description}`}
+                          key={provider.id}
+                          value={provider.id}
+                          onClick={handleChangeProvider}
+                        >
+                          {provider.method.description}
+                        </BasicAuthButton>
+                      )}
```
kristojorg, line 147: Is it possible to just define this as a `buttonComponent` in the `basicAuthProvider` and then you can just render the component and pass in the provider just like you're doing for clever. That way we aren't doing custom checks for different provider types and they all work the same way

src/auth/BasicAuthButton.tsx
M0nica: Yes, these measurements are so that it matches the size of the buttons' background images. I will look into making sure all of the buttons on the form are a uniform size.

src/components/Auth.tsx
M0nica: The default state of authProvider is being changed to be no provider instead of defaulting to the first provider. Although, if there is only 1 provider we'd want the default state to be that 1 provider which isn't explicitly handled at the moment but I  need to see if that is causing some of the issues you were seeing with button alignment. I am going to try to reproduce your above screenshot.

src/components/Auth.tsx
M0nica: in the case where there are no providers I wanted to cast it to an empty provider. I was running into a lot of other type issues when the provider was explicitly undefined but I can look into if there's another way to approach those type issues.

src/components/Auth.tsx
kristojorg: Okay so maybe just set it to `undefined` then yea? And set the type to be a provider or undefined:
```
React.useState<AuthProvider<AuthMethod> | undefined>(undefined);
```

src/components/Auth.tsx
kristojorg: See the comment I made above. Hopefully that helps?

src/auth/BasicAuthButton.tsx
kristojorg: Thanks : )

kristojorg: Maybe extract it into an `<AuthFormButton` that has those styles and then use that one for all of them?

hokei:
If you hit "cancel" on this modal, it doesn't look like hitting "sign in" brings it up again (or brings you to a sign in page).  It will show up again if you hit "refresh"

M0nica:
> This looks great in the gif! I think it still needs some massaging in some cases though. This is what I get with a single SAML provider:
> 
> image
> 
> I think buttons should be aligned and the same size, and the first login button should have tab focus initially I think

I was able to reproduce some misalignment (although it looked different than the screenshot you shared) when the only provider option was Clever. I was then able to get the button to center in that case. Can you check if the new `ModalButtonStyles` I've applied to all of the buttons in the modal appear to resolve the alignment issue you were seeing? If not, I might need help reproducing in a SAML environment. 

Screen Shot 2020-08-20 at 3 52 17 PM

src/components/Auth.tsx
M0nica: I was running into a lot of issues when re-typing the state value without casting 🤔

src/components/Auth.tsx
```diff
@@ -92,14 +121,42 @@ const Auth: React.FC = ({ children }) => {
...
 
+          {showProviderButtons && !authProvider && (
+            <Stack direction="column">
+              {providers?.map((provider, idx) => (
+                <>
+                  {provider.plugin.buttonComponent && (
+                    <provider.plugin.buttonComponent
+                      key={`${provider.id}${idx}`}
+                      provider={provider}
+                      onClick={handleChangeProvider as () => void}
```
kristojorg, line 132: The error this was giving you before casting it was actually pointing to a valid runtime error. The onClick method here was not guaranteed to provide an event object, and especially in this situation would not be a Select Event. Take a look at how I've changed it to see a safer way to approach this than casting

kristojorg:
I'm making a commit to remove the type casting, but I'm still seeing style issues. It looks like it has to do with the length of the library name making the modal wider than the one you're using in development:

image

src/components/Modal.tsx
```diff
@@ -3,6 +3,19 @@ import { jsx } from "theme-ui";
 import * as React from "react";
 import { DialogStateReturn, DialogBackdrop, Dialog } from "reakit";
 
+export const ModalButtonStyles = {
```
kristojorg, line 6: This shouldn't be capitalized. We only capitalize components and classes and types

kristojorg (changes requested):
just this capitalization, the button styles in the modal still need some massaging with long library names, and also check the code with the updates that I pushed

M0nica:
I updated the styling to make sure the buttons are aligned with longer library names (manually changed the library name)
Screen Shot 2020-08-26 at 2 09 26 PM
Screen Shot 2020-08-26 at 2 09 22 PM
Screen Shot 2020-08-26 at 2 09 08 PM
Screen Shot 2020-08-26 at 2 09 04 PM

kristojorg (approved):
looks good!

CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
 ## CHANGELOG
 
+### Unreleased Changes
+
+- Feature: Add support for Clever auth providers.
```
M0nica, line 5: Clever auth was added in a previous PR but wasn't reflected in the CHANGELOG.

PR #117, opened 2020-08-18
update React aXe configuration

M0nica:
- There were some type issues with React-Axe in circulation-patron-web and it needed some tweaks to be functional again. 
- In terms of testing I confirmed that I am now able to see the aXe logs when running `yarn dev:axe`

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -15,6 +15,9 @@ const config = {
     // Perform customizations to webpack config
     // Important: return the modified config
     !isServer && config.plugins.push(new webpack.IgnorePlugin(/jsdom$/));
+    // react-axe should only be bundled when REACT_AXE=true
+    !process.env.REACT_AXE &&
+      config.plugins.push(new webpack.IgnorePlugin(/react-axe$/));
     // Fixes dependency on "fs" module.
     // we don't (and can't) depend on this in client-side code.
     if (!isServer) {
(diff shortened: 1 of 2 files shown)


kristojorg:
Thanks! 

The one thing I want to be 100% sure of with the AXE related stuff is that it doesn't get bundled and shipped to the client unless REACT_AXE=true. Would you testing this out for me with these changes:

1. Run `npm run build`. Note the "First load JS share by all" size.
2. Comment out the import code and run again. See if the size jumped. 

AXE is huge and just want to make sure we don't ship it in production.

next.config.js
```diff
@@ -15,6 +15,9 @@ const config = {
     // Perform customizations to webpack config
     // Important: return the modified config
     !isServer && config.plugins.push(new webpack.IgnorePlugin(/jsdom$/));
+    // react-axe should only be bundled when REACT_AXE=true
+    !process.env.REACT_AXE &&
```
M0nica, line 19: @kristojorg  The First Load JS shared by all with react-axe _was_ 406 kB. I made an update to the webpack config in `next.config.js` to ignore react-axe unless REACT_AXE=true so now the build bundle size is  278 kB.

kristojorg: Amazing thank you!!

PR #118, opened 2020-08-20
Refactor fetching of library data

kristojorg:
This is a refactor preparing for future updates. It does a few things:

1. Adds OPDS 2.0 types. These aren't final and are still be developed as needed. A lot is left to be done especially around links. I'd like to have all the link relations and media types explicitly typed in the application.
2. Remove the `LibraryDataCache`. This means on app load we will be fetching an auth document every time. It also means we will be fetching from the registry base on every app load if running with a registry base. Ultimately I would like pages to be statically generated, meaning that these things will only be fetched once at build time, not on every app load. Next.js will allow that, but it's for the future. 
3. Refactor the process by which we set up the application. I also add a lot of tests for this functionality which previously didn't exist, and I added types to the whole process. 
4. Add some custom errors: `ApplicationError`, `UnimplementedError`, `AppSetupError` and `PageNotFoundError`. I would like to start using these and throwing errors when it makes sense. We can then catch them with error boundaries and render more helpful error messages. 
5. Move our app-wide data fetching from `_app.getInitialProps` to per-page `getServerSideProps`. This sets us up in the future to fetch data server side that we need to construct pages, such as the book data or the feed data. Ultimately we can move to `getStaticProps` to have the pages be statically rendered, which will reduce server load for CPW and for the CM while also making the app much faster.


src/interfaces.ts
```diff
@@ -12,76 +93,62 @@ export interface ComplaintData {
...
-  onlyLibrary?: boolean;
+  slug: string | null;
   catalogUrl: string;
   catalogName: string;
-  logoUrl?: string;
-  colors?: {
-    primary?: string;
-    secondary?: string;
+  logoUrl: string | null;
+  colors: {
```
M0nica, line 128: Is `colors` optional or will all Libraries have `colors` key even if the properties within it are `null`?

kristojorg: Maybe you're right. I changed this because Next.js now constructs this on the server and then serializes it to send it to the client, but it can't serialize `undefined`. So I made it `null` instead. But maybe it makes more sense that either the whole object is defined or the whole object is `null`.

kristojorg: I've changed it to be null or fully defined

kristojorg:
If you guys would mind testing this with your local setups and config files that would be helpful @M0nica @hokei @EdwinGuzman

src/dataflow/getLibraryData.ts
```diff
@@ -1,158 +1,221 @@
...
+        return { ...links, about: link };
+      case "alternate":
+        return { ...links, libraryWebsite: link };
+      case "privacy-policy":
+        return { ...links, privacyPolicy: link };
+      case "terms-of-service":
+        return { ...links, tos: link };
+      case "help":
+        if (link.type === HTMLMediaType) return { ...links, helpWebsite: link };
+        return { ...links, helpEmail: link };
```
M0nica, line 191: I believe the help `rel` link  can be either a website, e-mail or telephone number. It looks like it will treat a phone number like a helpEmail.

src/dataflow/getLibraryData.ts
```diff
@@ -1,158 +1,221 @@
...
+      case "register":
```
M0nica, line 192: Should register and logo just return `links`? I thought returning nothing for certain cases in a reduce could be dangerous and cause the end result to be undefined.

M0nica:
The last time I checked out this branch with the SIMPLIFIED_CATALOG_BASE explicitly set it worked...but when I pull the latest , install dependencies and then run the app with SIMPLIFIED_CATALOG_BASE explicitly set I get a 404 on the index page and the below error:

```
Module parse failed: Unexpected token (8:9)
File was processed with these loaders:
 * ./node_modules/next/dist/build/webpack/loaders/next-babel-loader.js
You may need an additional loader to handle the result of these loaders.
|  * an OPDS 2 Feed of OPDS 2 Catalogs from which we extract the catalog root url
|  */
> export * as OPDS2 from "types/opds2";
| /**
|  * OPDS1.2 DATA TYPES
```

src/dataflow/getLibraryData.ts
kristojorg: Interesting... I stole this logic from the previous implementation. Do you know if there is a way to test if it is an email or phone number without using a regex on the `href`? Do they have different media types?

src/dataflow/getLibraryData.ts
kristojorg: This syntax means that all of "register", "logo" and "navigation" will execute the same block (return links). I did this in addition to the default case just to be explicit with all the `rel` values the CPW type system expects.

EdwinGuzman:
> Ultimately I would like pages to be statically generated, meaning that these things will only be fetched once at build time, not on every app load.

This makes sense but OPDS feeds can change often if there are new books, lanes, or lists. If pages are statically rendered with one fetch from the server, how will subsequent updates be fetched? E.g. a new opds feed a week after the initial app is statically rendered?

EdwinGuzman:
When I run the app, every external CM works except for my own libraries - so it's probably some bad configs. But, the errors could be fixed perhaps to show an empty library rather than an error:

Screen Shot 2020-08-27 at 3 21 13 PM

src/errors.ts
```diff
@@ -0,0 +1,35 @@
+export default class ApplicationError extends Error {
+  readonly statusCode?: number = undefined;
+
+  constructor(m: string, baseError?: Error) {
+    super(`${m}${baseError ? `\nBase Error:\n${baseError.message}` : ""}`);
+    Object.setPrototypeOf(this, ApplicationError.prototype);
```
EdwinGuzman, line 6: I'm not sure why you need to use `setPrototypeOf`. You're also creating and adding the prototype every time a new instance of `ApplictionError` is invoked, no? Wouldn't it be better to assign the prototype directly outside the function so it only adds it once?

src/interfaces.ts
```diff
@@ -1,7 +1,89 @@
...
-import { CollectionData, BookData } from "opds-web-client/lib/interfaces";
+import {
+  CollectionData,
+  BookData,
+  MediaType
+} from "opds-web-client/lib/interfaces";
+
+/**
+ * OPDS 2.0 DATA TYPES
+ * Currently only used for support of a Library Regristry, which is
```
EdwinGuzman, line 11: Small typo in "Regristry".

src/interfaces.ts
```diff
@@ -1,7 +1,89 @@
...
+/**
+ * OPDS1.2 DATA TYPES
+ */
+export const AuthDocLinkRelation = "http://opds-spec.org/auth/document";
+export type OPDSLinkRelation =
+  | typeof AuthDocLinkRelation
+  | AuthDocLinkRelations
+  | "related";
+
+export type OPDSLinkRole = string;
```
EdwinGuzman, line 26: Does this _need_ to be a separate variable? I don't see an issue with it but just curious why you made the decision for this.

src/interfaces.ts
```diff
@@ -1,7 +1,89 @@
...
+export interface AuthDocumentLink extends Omit<OPDSLink, "role"> {
+  rel: AuthDocLinkRelations;
+}
+
+export interface OPDSAuthProvider {}
+
+export interface Announcement {
+  id: string;
+  content: string;
+}
```
EdwinGuzman, line 66: You're not using this in the PR but it's declared. Is this because you're expecting this from the CM? I know it was just recently implemented but it'll be good to see implemented in the web clients besides the mobile clients.

src/pages/[library]/index.tsx
```diff
@@ -1,23 +1,19 @@
...
 
-import Collection from "../../components/Collection";
-import Layout from "../../components/Layout";
-
-import ErrorPage from "../404";
-import useLibraryContext from "components/context/LibraryContext";
-
-const LibraryHome = ({ statusCode }: { statusCode?: number }) => {
-  const libraryTitle = useLibraryContext().catalogName;
+type Props = AppProps;
```
Comment (EdwinGuzman) [inline src/pages/[library]/index.tsx:7]: Is this just convention? I noticed this in `src/pages/[library]/collection/[collectionUrl].tsx` but not in `src/pages/[library]/book/[bookUrl].tsx`. Let's just use one way of setting the prop type.

src/test-utils/fixtures/opds2.ts
```diff
@@ -0,0 +1,45 @@
+import { OPDS2 } from "interfaces";
+/* eslint-disable camelcase */
+/* eslint-disable @typescript-eslint/camelcase */
```
EdwinGuzman, line 3: Not a big deal but can we set up a convention for this? I typically see the `eslint-disable` comment as the first line in a file and usually all the rules are on one line.

EdwinGuzman:
> Move our app-wide data fetching from _app.getInitialProps to per-page getServerSideProps. 

I thought that if you get data for the app from the start on the index page, the rest of the pages would still render on the server-side but you just need to pass the data to those pages. Do you _need_ to implement the `getServerSideProps` on every page for it to be ssr?

src/dataflow/getConfigFile.ts
```diff
@@ -0,0 +1,43 @@
...
+    const response = await fetch(configFileUrl);
+    const text = await response.text();
+    const parsed = parseConfigText(text);
+    return parsed;
+  } catch (e) {
+    throw new Error("Could not fetch config file at: " + configFileUrl);
+  }
+}
+
+function parseConfigText(raw: string): AppConfigFile {
```
EdwinGuzman, line 31: Can you add some comments to the functions to explain what they're doing?

src/dataflow/__tests__/getConfigFile.test.tsx
```diff
@@ -0,0 +1,62 @@
+import { existsSync, readFileSync } from "fs";
+import { AppSetupError } from "errors";
+import getConfigFile from "dataflow/getConfigFile";
+import fetchMock from "jest-fetch-mock";
+
+jest.mock("fs");
+const mockedExistsSync = existsSync as jest.MockedFunction<typeof existsSync>;
```
EdwinGuzman, line 7: Is this the same as `jest.fn<typeof existsSync>()`?

src/dataflow/getLibraryData.ts
```diff
@@ -1,158 +1,221 @@
...
-      configText = fs.readFileSync(CONFIG_FILE, "utf8");
-    }
-    // read the entries.
-    for (const entry of configText.split("\n")) {
-      if (entry && entry.charAt(0) !== "#") {
-        const [path, circManagerUrl] = entry.split("|");
-        config[path] = circManagerUrl;
-      }
+/**
+ * Returns a function to construct a registry catalog ink, which leads to a
```
EdwinGuzman, line 34: Small typo: `ink`.

src/dataflow/getLibraryData.ts
```diff
@@ -1,158 +1,221 @@
...
+    if (!catalogRootUrl)
+      throw new ApplicationError(
+        `CatalogEntry did not contain a Catalog Root Url. Library UUID: ${librarySlug}`
+      );
+    return catalogRootUrl;
   }
+  throw new AppSetupError(
+    "One of CONFIG_FILE, REGISTRY_BASE, or SIMPLIFIED_CATALOG_BASE must be defined."
+  );
+}
```
EdwinGuzman, line 129: While testing, when I run the app with no environment variables, I don't see this error thrown anywhere. The main page, localhost:3000 does render this:

Screen Shot 2020-08-27 at 5 57 57 PM

src/dataflow/__tests__/getLibraryData.test.tsx
```diff
@@ -0,0 +1,388 @@
...
+
+  test("Throws AppSetupError if no env var is defined", async () => {
+    setEnv({});
+    const promise = getCatalogRootUrl("hello");
+    expect(promise).rejects.toThrowError(AppSetupError);
+    expect(promise).rejects.toThrowErrorMatchingInlineSnapshot(
+      `"One of CONFIG_FILE, REGISTRY_BASE, or SIMPLIFIED_CATALOG_BASE must be defined."`
+    );
+  });
+  describe("with Registry Base", () => {
```
EdwinGuzman, line 111: By the way, the older env variable was `LIBRARY_REGISTRY` and it still appears in the README. Can you update that reference to `REGISTRY_BASE`?

kristojorg:
@EdwinGuzman 

>This makes sense but OPDS feeds can change often if there are new books, lanes, or lists. If pages are statically rendered with one fetch from the server, how will subsequent updates be fetched? E.g. a new opds feed a week after the initial app is statically rendered?

I oversimplified actually. They would be rendered likely at request time and then basically cached that way for some period of time, so they would eventually sync up with CM feed changes. Alternatively, we could make the CM emit a webhook whenever the feed changes which would then trigger an invalidation of all the static pages

src/errors.ts
kristojorg: This has to do with the way that native Errors are extended being different from a typical class. It allows us to later check whether an error is an instance of a particular error. I'll admit I'm not 100% sure how this works, but this was the recommended method I found from multiple sources online. How would you suggest assigning the prototype externally? I think the instance (`this`) needs to exist before it can be assigned. I don't imagine it's costly to do that though.

src/interfaces.ts
kristojorg: I was going to start specifying all the possible roles here, but decided to leave that for later, so for now it's just a string.

src/interfaces.ts
kristojorg: Yea basically I am making an effort to type out the server response as completely as possible

src/test-utils/fixtures/opds2.ts
kristojorg: Yep good point. I think I'm going to disable one of those eslint rules globally so we don't have a duplicate, and then yea I'll move to the first line.

kristojorg:
> I thought that if you get data for the app from the start on the index page, the rest of the pages would still render on the server-side but you just need to pass the data to those pages. Do you need to implement the getServerSideProps on every page for it to be ssr?

This is true when using `getInitialProps`, but I am switching to use `getServerSideProps` which ultimately simplifies the code because it _never_ runs on the client, meaning we can read files and etc without having to check if we are on the client, whereas `getInitialProps` runs both on the server and on the client. 

However, I may be able to use `getInitialProps` in the `_app` along with `getServerSideProps` in the pages, which would actually be better anyway. Let me check on that.

src/errors.ts
EdwinGuzman: I looked it up and saw that it was the recommended way of doing it, except it was typically the class being set rather than `this` as the first argument. It might make a difference but I don't know enough about it. I agree it's not costly.

src/dataflow/__tests__/getConfigFile.test.tsx
kristojorg: Yes except that it's a module mock, so code that imports from `fs` will get the mocked version instead of the real version by using `jest.mock("fs")`

src/types/opds2.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface NavigationLink extends Link {
+  title: string;
+}
+
+type GroupMetadata = { title: string };
+export interface Group extends Collection<GroupMetadata> {
+  /**
+   * For when a catalog has more than one collection of
+   * navigation or publications. It contains an array of
```
EdwinGuzman, line 35: I think you're missing the end of this sentence.

src/dataflow/__tests__/getLibraryData.test.tsx
kristojorg: Weirdddd, I don't know how or when this env var got changed. I'll try to figure that out, I don't think it should be changed from what it always was

EdwinGuzman: Agreed. I don't remember it changing but just one should be used.

EdwinGuzman:
Code looks good with minor comments here and there. I didn't review the OPDS 2 interface too closely since it's a file that will be updated but overall it looks good (I implemented an OPDS 2 API in python but I know I have some things mixed up and this makes it clearer).

Since @hokei and @M0nica are actively working on this repo, I'd like them to approve the PR as well when they understand the changes going on.

src/dataflow/getLibraryData.ts
M0nica: I saw this draft for authentication in opds 1.0 that mentions telephone number but am having trouble finding the final spec or the 2.0 version https://drafts.opds.io/authentication-for-opds-1.0.html. I don't see a type specific to phone numbers and am not sure exactly what the feed would look like when it returns a number so the only approach I can think of at the moment would involve RegEx  🤔.

M0nica:
> The last time I checked out this branch with the SIMPLIFIED_CATALOG_BASE explicitly set it worked...but when I pull the latest , install dependencies and then run the app with SIMPLIFIED_CATALOG_BASE explicitly set I get a 404 on the index page and the below error:
> 
> ```
> Module parse failed: Unexpected token (8:9)
> File was processed with these loaders:
>  * ./node_modules/next/dist/build/webpack/loaders/next-babel-loader.js
> You may need an additional loader to handle the result of these loaders.
> |  * an OPDS 2 Feed of OPDS 2 Catalogs from which we extract the catalog root url
> |  */
> > export * as OPDS2 from "types/opds2";
> | /**
> |  * OPDS1.2 DATA TYPES
> ```

Overall the logic looks good to me but I am not sure why now I am running into node_module issues when trying to run the app with `SIMPLIFIED_CATALOG_BASE=https://qa-circulation.openebooks.us/USOEI/groups SHORTEN_URLS=true REACT_AXE=true npm run dev:https` 

 I tried removing all packages and re-installing and am still running into this issue with the latest changes for this branch,

src/dataflow/__tests__/getLibraryData.test.tsx
kristojorg: Actually it looks like maybe it was always `REGISTRY_BASE`? This is what is in the readme: https://github.com/NYPL-Simplified/circulation-patron-web#application-startup-configurations

kristojorg:
> There's also an issue when the app runs with multiple configs.

I've solved this by adding an error if there are multiple configs.

kristojorg:
@M0nica I'm not really sure what is causing the error you're seeing. Can you check if vscode (or whatever editor) is using a local or global version of typescript, and which version that would be?

src/dataflow/getLibraryData.ts
kristojorg: Have resolved this. Should now be properly erroring.

kristojorg:
@EdwinGuzman I've fixed the error that was obfuscating the actual errors in your libraries. Can you try running again and tell me if the errors you're seeing are any better?

.eslintrc.js
```diff
@@ -45,7 +44,7 @@ module.exports = {
     // redundant and doesn't function properly
     "react/prop-types": 0,
     "@typescript-eslint/no-unused-vars": [
-      "warn",
+      "error",
```
M0nica, line 47: 👍

src/dataflow/__tests__/getLibraryData.test.tsx
EdwinGuzman: `LIBRARY_REGISTRY` appears twice in the link.

EdwinGuzman:
Updates look good. I'm not too concerned about the registry environment variable but it should be updated eventually.

kristojorg:
@EdwinGuzman Ah I see what you're seeing now. I removed references to `LIBRARY_REGISTRY` in the README and replaced with `REGISTRY_BASE`

PR #119, opened 2020-08-25
a11y: add visible focus to filled and ghost buttons

M0nica:
- focus was barely visible in Firefox since it was using browser defaults instead of styling from circulation-patron-web
- updated to use the same focus style as the Reakit button example: https://reakit.io/docs/button/#main 

After (in Firefox)
- filled/img button Screen Shot 2020-08-25 at 12 24 38 PM
- ghost button Screen Shot 2020-08-25 at 12 24 19 PM

Before (in Firefox)
Screen Shot 2020-08-17 at 10 18 35 AM

### DIFF
--- a/src/components/Button/styles.ts
+++ b/src/components/Button/styles.ts
@@ -1,9 +1,6 @@
 import { SystemStyleObject } from "@styled-system/css";
 import { darken, lightness } from "@theme-ui/color";
 import { ButtonSize, ButtonVariant } from "./index";
-import theme from "../../theme/theme";
-
-const { shadows } = theme;
 
 export const sizes = {
   sm: {},
@@ -57,7 +54,7 @@ export const styleProps = (
           textDecoration: "none"
         },
         "&:focus": {
-          boxShadow: shadows.focus
+          boxShadow: "focus"
         },
         "&:active": {
           bg: darken(color, 0.1)
@@ -84,7 +81,7 @@ export const styleProps = (
           textDecoration: "none"
         },
         "&:focus": {
-          boxShadow: shadows.focus
+          boxShadow: "focus"
         },
         "&:active": {
           // bg: darken(color, 0.1)
--- a/src/components/Button/styles.ts
+++ b/src/components/Button/styles.ts
@@ -1,6 +1,9 @@
 import { SystemStyleObject } from "@styled-system/css";
 import { darken, lightness } from "@theme-ui/color";
 import { ButtonSize, ButtonVariant } from "./index";
+import theme from "../../theme/theme";
+
+const { shadows } = theme;
 
 export const sizes = {
   sm: {},
@@ -51,8 +54,10 @@ export const styleProps = (
         "&:focus,&:hover": {
           bg: darken(color, 0.05),
           color: "white",
-          textDecoration: "none",
-          boxShadow: `0 0 0 0.2em rgba(0, 109, 255, 0.4)`
+          textDecoration: "none"
+        },
+        "&:focus": {
+          boxShadow: shadows.focus
         },
         "&:active": {
           bg: darken(color, 0.1)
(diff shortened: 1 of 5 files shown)


src/components/Button/styles.ts
```diff
@@ -51,7 +51,8 @@ export const styleProps = (
         "&:focus,&:hover": {
           bg: darken(color, 0.05),
           color: "white",
-          textDecoration: "none"
+          textDecoration: "none",
+          boxShadow: `0 0 0 0.2em rgba(0, 109, 255, 0.4)`
```
M0nica, line 55: Is there a variable I can use here instead?

M0nica: Right now the focus/hover styles are coupled but I am wonder if he boxShadow outline should only apply to focus and not hover.

hokei:
If this can wait, I'd wait until https://github.com/NYPL/nypl-design-system/pull/354 gets merged, tbh.  We're pulling in DS anyway, it means we don't have to maintain our own CSS for this kind of stuff (also IE fallback, etc)

M0nica:
> If this can wait, I'd wait until [NYPL/nypl-design-system#354](https://github.com/NYPL/nypl-design-system/pull/354) gets merged, tbh. We're pulling in DS anyway, it means we don't have to maintain our own CSS for this kind of stuff (also IE fallback, etc)

We're not currently using the DS buttons yet. Is the plan to convert from Reakit to DS once #354 is merged?

src/components/Button/styles.ts
kristojorg: I think it should only apply to focus, and I would put this in the theme object and then reference it that way

kristojorg: I think you can add a second `&:hover` setting in addition to the shared `&:focus,&:hover`

kristojorg (changes requested):
I am not totally sure, but I think there will be more work to do before we can use the DS buttons. Ours seem significantly different from the buttons I saw in DS before, because they allow for variants, icons, and different buttons for different purposes (anchor buttons, nav buttons, regular buttons)

The only change here is that the shadow should only be on focus and we should move the setting into the theme object as a variable. You can find out where to put it on the theme-ui website

scooooooooby:
> If this can wait, I'd wait until [NYPL/nypl-design-system#354](https://github.com/NYPL/nypl-design-system/pull/354) gets merged, tbh. We're pulling in DS anyway, it means we don't have to maintain our own CSS for this kind of stuff (also IE fallback, etc)

As an FYI on this PR, it is applying styling directly to browser elements over the DS components (a hold-out from grabbing this from the other DS we have at NYPL). Reading through this I think it's worth asking whether the PR over there should even be doing that.

That said, it does mean that the styling would get applied here regardless or whether or not you were using DS buttons or not—if you're using semantic `<button>`s, it will be applied.

src/components/Button/styles.ts
```diff
@@ -51,8 +54,10 @@ export const styleProps = (
         "&:focus,&:hover": {
           bg: darken(color, 0.05),
           color: "white",
-          textDecoration: "none",
-          boxShadow: `0 0 0 0.2em rgba(0, 109, 255, 0.4)`
+          textDecoration: "none"
+        },
+        "&:focus": {
+          boxShadow: shadows.focus
```
kristojorg, line 60: Once you've added it to the theme, you can simply do `boxShadow: "focus"` and it will pull in the value from the theme in context

kristojorg (changes requested):
Just a small change

src/components/Button/styles.ts
M0nica: Great, thanks for sharing that syntax. I just updated it to `boxShadow: "focus"`.

kristojorg (approved):
Looks good!

PR #120, opened 2020-08-25
a11y: resolve lang and heading aXe related issues

M0nica:
> <html> element must have a lang attribute. 
This PR adds `_document.tsx` to set the HTML lang to `en` by default. 

On the book details the title is h1 but the summary is h3. This PR updates the summary to H2 with the same styles applied.
> moderate: Heading levels should only increase by one https://dequeuniversity.com/rules/axe/3.5/heading-order?application=axeAPI

### DIFF
--- a/src/components/bookDetails/index.tsx
+++ b/src/components/bookDetails/index.tsx
@@ -22,7 +22,7 @@ import ReportProblem from "./ReportProblem";
 import useTypedSelector from "../../hooks/useTypedSelector";
 import { NavButton } from "../Button";
 import Head from "next/head";
-import { H3, Text, H1 } from "components/Text";
+import { H2, Text, H1 } from "components/Text";
 import MediumIndicator from "components/MediumIndicator";
 import SimplyELogo from "components/SimplyELogo";
 import IosBadge from "components/storeBadges/IosBadge";
@@ -109,7 +109,7 @@ const Summary: React.FC<{ book: BookData; className?: string }> = ({
   className
 }) => (
   <div sx={{ my: 2 }} className={className} aria-label="Book summary">
-    <H3 sx={{ mb: 2 }}>Summary</H3>
+    <H2 sx={{ mb: 2 }}>Summary</H2>
     <div
       dangerouslySetInnerHTML={{
         __html: book.summary ?? "Summary not provided."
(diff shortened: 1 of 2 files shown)


src/pages/_document.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+class MyDocument extends Document {
+  static async getInitialProps(ctx) {
+    const initialProps = await Document.getInitialProps(ctx);
+    return { ...initialProps };
+  }
+
+  render() {
+    return (
+      <Html lang="en">
```
M0nica, line 12: _Can we just hard code set the attribute to `en` like  <html lang="en"> or does the lang need to be configurable?_

kristojorg: For now we should be fine hardcoding it

src/components/bookDetails/index.tsx
```diff
@@ -109,7 +109,7 @@ const Summary: React.FC<{ book: BookData; className?: string }> = ({
   className
 }) => (
   <div sx={{ my: 2 }} className={className} aria-label="Book summary">
-    <H3 sx={{ mb: 2 }}>Summary</H3>
+    <H2 sx={{ mb: 2 }}>Summary</H2>
```
kristojorg, line 112: I know the same margin is applied, but are the base styles for these components not different? As in, does this not change the way it looks visually?

kristojorg (approved):
So long as it doesn't change visually from before, I'm good with this. Thanks!

src/pages/_document.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from "react";
+import Document, { Html, Head, Main, NextScript } from "next/document";
+
+class MyDocument extends Document {
+  static async getInitialProps(ctx) {
+    const initialProps = await Document.getInitialProps(ctx);
+    return { ...initialProps };
+  }
```
kristojorg, line 8: Is this necessary if there is no _document getInitialProps?

M0nica: I removed this `getInitalProps` as it wasn't necessary.

src/components/bookDetails/index.tsx
M0nica: I ended up passing a specific variant to ensure the visual styling is consistent for the summary header

PR #122, opened 2020-08-28
Loan Error message should not be global

M0nica:
## Loan Error message should not be global 

- When there is an error in loans, it has no concept of which loan has errored, leading to an error message on all borrow buttons.

- This PR updates it so that the error message is only returned when the error url matches the book url.

The error URL for a book: https://qa-circulation.openebooks.us/USOEI/works/BOOKID looks like: https://qa-circulation.openebooks.us/USOEI/works/BOOKID/borrow/4 

Before: 
Screen Shot 2020-08-27 at 7 41 27 PM

After: 
Screen Shot 2020-08-28 at 2 52 43 PM

### DIFF
--- a/src/hooks/useBorrow.ts
+++ b/src/hooks/useBorrow.ts
@@ -8,9 +8,9 @@ export default function useBorrow(book: BookData) {
   const isUnmounted = React.useRef(false);
   const [isLoading, setLoading] = React.useState(false);
   const bookError = useTypedSelector(state => state.book?.error);
-  const errorObj = getErrorMsg(bookError);
+  const errorStr = getErrorMsg(bookError);
   const errorMsg =
-    book.url && bookError && bookError.url.startsWith(book.url) ? errorObj : "";
+    book.url && bookError && bookError.url.startsWith(book.url) ? errorStr : "";
   const { actions, dispatch } = useActions();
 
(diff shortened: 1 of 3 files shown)


src/hooks/useBorrow.ts
```diff
@@ -8,7 +8,9 @@ export default function useBorrow(book: BookData) {
   const isUnmounted = React.useRef(false);
   const [isLoading, setLoading] = React.useState(false);
   const bookError = useTypedSelector(state => state.book?.error);
-  const errorMsg = getErrorMsg(bookError);
+  const errorStr = getErrorMsg(bookError);
+  const errorMsg =
+    book.url && bookError && bookError.url.startsWith(book.url) ? errorStr : "";
```
kristojorg, line 13: I sort of feel that the errorMsg is more properly `undefined` instead of an empty string if it doesn't match the bookUrl. What do you think?

kristojorg (approved):
This is a good temporary solution given the way things work! I like it. Left one minor comment but it's up to you to take it or leave it.

src/hooks/useBorrow.ts
M0nica: `undefined` makes sense to me. I just updated the fallback value for the errorMsg accordingly.

PR #123, opened 2020-08-31
Typescript strict mode

kristojorg:
This PR enables [Typescript Strict Mode](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html#getting-stricter-checks). The main thing this does in practice in our codebase is it disallows implicit `any` types. If you want a type to be `any`, you must include a type annotation specifying such. 

This is an update in preparation for future work. It's taken a while to get here, but I think we're finally ready for strict mode here.

This is a draft because it depends on a PR from opds-web-client: https://github.com/NYPL-Simplified/opds-web-client/pull/282


package.json
```diff
@@ -39,7 +39,7 @@
     "next": "^9.5.2",
     "next-transpile-modules": "^4.0.0",
     "opds-feed-parser": "0.0.18",
-    "opds-web-client": "^0.5.4",
+    "opds-web-client": "github:NYPL-Simplified/opds-web-client#opds-web-client-v0.5.4-gitpkg",
```
kristojorg, line 42: This will be changed when a new release is made from OWC

package.json
```diff
@@ -51,15 +51,16 @@
     "request": "^2.88.2",
     "styled-system": "^5.1.2",
     "system-font-css": "^2.0.2",
-    "theme-ui": "^0.3.1",
-    "universal-url": "^1.0.1"
```
kristojorg, line 55: We also got to remove this. It's no longer necessary with Next.js and it was a pretty big bundle bloater! yay : )

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": false,
     "skipLibCheck": true,
-    "strict": false,
+    "strict": true,
```
kristojorg, line 7: 🎉

hokei:
Yay!  I like this!  I just want to ask about order, because gah I expect turning this on may put another round of re-reviews on https://github.com/NYPL-Simplified/circulation-patron-web/pull/93 and I'm kind of antsy to get that PR in.  But I also see an argument for, "well that PR should be following this anyway", which is ... strictly true, but in practice, not.

kristojorg:
No let's go ahead and merge yours first and then we can update this one with that code and make some further type refinements to get this one in

kristojorg:
Okay I think this is ready and it also upgrades us to typescript `4.0.2` so we are on the latest version

package.json
```diff
@@ -36,56 +36,58 @@
...
-    "jest": "^25.3.0",
+    "jest": "^26.4.2",
     "jest-emotion": "^10.0.32",
     "jest-fetch-mock": "^3.0.3",
-    "prettier": "2.0.5",
+    "prettier": "^2.1.1",
     "react-axe": "^3.4.1",
     "react-test-renderer": "^16.13.1",
-    "typescript": "^3.8.3"
+    "typescript": "^4.0.2"
```
kristojorg, line 90: This is a major version change but the breaking changes didn't apply to us, so I bumped us

src/components/Text.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from "react";
 
 type NativeComponent<
   T extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>,
-  P = {}
+  P = Record<string, unknown>
```
kristojorg, line 7: I was wrong in thinking that `{}` was the type of an empty object. This is the proper way to type an unknown object. Eslint now tells us this

kristojorg:
Okay this is finally ready!! woohoo

jest.config.js
```diff
@@ -156,7 +157,7 @@ module.exports = {
   ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  testPathIgnorePatterns: ["/node_modules/", "/.next/"],
+  testPathIgnorePatterns: ["/node_modules/", "/.next/", "/axisnow-access-control-web"],
```
hokei, line 160: I'm not sure if I understand what's going on here, but we are relying on CPW to run `axisnow-access-control-web` tests, so they shouldn't be skipped.

src/components/bookDetails/Recommendations.tsx
```diff
@@ -77,7 +77,7 @@ const getRelatedUrl = (book: BookData): null | string => {
   const links = book.raw.link;
   if (!links) return null;
 
-  const relatedLink = links.find(link => link.$.rel.value === "related");
+  const relatedLink = links.find((link: any) => link.$.rel.value === "related");
```
hokei, line 80: would it benefit us to define an interface so we don't have to use `any` here? 
(Why is the data returned by `link` so weird?  haha)

jest.config.js
kristojorg: Oh I didn't realize that the repo isn't tested internally, I'll re-enable. I was seeing 0% coverage on that folder so I'm not sure it is actually running them. Have you checked that?

src/components/bookDetails/Recommendations.tsx
kristojorg: Yes that'll come in my future updates. The reason this is happening is because at some point someone decided to access the "raw" values to get something that wasn't parsed by the `opds-web-client` `OPDSDataAdapter`. They should have updated that module, but instead they parsed it themselves sort of dangerously from "raw" which is an object with `any` type already. So this was already `any`, it just now is explicit about it (same with everywhere else I put `any`).

jest.config.js
hokei: ... that's totally possible, I will check that.

src/components/bookDetails/Recommendations.tsx
kristojorg: Once I separate CPW and OWC, it will be much easier to update the data parsing so we can get rid of the `raw` object entirely

hokei (approved):
this is super exciting!  

I think other than removing the `axisnow-access-control-web` from the test ignore path, this looks pretty good to me.  

I'm seeing a lot of `string | undefined` types here.  I wonder if there's a way we can reduce them going forward, now that we know they're here.  Like, doing explicit undefined checks further up, so we're not always checking for undefined.  But yeah!  The kind of stuff that this work helps us with!

package.json
hokei: The only thing that this looked like it did was throw an error at `delete window.location;` in `useAuth.test.ts` , I guess https://github.com/NYPL-Simplified/circulation-patron-web/pull/123/files#diff-1923ffd3df464acac96035dc8fc547d5R9 this fixed it?

.eslintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     node: true
   },
   parser: "@typescript-eslint/parser",
+  ignorePatterns: ["/axisnow-access-control-web/**"],
```
hokei, line 8: As a best practice, do you think we should try to get this on strict mode as well? I think we /could/ if we wanted to, and I kind of want this folder to be one that could /hypothetically/ live inside CPW (for example, if we were somehow able to make this open source)

kristojorg: Actually, thinking about it more, I don't think submodules should be checked by this app. We should set up typescript in that module and a test runner as well because what happens when someone who doesn't have access to it tries to work on this repo?

hokei: When someone doesn't have access to it, it wouldn't even get pulled, so it shouldn't make a difference?  Since we have to actively ignore it, if someone doesn't have access to it, wouldn't it just not know that this repo exists to begin with? A lot of these config files are setup like, "include this big folder, but exclude these subfolders". 

But if something isn't in the big folder, it doesn't need to be excluded because it doesn't exist ... am I making sense?

hokei: The submodule isn't an npm module, so it doesn't have its own npm install or npm test.  If we don't have CPW check it by default, I think we'll want to set up an `install-with-submodule`, `lint-with-submodule` and `test-with-submodule` command.  Which I had considered and thought it might be really messy, but maybe it might not be too bad?

package.json
kristojorg: Exactly : )

jest.config.js
hokei: It has been running them:  
Screen Shot 2020-09-09 at 2 03 21 PM

.eslintrc.js
hokei: In any case, I don't think this has to block merging this.  We can take the submodule out for now, and put it back in in a future PR.  I'd rather not let CPW get too far away from this since there are so many spread-out changes here!

kristojorg:
Okay I added the submodule back in to the tests and you were right it is running them!

kristojorg:
Okay I ran in to errors with the submodule and don't fully know how to fix them. I disabled linting on the submodule, but testing and coverage is re-enabled. We should make a follow up PR to enable linting the submodule. Gonna merge this so we can get the important stuff in!

PR #124, opened 2020-09-02
Clever auth button needs to use AnchorButton

M0nica:
Fixes:
> The Clever auth button currently has a <button> nested inside an <a> tag. This isn't proper semantic HTML, and it isn't a11y friendly. It also may cause bugs because it is possible to follow the anchor without causing a click event on the button. We need to make this component use an <AnchorButton> instead, which is designed for the purpose of an anchor that looks like a button.


M0nica:
> to fix the typecheck issue copy over this line: https://github.com/NYPL-Simplified/circulation-patron-web/pull/93/files#diff-e9f950f17198d3d5e3122a44230a09b9R50
> 
> (It's cuz typescript got updated so github ci is now using a different version than we are)

Thanks, that seemed to resolve TypeScript complaining about a file that's unrelated to my changes.

hokei:
I don't quite understand -- why isn't this a pure button?  What is `href={authUrl}` doing here?  There's no navigation, so why is it using an anchor in the first place?

M0nica:
> I don't quite understand -- why isn't this a pure button? What is `href={authUrl}` doing here? There's no navigation, so why is it using an anchor in the first place?

authUrl is the redirectUrl to get the circulation manager to handle authentication with Clever. It's an external link that redirects to the Clever login.

The user flow is: Click Clever button -> navigates to authUrl -> authUrl switches to clever auth callback url -> login -> are redirected back to open e-books.

kristojorg:
So then I guess I'm curious how both the external link and the `onClick` work together? Which one happens first? Do they both need to happen? Maybe it would be better actually to navigate programatically after saving the auth credentials with `window.open`?

M0nica:
> So then I guess I'm curious how both the external link and the `onClick` work together? Which one happens first? Do they both need to happen? Maybe it would be better actually to navigate programatically after saving the auth credentials with `window.open`?

I updated  to programmatically navigate to the authLink after saving the authCredentials

src/auth/cleverAuthButton.tsx
```diff
@@ -3,45 +3,52 @@ import { AuthMethod } from "opds-web-client/lib/interfaces";
 import { AuthButtonProps } from "opds-web-client/lib/components/AuthProviderSelectionForm";
 import { useActions } from "opds-web-client/lib/components/context/ActionsContext";
 import Button from "components/Button";
+
 import { modalButtonStyles } from "components/Modal";
+import { useRouter } from "next/router";
+
+export function getAuthUrl(provider, currentUrl) {
```
kristojorg, line 10: Please type the arguments to this function

src/auth/__tests__/cleverAuthButton.test.tsx
```diff
@@ -1,39 +1,26 @@
...
+  }
+};
 
-  it("links to authenticate with redirect", () => {
-    expect(
-      wrapper.getByText("Log In With Clever").closest("a")
-    ).toHaveAttribute(
-      "href",
-      "https://testing.com/USOEI/oauth_authenticate?provider=Clever&redirect_uri=http%253A%252F%252Ftest-domain.com%252F"
+describe("CleverButton", () => {
```
kristojorg, line 20: Can you add a test in here for whatever conditions cause it to render `null` (even if we weren't testing this before, I think we should have that code path covered).

kristojorg (changes requested):
Looks good, just a couple of changes

kristojorg:
Also please update the changelog with a `- fix: xyz...`

src/auth/__tests__/cleverAuthButton.test.tsx
M0nica: I updated to add tests for the button (whether or not it renders based on if provider is passed in)

src/auth/cleverAuthButton.tsx
```diff
@@ -1,48 +1,63 @@
...
+
+export function getAuthUrl(
+  provider: AuthProvider<AuthMethod> | undefined,
+  currentUrl: string
+) {
+  // double encoding is required for unshortened book urls to be redirected to properly
+
+  if (!provider || !currentUrl) {
+    return "";
+  }
```
kristojorg, line 18: Do we really want to return an empty string here or should we return null? What is more clear for the caller of this function? Or do we not even want to accept the case where `provider` is `undefined` and handle that case upstream?

kristojorg: It looks like you're already testing for whether `provider` is defined below, so maybe change the arguments of the function so that it can't be `undefined` and then you shouldn't need this check at all I think?

src/auth/cleverAuthButton.tsx
```diff
@@ -1,48 +1,63 @@
...
+
+  return `${
+    (provider?.method.links || []).find(link => link.rel === "authenticate")
+      ?.href
+  }&redirect_uri=${encodeURIComponent(encodeURIComponent(currentUrl))}`;
+}
 
-const CleverButton: React.FC<AuthButtonProps<AuthMethod>> = ({ provider }) => {
+export function CleverButton({ provider }): JSX.Element | null {
```
kristojorg, line 26: Curious why you changed how this was defined? I think this leaves `provider` untyped in this definition. Seems like it should take the `AuthButtonProps<AuthMethod>` type

src/auth/cleverAuthButton.tsx
```diff
@@ -1,48 +1,63 @@
...
-    </a>
+    <Button
+      onClick={() => {
+        dispatch(
+          actions.saveAuthCredentials({
+            provider: "Clever",
+            credentials: ""
+          })
+        );
+        router.push(authUrl);
```
kristojorg, line 47: I'm actually not sure this is what we want to use. It's an external url isn't it? Does `router.push` work for external urls? Maybe check the nextjs docs

M0nica: `router.push` works for external urls but the nextjs documents recommend using window.location to navigate to external URLs so I have updated accordingly. https://nextjs.org/docs/api-reference/next/router#routerpush

src/auth/cleverAuthButton.tsx
M0nica: I updated the definition while trying to debug some issues with hooks errors in the unit tests but this change was not necessary so I have reverted.

src/auth/cleverAuthButton.tsx
M0nica: I just updated to make provider and currentUrl required inputs for the getAuthUrl function.

M0nica: Now when either, current location or provider is, undefined then the CleverButton will return null for the AuthUrl and when the AuthUrl is undefined the CleverButton doesn't render.

kristojorg (approved):
looks good to me

PR #125, opened 2020-09-02
display loans from redux on MyBooks page

M0nica:
- It looks like MyBooks is not fetching all of the loans from Redux. It's checking state.collection.data?.books for loans when it should also be checking state.loans.books. This seems to resolve the issue with loans not appearing when navigating _to_ the MyBooks page within the app. 

This is intended to be a short-term fix until opds-web-client/redux are removed from circulation-patron-web.

🤔 However, a hard refresh or navigating directly to https://localhost:3000/loans is displaying an empty state as the redux state is cleared. I am thinking that should be addressed in a separate ticket.


kristojorg:
Should we _only_ be checking `state.loans.books`? And then we shouldn't use `setCollectionAndBook` when rendering the component, but use the opds-web-client `fetchLoans` action instead? That seems like it's the way this is supposed to work and I'm not sure why I had it working with `state.collection` at all to be honest. Must have just been my mistake

src/components/MyBooks.tsx
```diff
@@ -56,13 +55,14 @@ export const MyBooks: React.FC<{
       <Head>
         <title>My Books</title>
       </Head>
+
       <BreadcrumbBar currentLocation="My Books" />
       <PageTitle>My Books</PageTitle>
-      {collection.isFetching ? (
+      {!loans ? (
```
M0nica, line 61: LoansState doesn't have a loading state https://github.com/NYPL-Simplified/opds-web-client/blob/ad836e0e6a1d94e34caca2427c1620729ac87861/packages/opds-web-client/src/reducers/loans.ts#L9 unlike CollectionState https://github.com/NYPL-Simplified/opds-web-client/blob/ad836e0e6a1d94e34caca2427c1620729ac87861/packages/opds-web-client/src/reducers/collection.ts#L15 so I'm not sure how to best handle the loading state.

M0nica: the test that is currently failing is related to the loading state ❌

M0nica: I wasn't sure if loading state can be removed since LoansState doesn't have a fetching/loading state?

src/components/MyBooks.tsx
```diff
@@ -34,20 +33,25 @@ function sortBooksByLoanExpirationDate(books: BookData[]) {
...
-  // here we pass in "loans" to make it look like we are at /collection/loans
-  // which is what used to be the route that is now /loans (ie. this page)
-  useSetCollectionAndBook(setCollectionAndBook, "loans");
-  const collection = useTypedSelector(state => state.collection);
+export const MyBooks: React.FC<{}> = () => {
+  const { actions, dispatch } = useActions();
+
+  const [books, setBooks] = React.useState<false | BookData[] | undefined>(
+    undefined
+  );
```
kristojorg, line 41: I'm curious why you introduced this state? Aren't the books already stored in redux state under `loans.books`? It seems like we can just use that but maybe there is a reason to do it this way instead?

src/components/MyBooks.tsx
```diff
@@ -34,20 +33,25 @@ function sortBooksByLoanExpirationDate(books: BookData[]) {
...
   const { isSignedIn } = useAuth();
 
-  const books =
-    collection.data?.books &&
-    collection.data.books.length > 0 &&
-    collection.data.books;
+  React.useEffect(() => {
+    if (loansUrl) dispatch(actions.fetchLoans(loansUrl));
+    setBooks(loans?.books && loans.books.length > 0 && loans.books);
+  }, [loans, loansUrl, actions, dispatch]);
```
kristojorg, line 54: This is going to fire off the fetchLoans network event every time the `loans` object changes (via reference equality), which doesn't seem like what we want or am I missing something?

kristojorg: Just checked and it is indeed creating an infinite loop:
image

kristojorg (changes requested):
To me it looks like you can remove the local state and that will fix this issue

kristojorg:
Also please update CHANGELOG with a `- fix: xyz...`

src/components/MyBooks.tsx
```diff
@@ -56,11 +60,10 @@ export const MyBooks: React.FC<{
       <Head>
         <title>My Books</title>
       </Head>
+
       <BreadcrumbBar currentLocation="My Books" />
       <PageTitle>My Books</PageTitle>
-      {collection.isFetching ? (
-        <PageLoader />
-      ) : !isSignedIn ? (
```
kristojorg, line 63: I think removing the loading state is confusing, but it's okay for now because this will all be replaced somewhat soon

src/components/MyBooks.tsx
M0nica: I just updated to select `state.loans` instead of using the `useEffect()`

src/components/__tests__/MyBooks.test.tsx
```diff
@@ -126,50 +113,13 @@ test("displays books when signed in with data", () => {
...
-const loading: State = merge(fixtures.initialState, {
-  auth: {
-    credentials: authCredentials
-  },
-  collection: {
-    isFetching: true
-  }
-});
-
-test("shows loading state", () => {
```
hokei, line 164: why does loading state need to no longer show?

kristojorg:
I notice we are no longer calling `fetchLoans` at all. Does this work without that?

M0nica:
> I notice we are no longer calling `fetchLoans` at all. Does this work without that?

It seems to work with or without fetchLoans but I am re-adding to try to avoid stale loans

kristojorg:
Okay that's fine for now thanks

kristojorg (approved):
looks good

PR #126, opened 2020-09-07
Analytics

kristojorg:
This PR adds Google Tag Manager for tracking application usage. It is a relatively simple implementation so far, but it tracks: 

- Page views
- Book and Collection load events (when it's actually been loaded vs when you first navigate to the page)
- Fulfilled book event: This is either a download leads to an external read online. We can't differentiate the two because the logic is currently in opds-web-client. That will change in the near future.
- Borrowed book event
- Reserved book event
- Search event
- Performance metrics [provided by Next.js](https://nextjs.org/docs/advanced-features/measuring-performance) including the full scope of standard Web Vitals metrics.

It also places the client-side env of the app into the GTM datalayer so it can be used for triggering events.


hokei:
Can you provide some instructions on how to test/verify this?  Is there a specific account to use? Is it tracking everything across all deployments?  I'm not seeing any configuration at all, should there be some?

kristojorg:
So in order to test this you'll have to use google tag manager and set a `NEXT_PUBLIC_GTM_ID` env var, then you can enable "preview mode" in GTM and it will show you on the site as the dataLayer gets updated and events are fired. It's a bit of a confusing system. 

What do you mean by configuration?

It's only tracking events for the instance that has the `NEXT_PUBLIC_GTM_ID` set

hokei:
@kristojorg sorry to keep putting this off - I'm finally getting to it, can you. merge beta so that it no longer has the collections issue?

kristojorg:
Okay merged beta in!

hokei:
So I wanted to set this up and test it, but it looks like NYPL won't just let me add test account properties. 

But I want to make sure I understand this work: 

1) Get the Tag Manager ID from Google Analytics
2) Run the instance with the tag manager env variable
3) See the analytics updated in dashboard?  

(Can you send a screenshot of the dashboard with all of this working?  I can't verify, but I'm just curious what it'll look like)

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -267,7 +272,12 @@ const AccessCard: React.FC<{
                 return <ReadOnlineButton key={link.url} link={link} />;
               }
               return (
-                <DownloadButton key={link.url} link={link} title={title} />
+                <DownloadButton
```
hokei, line 275: can you do this with the read online buttons as well?  They won't get loaded in the SimplyE versions of the app, but might as well.

src/analytics/PageviewTracker.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from "react";
+import { useRouter } from "next/router";
+import * as env from "utils/env";
+import useLinkUtils from "components/context/LinkUtilsContext";
+import useTypedSelector from "hooks/useTypedSelector";
+import track from "analytics/track";
+
+const PageviewTracker: React.FC = ({ children }) => {
```
hokei, line 8: why does this have to be its own component rather than a method under `track`?

hokei (changes requested):
some questions

src/analytics/GoogleTagManager.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const GTMScript = () => {
+  return (
+    <script
+      dangerouslySetInnerHTML={{
+        __html:
+          `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':` +
+          `new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],` +
+          `j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=` +
+          `'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);` +
+          `})(window,document,'script','dataLayer','${GTM_ID}');`
```
hokei, line 13: what happens if none got passed?  Do all of these still get called with "undefined"?

src/analytics/PageviewTracker.tsx
kristojorg: This basically uses react hooks to call the methods on `track` at appropriate times when things change

src/components/bookDetails/FulfillmentCard.tsx
kristojorg: I'm planning to add read online tracking in a future PR once the data layer rewrite is merged

src/analytics/GoogleTagManager.tsx
kristojorg: Good point, I shouldn't render these at all if the env var isn't set

kristojorg:
> Get the Tag Manager ID from Google Analytics
Run the instance with the tag manager env variable
See the analytics updated in dashboard?
(Can you send a screenshot of the dashboard with all of this working? I can't verify, but I'm just curious what it'll look like)

Sooo basically what this does is 

1. Get the GTM id from the env, and set up Google Tag Manager.
2. Send events to Google Tag Manager

The way it is used is actually pretty complicated in my opinion, but GTM was the preferred way of doing this from when I asked around. Basically you have to set up a bunch of variables, triggers and tags in GTM, which then sends data to google analytics which you can finally view. I actually don't have a good dashboard for this yet... that fell outside the scope of my work. I was just supposed to set it up to send the data and set up GTM to get the data into GA, and James at Lyrasis is now working on creating nice views from the information with Google Data Studio

And yes, you're right I probably should add tests for the tracking. Will do that tomorrow

leonardr:
Ducking in to flag some serious architectural concerns I have about this, which I described in https://jira.nypl.org/browse/SIMPLY-3151. Basically, this looks like a duplicate of the analytics system we already have. 

With two possible exceptions, I'm not convinced that any of these analytics events should be sent directly from the web browser to Google. (I'm withholding judgement on two events that I don't understand very well.) In particular, I'm not convinced by the argument that we need a separate analytics system to distinguish between events originated in a mobile app and events originated in a web browser.

Since people who don't want this system are free not to enable it, I don't think this raises to the level of "should not be merged". But this should, as soon as possible, be redone as a set of improvements to the existing analytics system, so that we have one consistent system across web and mobile.

christhepianist:
This sounds like a large analytics project which should be discussed with the community. Is there a reason it is just being done with it not being vetted by at least NYPL, e.g., Leonard?

leonardr:
Most of the time Kristo's leadership on c-p-w has worked out fine because he knows the web side very well, and most c-p-w decisions don't reach beyond the web side. There are a couple of situations where he and I have had to talk things out, and this is one I missed. I should have noticed this and stepped in earlier.

Right now this branch is in a state where I don't think anyone but LYRASIS will use the feature. I'll talk to Kristo and we'll figure out whether it's better to merge it and fix it, or have LYRASIS work off a fork until we can do a more comprehensive implementation.

kristojorg:
Okay I've updated this PR:

1. Google Tag Manager is integrated on an optional basis, if the NEXT_PUBLIC_GTM_ID is set in the environment. If not, the package is not loaded at all.
1. I removed all custom events except "pageview", which just sends some CPW specific information to GTM, such as url variables (collection url, book url, library short name).
1. I filed a ticket here to complete the backend work needed to make analytics more useful. That ticket is [SIMPLY-3164](https://jira.nypl.org/browse/SIMPLY-3164)

hokei:
I'll let Leonard sign off on the approach first and then I'll do a code pass =)

src/pages/_document.tsx
```diff
@@ -1,14 +1,18 @@
 import * as React from "react";
 import Document, { Html, Head, Main, NextScript } from "next/document";
+import { GTMScript, GTMNoscript } from "analytics/GoogleTagManager";
 
 class MyDocument extends Document {
   render() {
     return (
       <Html lang="en">
-        <Head />
+        <Head>
```
hokei, line 9: nit: would it make sense to do the check on whether analytics is enabled here, rather than inside `GTMScript`?

src/utils/env.ts
```diff
@@ -23,5 +23,6 @@ export const NEXT_PUBLIC_AXIS_NOW_DECRYPT =
 export const IS_DEVELOPMENT = process.env.NODE_ENV === "development";
 export const IS_SERVER = typeof window === "undefined";
 
+export const GTM_ID = process.env.NEXT_PUBLIC_GTM_ID;
 export const BUGSNAG_API_KEY = process.env.NEXT_PUBLIC_BUGSNAG_API_KEY;
```
hokei, line 27: random: whenever I see this I think of the Bugsnax song https://youtu.be/6-v1BxSETvI?t=50

hokei (approved):
The code looks OK -- I'm not able to actually verify functionality due to not being able to set up GTM, but I'm ok with having it in.

src/pages/_document.tsx
kristojorg: I thought about that... but it seems cleaner in GTMScript, especially in case someone renders that elsewhere somehow.

src/utils/env.ts
kristojorg: I have so many questions after watching that...

PR #127, opened 2020-09-09
Fix decryptor tsc errors

hokei:
Re-assigns decryptor location so that tsc doesn't throw an error when the decryptor submodule is not available.


kristojorg (approved):
Interesting that this solves it. Typescript so weird. Looks good tho

hokei:
yeah I went down all these weird module declaration rabbit holes and was like oh. lol. I guess this works.   I'm just gonna make sure I haven't broken the submodule cuz that "submodule deleted" diff is messing with me now =(

PR #132, opened 2020-09-10
Fix eslint in axisnow

kristojorg:
This fixes the axisnow eslint setting so that submodule continues to be linted

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,7 +5,6 @@ module.exports = {
     node: true
   },
   parser: "@typescript-eslint/parser",
-  ignorePatterns: ["/axisnow-access-control-web/**"],
   parserOptions: {
     project: "tsconfig.json",
     sourceType: "module",


hokei:
to get this to build I think you want to re-pull the latest axis-now-access-control-web.

kristojorg:
Just updated the axisnow subrepo

kristojorg:
@hokei mind giving this an approve?

hokei (approved):

kristojorg:
@hokei actually I'm not sure this should be merged anymore... it updated the submodule to a different branch where I removed the typescript error, but I think the submodule has moved on from there and now this would effectively be moving it backwards? How would I know if that's the case?

hokei:
@kristojorg ... there hasn't been changes to the submodule since the last merge though?

kristojorg:
@hokei yea I suppose that's true... interesting. Okay well let's merge it then!

PR #135, opened 2020-09-11
Remove url shortener

kristojorg:
This PR removes the `SHORTEN_URLS` environment variable. This variable would make urls that look like this:
```
http://localhost:3000/howard/collection/https%3A%2F%2Fhoward.simplye-md.org%2FMDHOWRDCTLIB%2Fgroups%2F236%3Fentrypoint%3DAll
```
look like this:
```
http://localhost:3000/howard/collection/groups%2F236%3Fentrypoint%3DAll
```
The env var is intended to make the urls prettier for the user, but it doesn't really accomplish that and it adds significant complexity to the app. I'm proposing we remove it in favor of a server solution down the line that gives books a pretty uuid like `/catcher-in-the-rye-HG5M34` or similar. 

Removing this is the first step in some larger changes I am making to simplify the application to make it more reliable, testable, and easier to update.


kristojorg:
@leonardr I would like your opinion on this, since it is technically a breaking change. If people are depending on this functionality and they update to the most recent app, external URLs pointing to the app could break. However, we have already changed the url structure a bit since it was last deployed into production (by DPLA I believe), so it's sort of a moot point.

leonardr:
The place where this might break things is MARC records. We have code in the circulation manager that (given an appropriate configuration) generates a URL into this system like http://cpw.library.org/howard/works/ISBN/9781010101010.

Web catalogs import those MARC records and provide a link from the web catalog to the library's CPW installation. So http://catalog.library.org/books/9781010101010 would have a link to link to http://cpw.library.org/howard/works/ISBN/9781010101010. If you remove the URL shortener then that link will break.

However I don't know if the MARC record functionality is actually used anywhere. I recently generated some MARC records for Open eBooks, which resulted in MARC records containing URLs like https://catalog.openebooks.us/book/USOEI%2FAxis%2520360%2520ID%2F0012415473. I sent those off to some librarians who'd asked for MARC but haven't heard back from them.

I don't know of any other real CPW installations that would be affected by this (DPLA intended to set some up but I don't know if they've done it yet). And it took a lot of hacking of circulation manager code to actually generate working MARC records, so it's not a feature that's currently in heavy use, or someone would have noticed before I did.

leonardr:
From my perspective the server solution is in place. Every OPDS entry has two identifiers: its Atom <id> (`urn:librarysimplified.org/terms/id/Bibliotheca%20ID/s75gog9`)
and its OPDS entry `alternate` link (https://circulation.librarysimplified.org/NYNYPL/works/Bibliotheca%20ID/s75gog9/)

The <id> identifies the book in terms of the library's license for that book, but it's missing an important piece of information: the book is being accessed in the context of a specific library (NYNYPL, in this case) and only the `alternate` link contains that information.

We can add a slug to the OPDS but it's going to need to contain the same crucial information as the URL: the library and the book identifier. Then we'll need to give the client some rules for turning a book's slug into a URL. Which seems like adding back the complexity being removed in this PR.

I can see how "/catcher-in-the-rye-HG5M34" looks nicer than "/USOEI/Axis 360 ID/2F0012415473", but "HG5M34"  is doing all the work.

kristojorg:
@leonardr the main piece of complexity that I'm trying to remove is actually the ability to configure this behavior, not necessarily how the URLs are generated. I do think the alternate IDs should be shorter, whether they include something readable or not doesn't matter much (we can add that functionality in CPW directly). 

> Web catalogs import those MARC records and provide a link from the web catalog to the library's CPW installation. So http://catalog.library.org/books/9781010101010 would have a link to link to http://cpw.library.org/howard/works/ISBN/9781010101010. If you remove the URL shortener then that link will break.

This link shouldn't break. What will happen is that the web catalog url for that resource would now be: `http://catalog.library.org/books/http%3A%2F%2Fcpw.library.org%2Fhoward%2Fworks%2FISBN%2F9781010101010`. This is admittedly uglier, but I think we should solve this for _all books_, not just those with MARC records. 

I would default to short urls since they are marginally nicer, and they work well for the MARC records, but there have been issues in the past with this functionality across circulation managers, so I decided to default to safe long urls that encode the full url of the needed resource instead, and then we can work adding a short ID solution to the CM in the future.

leonardr:
I'm OK with going to long URLs for a while and shortening them later, but I'm not convinced that the links won't break. It's my own fault for not explaining my URLs; "catalog" and "cpw" sound very similar but I was using them to refer to totally different pieces of software. Let me try again with more realistic URLs.

Here's [the NYPL catalog page](https://browse.nypl.org/iii/encore/record/C__Rb19957147__S%20longitude__Ff%3Afacetavailability%3AwANDz%3AwANDz%3AOnline%3A%3A__Orightresult__U__X7?lang=eng&suite=def) for a book called _Longitude_. Note that the hostname is "browse.nypl.org".

Here's [the circulation manager's OPDS entry for the same book](https://circulation.librarysimplified.org/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2). The hostname is circulation.librarysimplified.org.

Now let's imagine we set up a CPW install for NYPL's collection, at cpw.nypl.org. It's going to handle URLs like https://cpw.nypl.org/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2.

Then we generate a bunch of MARC records with URLs like https://cpw.nypl.org/NYNYPL/works/..., and import those MARC records into the server at browse.nypl.org.

Then we make the change proposed in this PR. At that point URLs like https://cpw.nypl.org/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2 stop working. We need to start using URLs like https://cpw.nypl.org/books/https://circulation.librarysimplified.org/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2.

But browse.nypl.org is full of URLs like https://cpw.nypl.org/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2, and those URLs have stopped working. We need to regenerate and re-import the MARC records to update the URLs to the new format.

Does this make sense?

kristojorg:
Yep that makes sense I think... basically, anytime we change the url structure we will break any external URLs to the product. This is part of why I think the URL structure should be baked-in and not configurable - you shouldn't really be switching back and forth. 

Here is my proposed solution: 

1. We switch to long-urls for now, using the full `alternate` link. 
2. At some point in the future we configure the CM to generate a shorter `uuid` for resources.
3. We update CPW to use the new `uuid` as the slug for the resource, but under `/book/id/:uuid` (or something). This way all of the old url-based links still work at `book/:bookUrl` and minimal complexity is added to CPW. The difficulty of adding a page (or even redirect) to support that is minimal compared to the difficulty of supporting multiple internal link generators.

leonardr:
I'm not happy about this, but I'm not _unhappy_ enough to veto the PR, and I agree that making this change as soon as possible will minimize the risk of broken links.

kristojorg:
Can I ask what you find objectionable about it?

leonardr:
Two things:

I don't like breaking URLs, even though I'm pretty sure they're not being used in production.

If we're talking about expanding the URLs to https://cpw.nypl.org/books/https://circulation.librarysimplified.org/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2 and then re-shortening them later, I think the obvious first shortening step is to remove the _hostname_. On any given CPW instance the hostname is redundant. That would re-shorten the URL to https://cpw.nypl.org/books/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2, which is basically what we have now.

I don't see a lot of value in replacing "Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2" (a persistent identifier which is readable to someone diagnosing a problem, though not to a library patron) with "HG5M34" (which is short but completely opaque to everyone involved). But that one is a concern I have about what comes _after_ this branch, not about this branch per se.

kristojorg:
I see. I had thought about removing the configurability but keeping the shortening functionality as the default, but the issue is that we have run in to problems with that, especially with the e-reader changes. As of right now, I don't think the reader works without `SHORTEN_URLS=false` (@hokei can you confirm?). If we fix that, then I see no reason not to use the shortened urls always instead of the long urls always for right now. 

However, I think that the complexity of shortening a url is only worthwhile if it makes that url meaningful to a user or easier to use/share. In my mind, `http://cpw.library.org/book/id/HG5M34` is easier than `https://cpw.nypl.org/books/NYNYPL/works/Overdrive%20ID/d64091a1-a96d-4e4f-9484-0dcb1f6ad3d2`, but yea I guess it's only marginal. 

So maybe the right solution here is to make it work with the shortened urls always, like this `http://localhost:3000/howard/collection/groups%2F236%3Fentrypoint%3DAll` ? Again, what I really want to get rid of is the ability to configure this behavior.

leonardr:
I'm OK with getting rid of the ability to configure this behavior. The only reason I can think of why shortening wouldn't work is if you have a single CPW instance that goes against several different circulation managers.

And of course that shortened URLs currently don't work with the reader. Maybe that's because reader URLs generally don't come from the circulation manager hostname, so "shortening" those URLs doesn't work? If so (SPECULATION FOLLOWS), maybe the solution is to remove the circulation manager hostname for URLs at that hostname, and leave other URLs alone? Would that reintroduce the problem you're trying to fix?

kristojorg:
It currently isn't a problem when one CPW instance is backed by separate CMs. Each library in the CPW instance has a catalog root url, so we use that to get the hostname we need to remove when shortening and add back in before fetching. 

That speculation does seem reasonable, and we could probably design a regex to do what you're suggestion though it would worry me a little. We would basically need two functions: `(Hostname, FullUrl) -> ShortUrl | FullUrl` and `(Hostname, ShortUrl | FullUrl) -> FullUrl`. The thing I worry about is being able to tell in that second function if you've received a `ShortUrl` or a `FullUrl` since at the end of the day they're both just strings and the client has no control over how they're formed. But again, a regex could probably do it fine if we know the guarantees around how the url is structured from the backend.

leonardr:
I'd think you could distinguish `ShortUrl` from `FullUrl` by checking for a URI scheme at the front?

kristojorg:
Okay so something like this?:

```ts
function shortenResourceUrl(
  catalogRootUrl: string,
  longResourceUrl: string
): string {
  const catalogOrigin = new URL(catalogRootUrl).origin; // returns "https://example.org" from 'https://example.org/foo/bar?baz'
  const resourceOrigin = new URL(longResourceUrl).origin;
  // bail out if the long url is not relative to the catalog root origin
  if (catalogOrigin !== resourceOrigin) return longResourceUrl;
  return longResourceUrl.replace(catalogOrigin, ""); // will replace only the first occurrence
}

function expandResourceUrl(
  catalogRootUrl: string,
  shortenedResourceUrl: string
): string {
  // bail out if shortened url already starts with a protocol. This means it was never really shortened.
  if (
    shortenedResourceUrl.startsWith("http://") ||
    shortenedResourceUrl.startsWith("https://")
  ) {
    return shortenedResourceUrl;
  }
  const catalogOrigin = new URL(catalogRootUrl).origin;
  return `${catalogOrigin}${shortenedResourceUrl}`;
}
```

kristojorg:
My personal feeling is that this is a bit flaky and I would rather go with the long urls always unless we have a short url that actually provides a better user experience, or we know that there are existing (and working) production backlinks using the short url syntax that need to be preserved

leonardr:
Yeah, that's what I was thinking. We have similar code on the server side for managing URLs.

After all this I'm back to my first comment: I'm OK with going to long URLs for a while and adding back shorteners later.

tdilauro:
> The place where this might break things is MARC records. We have code in the circulation manager that (given an appropriate configuration) generates a URL into this system like http://cpw.library.org/howard/works/ISBN/9781010101010.

Gents: Sorry that I didn't notice this thread earlier, but we do have MARC export configured in at least one CM. 

@jonathangreen: Can you weigh in here?

leonardr:
@tdilauro Can you give a specific example of a URL that's embedded in a MARC record? When I generated MARC records for Open eBooks I found a bunch of little problems and the upshot is I think the URLs you generated might have always been broken.

tdilauro:
https://ebooks.lyrasistechnology.org/clp/book/URI%2Fhttps%3A%2F%2Funglue.it%2Fapi%2Fid%2Fwork%2F145870%2F

leonardr:
OK, that definitely works. So now I'm more concerned about maintaining backwards compatibility with URLs like the one Tim gave. But it's fine to strip out the ability to configure the shortener.

leonardr:
One low-impact way of doing this might be to specifically introduce backwards compatibility such that a URL like https://ebooks.lyrasistechnology.org/clp/book/URI%2Fhttps%3A%2F%2Funglue.it%2Fapi%2Fid%2Fwork%2F145870%2F is handled by _lengthening_ it, by prepending the circulation manager root URL.

kristojorg:
Okay well if we are already using some shortened urls in production MARC records that are difficult to change, I can use the code above to ensure we always shorten urls. 

One question though - are we also using long urls in production somewhere? Do we need to maintain backwards compatibility with those as well?

leonardr:
I don't want to 100% give up on changing the MARC records. @tdilauro would it be a lot of work to regenerate and import a complete set of MARC records for the affected CMs? One mitigating factor is I suspect any delta MARC records aren't being generated, so those libraries may not have up to date data anyway. I can spend some time working with you on this.

However, I think the best overall solution is to simplify the codebase as much as possible while preserving backwards compatibility.

leonardr:
I don't think we are using long URLs anywhere. MARC records are always generated with the short URLs and I don't know of any other place we _store_ these URLs.

tdilauro:
> I don't want to 100% give up on changing the MARC records. @tdilauro would it be a lot of work to regenerate and import a complete set of MARC records for the affected CMs? 

We have verified that the library in question has not yet imported any MARC records, so that should not be a problem.

> One mitigating factor is I suspect any delta MARC records aren't being generated, so those libraries may not have up to date data anyway. I can spend some time working with you on this.

I'm not sure what you mean by this. Is there an issue we should be tracking related to the deltas? We are getting deltas. I just checked the most recent one for "All books" and it seems to contain valid `856|u`s, at least.

leonardr:
There's no ticket; I had problems getting MARC export working in a non-realistic scenario but it might work fine in a realistic scenario -- I couldn't set up a realistic scenario to test it.

If you're getting deltas then that's evidence that it works and the worst I can say is we should support some different setups.

@kristojorg Sorry to go back and forth on this but it sounds like regenerating MARC records is not a deal-breaker. But if you change the URL format we'll need to adjust the circulation manager to match.

kristojorg:
@leonardr okay sounds good. I am going to plan on merging this change to use the full url instead of shortened urls then. I think it suits our needs of simplifying the code base well and will allow us to add short urls at a later date when they provide a meaningful UX improvement. 

So it sounds like we need to change the CM to match the URL structure when generating MARC records. Should I merge this now or wait until that is done on the backend?

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3093 to cover the circulation manager work. @tdilauro can you take a look and let me know if you're available to do this work? I don't know when I'll be able to get to it (plus I'm not equipped to run an end-to-end test) but I don't want to hold up Kristo.

kristojorg:
Thanks @leonardr and @tdilauro for your help here! Much appreciated : )

tdilauro:
@leonardr, @kristojorg I just took a quick look at -- and left a couple comments on -- the JIRA issue. I should be able to do this work early next week.

kristojorg:
Thanks @tdilauro . Can I get an approval in that case from @leonardr and/or @hokei ? We will move forward with long urls for the time being until short urls provide a valuable UX improvement, at which time we will simply support both.

leonardr:
I'm not qualified to review the actual code so I'll ask @hokei to do the review and sign off on the PR.

src/components/__tests__/PageTitle.test.tsx
```diff
@@ -13,16 +13,16 @@ const formatsFacet: FacetGroupData = {
...
     {
       label: "Audiobooks",
-      href: "/audiobooks",
+      href: "http://audiobooks",
       active: false
     },
     {
       label: "All",
-      href: "/all",
+      href: "http://all",
```
hokei, line 25: Is this more of a documentation thing, to enforce that all urls should be prefixed with http, but it doesn't automatically prefix if you don't pass http in?

src/utils/url.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Takes a url parameter and encodes it to be suitable
+ * for a url
+ */
+export default function encodeUrlParam(param: string) {
+  return encodeURIComponent(param);
```
hokei, line 6: why put this in a different function when we could call encodeURIComponent directly?

src/components/__tests__/PageTitle.test.tsx
kristojorg: It's basically just to make the tests a bit stronger. We want to make sure our code properly encodes valid urls to be used as url params, so I added http to make them more explicitly a valid url, if that makes any sense at all

kristojorg: It's just documentation and test explicitness

src/utils/url.ts
kristojorg: Yea maybe we should just do that. I just was thinking it would be good if there was a single place to update this in the future if need be, but this also doesn't enforce someone can't use `encodeURIComponent` elsewhere. I'll update the PR to remove this

kristojorg:
@hokei I removed that `encodeUrlParam` function

hokei (approved):
Comment out `publish-beta.yml` and then this is good!

PR #136, opened 2020-09-11
#1: Remove opds-web-client

kristojorg:
This PR is the first step of many steps to detach from opds-web-client, remove redux, and rewrite the data layer. Once I am done, the app will be dramatically simpler and easier to work on. Yeehaw!

Here, I simply copy over all of our dependencies from opds-web-client and update them slightly to work with our stricter type checking. Don't worry, we won't be keeping these files around for too long. I just want to get them in here so that I can make smaller PRs from here on out with digestible change summaries.

I am making this PR to the `data-layer-rewrite` branch so that we can do this piecewise and then merge it all in once it's ready. So, let the review games begin!


kristojorg:
Lint and testing is not expected to pass until we are merging the `data-layer-rewrite` into `beta` FYA

EdwinGuzman:
This seems fine but I see none of the respective test files were copied over. I get that the copied files won't stay around for too long but not having any tests at all doesn't feel right.

I also get this when trying to run on a new install:
Screen Shot 2020-09-14 at 12 26 34 PM

What has been the thoughts around getting this repo inside `circulation-admin`? This is an app whereas `opds-web-client` is a package. So, I'm wondering how many hurdles, if any, we'll have to get over to completely get rid of `opds-web-client` in this repo _and_ in `circulation-admin`.

kristojorg:
@EdwinGuzman I'll look in to that error, though these PRs are not meant to run individually necessarily. They're more trying to break down this work into logical chunks that can be reviewed independently. That's also why the tests are not there yet. I'll add the tests at a later date in another PR to this fork before eventually submitting a PR to merge the whole `data-layer-rewrite` branch into `beta`. That's when test coverage will be back to what it is now, lint will run, and we can QA the change before merging into beta

kristojorg:
As far as the changes we will need to make in `circulation-admin`, I don't yet know exactly what strategy we will take to do that, but even if it requires splitting some of CPW back out into a package (I don't think it will), the end result will be better because that package won't house all of the core data logic for the whole app, it will just have the couple of components it needs to render a collection.

src/__tests__/computeBreadcrumbs.test.tsx
```diff
@@ -1,5 +1,5 @@
 import computeBreadcrumbs from "../computeBreadcrumbs";
-import { LinkData } from "opds-web-client/lib/interfaces";
+import { LinkData } from "owc/interfaces";
```
hokei, line 2: Is the separation of "owc" related stuff into its own folder meant to be permanent?  It'd be nice to be separating folders based on functionality, but I understand if this is just a simple way to copy it over for now.

EdwinGuzman:
Okay, that said, this is good pending what you find with the error. Do you have a list or outline of all the changes or the breakdown of tickets/PRs for this "remove opds-web-client" project? It'll make it easier to know so I don't comment about something (e.g. tests) not being there.

hokei:
I agree with Edwin, though, that if it seems like the OWC tests will be similar in any way, let's copy them over.  It'll be easier to review a changelog on tests rather than a wall of New Test, a lot of which might be tweaks from Old Tests.

EdwinGuzman:
>As far as the changes we will need to make in circulation-admin, I don't yet know exactly what strategy we will take to do that, but even if it requires splitting some of CPW back out into a package (I don't think it will), the end result will be better because that package won't house all of the core data logic for the whole app, it will just have the couple of components it needs to render a collection.

As far as I know, the separation of `OWC` and `CPW` was to break out the core logic into a package, and the theme/UI into an app. As part of this "remove opds-web-client" project, can you add a ticket about testing this repo in circulation admin? It'll be a big task but we can walk through the codebase to see how `circulation-admin` uses the `OWC` package.

package.json
```diff
@@ -25,6 +25,7 @@
     "@emotion/core": "^10.0.28",
     "@emotion/styled": "^10.0.27",
     "@nypl/design-system-react-components": "^0.7.2",
+    "@nypl/dgx-svg-icons": "^0.3.12",
```
hokei, line 28: What icons will we need?  Let's try to use design-system for these icons -- we can get the icons in by the time the actual PR is ready.

kristojorg: Yep agreed. I'm not actually sure, but moving over OWC made me add this dependency. We can swap it out at a later date.

kristojorg:
Okay, useful feedback! [Here is a link](https://jira.nypl.org/browse/SIMPLY-2682) to a ticket that now describes the 10 step process I am embarking us upon. 

As far as the tests... two things. First, final tests will be very different from those in OWC since we won't be using redux at all. We will have a whole new set of things to test. Basically, the process that I laid out is one where I delete bits of this `/owc` directory one piece at a time and add some corresponding functionality outside this directory to the app. I can do my best to add tests when I add that functionality if you'd find that helpful, instead of all at the end. 

The second thing is that CPW and OWC use different test runners and syntax, so I couldn't really just copy them over anyways. 

> As part of this "remove opds-web-client" project, can you add a ticket about testing this repo in circulation admin? It'll be a big task but we can walk through the codebase to see how circulation-admin uses the OWC package.

I've added that to the list in my big ticket. Once this branch is up and running, I'm confident we can make it work well with circulation-admin. But if you want to step through what that might look like this week, let's set up a time to chat?

kristojorg:
@hokei @EdwinGuzman is there anything else you all want to discuss before we merge this in to the `data-layer-rewrite` branch?

hokei:
> I can do my best to add tests when I add that functionality if you'd find that helpful, instead of all at the end.

I think it's not necessary to do this if it's going to be a lot of extra work, but if the opportunity comes up to test stuff as you're working, I think that could be helpful to all of us!  In particular, I find tests to be great documentation for "When X edge case happens, Y is intended functionality"

hokei (approved):

package.json
hokei: If it's not too much work to get a list of the icons we will need, it'd be nice to get a ticket for that sooner rather than later.  Then it will be in DS in time for us to use.

hokei: ( if it's just these ones I can make that ticket now)

PR #137, opened 2020-09-11
#2: Reorganize types

kristojorg:
This is the second PR in my data layer rewrite. I'm simply adding better typing for OPDS1 data types. This will help us out a lot in the future.


src/interfaces.ts
```diff
@@ -125,16 +76,142 @@ export interface LibraryData {
...
+  | "http://bib.schema.org/Audiobook"
+  | "http://schema.org/EBook"
+  | "http://schema.org/Book";
+
+export type BookAvailability =
+  | "available"
+  | "unavailable"
+  | "reserved"
+  | "ready";
+export interface BookData {
```
kristojorg, line 105: I also moved this over from `own/interfaces`

EdwinGuzman:
If circulation admins start returning OPDS 2-spec feeds, how do you foresee the codebase being updated? Mostly asking because I see a few `OPDS1.NameOfType` and I'm wondering if those would change to `OPDS2.NameOftype` or will "OPDS2" become just "OPDS" and it'll be backward compatible?

kristojorg:
I previously added some `OPDS2` types in `/types/opds2.ts` because the library registry deals in OPDS 2.x. So basically if the CM starts returning 2.x feeds, we would just have to update the parser to go from `OPDS2.Feed -> CollectionData` instead of `OPDS1.Feed -> CollectionData`. A lot of the `OPDS1.Link` types that are used I believe are the same as in OPDS2 actually. So in that case it would just involve changing the name from `OPDS1.Link` to `OPDS2.Link`.

EdwinGuzman (approved):

kristojorg:
Ok I just pushed another commit which just updated a crapload of import statements. I don't think you guys need to care about that so I'm going to merge this in to the branch

PR #138, opened 2020-09-11
#3: Remove auth plugins

kristojorg:
This is the third PR in my data layer rewrite. It accomplishes # 3 and 4 in [SIMPLY-2682](https://jira.nypl.org/browse/SIMPLY-2682).

## Breakdown of changes

1. Remove the "AuthPlugin" concept. Just store available methods in the `LibraryDataContext`. This context is generally where we will start keeping everything that is library-wide, which includes the auth methods available. 
2. Refactor the `Auth` component. The logic for that is now split between the `UserContext` component and the `AuthForm` component. `UserContext` handles fetching the user's loans and providing the app with all necessary user state. The `AuthForm` component simply displays the appropriate auth flow modal when told to.
3. Remove the `loans` state from redux and move it all to SWR. We fetch loans in the `UserContext` component now, and then can use those loans anywhere throughout the app. 
4. In order to support our own fetching of loans, I have created `dataflow/opds1/fetch` which simply uses the `fetch` api to fetch OPDS1 content. It exports functions for fetching a collection or a book. To fetch a collection, we first fetch the raw data, then parse it to JSON with `opds-feed-parser`, then parse it into an internal `CollectionData` object with our new `dataflow/opds1/parse` functions. The functions in `/parse` are cleaned up versions of the opds-web-client's `OPDSDataAdapter` class. They're now just plain functions and we can continue to improve them over time.

## Notes

- The `OPDSDataAdapter` and `DataFetcher` still exist. We are still using them to fetch collections and books through the old redux flow. This PR only moved the loans fetching over to SWR. Future work will move the collection and book fetching to use this new method as well.
- This branch does not handle the `401` errors received when borrowing a book. That logic still happens in redux. My next PR will move that out of Redux and properly handle the errors by showing the auth page. 

## AuthForm Examples

The `AuthForm` component handles displaying the auth flow. It either shows the single auth method, buttons to select between for 1-4 method, or a combobox selector for 5+ methods. 

**With 5+ Auth Methods**
ezgif com-video-to-gif
_Ignore the blank clever auth button, I just don't have a valid background url set in my mock I was using to generate this._

**With 1 Auth Method**
ezgif com-video-to-gif (1)

**With 1-4 Auth Methods**
ezgif com-video-to-gif (2)


kristojorg:
Okay I am marking this ready for review. The app should run, though lint, typechecking and testing will currently fail. I think it makes sense to have you @hokei @EdwinGuzman begin reviewing however. This is going to be the largest of the changes I make in the course of this PR. From here on out, it should be pretty smooth sailing. 

Before I merge this, I am going to try and get at least the auth tests passing so you can see the test changes made as well, but I don't want to overwhelm you with files, so maybe it would be better to do that in a separate PR? What do you think? This is already pretty large and I don't really want it to grow any larger. Let me know.

If you'd like to have a meeting to walk through this tomorrow I can do that as well. It might be helpful so I can explain the basics of what's happening here.

src/errors.ts
```diff
@@ -35,3 +37,47 @@ export class AppSetupError extends ApplicationError {
...
+  title: string;
+  type?: string;
+};
+
+function isProblemDocument(
+  details: ProblemDocument | OPDS1.AuthDocument
+): details is ProblemDocument {
+  return !(typeof (details as OPDS1.AuthDocument).id === "string");
+}
+export class ServerError extends ApplicationError {
```
kristojorg, line 52: I added this error type so we can have better support for server-generated errors

src/components/Modal.tsx
```diff
@@ -26,6 +28,7 @@ type ModalProps = {
   className?: string;
   hideOnClickOutside?: boolean;
   role?: string;
+  showClose?: boolean;
```
kristojorg, line 31: The modal now has an optional close button

src/interfaces.ts
```diff
@@ -78,6 +79,8 @@ export interface LibraryData {
   } | null;
   headerLinks: OPDS1.Link[];
   libraryLinks: LibraryLinks;
+  authMethods: AppAuthMethod[];
+  shelfUrl: string | null;
```
kristojorg, line 83: The `authMethods` used to be extracted whenever a 401 was received from the `authDocument`. Since we fetch the auth document on startup, we will just extract them the first time and store them in `LibraryData`. Same with the shelf url. Both of these are things the app always needs and don't change often, so it makes sense to fetch them on startup.

src/utils/auth.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (!samlMethod.links) return [];
+  return samlMethod.links.map(idp => ({
+    href: idp.href,
+    type: samlMethod.type,
+    description: getEnglishValue(idp.display_names) ?? "Unknown SAML Provider"
+  }));
+}
+
+export const getEnglishValue = (arr: [{ language: string; value: string }]) =>
+  arr.find(item => item.language === "en")?.value;
```
kristojorg, line 36: These are just moved over verbatim from previous OWC code

EdwinGuzman:
The tests in a separate PR is fine.

src/auth/AuthForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+      {/* We render this to provide the dialog a focus target after it closes
+          even though we don't open the dialog with a button
+      */}
+      <DialogDisclosure sx={{ display: "none" }} {...dialog} />
+      <AuthFormProvider showForm={dialog.show}>{children}</AuthFormProvider>
+    </React.Fragment>
+  );
+};
+
+const Loading: React.FC = () => {
```
EdwinGuzman, line 86: You mentioned above that the loading indicator would show when the user is authenticating bu this variable isn't being used in the file.

src/auth/AuthForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+};
+
+/**
+ * Renders a form if there is one, or a button, or tells
+ * the user that the auth method is not supported.
+ */
+const SignInForm: React.FC<{
+  method: AppAuthMethod;
+  goBackToSelection?: () => void;
+}> = ({ method, goBackToSelection: _ }) => {
```
EdwinGuzman, line 107: Why are you adding `goBacktoSelection` if it's not being used?

src/auth/AuthForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const cancelSelection = () => setSelectedMethod(undefined);
+
+  return (
+    <Stack direction="column">
+      {!selectedMethod &&
+        authMethods.map(method => {
+          switch (method.type) {
+            case OPDS1.BasicAuthType:
+              return (
+                <Button
```
EdwinGuzman, line 170: Why do you use a `Button` here instead of `<BasicAuthForm method={method} />` like in the `SignInForm`?

src/auth/useCredentials.ts
```diff
@@ -0,0 +1,178 @@
...
+ * for credentials stored in a browser cookie as well as any credentials passed in
+ * to the app via a query param (eg. for Clever or SAML auth after authenticating).
+ */
+
+type UrlAuthError = { error?: string };
+type CredentialResult = (Partial<AuthCredentials> & UrlAuthError) | undefined;
+
+export default function useCredentials(slug: string | null) {
+  const router = useRouter();
+  const [credentials, set] = React.useState<CredentialResult>(undefined);
```
EdwinGuzman, line 19: Can you use `setCredentials` rather than just `set`?

src/auth/useCredentials.ts
```diff
@@ -0,0 +1,178 @@
...
+  const [error, setError] = React.useState<UrlAuthError | undefined>();
+  const urlCredentials = useMemoizedUrlCredentials(router);
+
+  React.useEffect(() => {
+    const cookie = getCredentials(slug);
+    if (cookie) set(cookie);
+  }, [slug]);
+
+  const setCredentials = React.useCallback(
```
EdwinGuzman, line 28: I see why you didn't use it above. Hmm the above is convention so maybe this can have a different name?

src/icons/Close.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+
+const SvgClose = (props: React.SVGProps<SVGSVGElement>) => (
```
EdwinGuzman, line 3: I think there might be close and download svgs you can use from the Design System, right @hokei ?

hokei:
Is there documentation around what the intended auth methods are, and how to know whether your library/product uses them, how to know which applies to you, and how to configure them?  My best guess is that these aren't all in one place, and I'm not saying that writing them is a part of this PR, but this continues to be a thing that really confuses me about this app haha

kristojorg:
@hokei not totally. All I've seen is the [auth document spec](https://drafts.opds.io/authentication-for-opds-1.0), but that's now outdated because we have added SAML. Basically, the circulation manager allows configuring different authentication methods, and then sends a list of those methods to this app in the auth_document (along with a bunch of other library-level info). We then just have to support those auth types. I've started to document this in the `types/opds1.ts` file [here](https://github.com/NYPL-Simplified/circulation-patron-web/blob/772683f7d700d3fd005445761bafec3b117e3fc8/src/types/opds1.ts#L115). There are string constants for different auth types defined there. Some of them are not even supported in the web app right now (Implicit grant and Password credentials). You can see the `AppAuthMethods` listed in `interfaces` [here](https://github.com/NYPL-Simplified/circulation-patron-web/blob/ebe47096f199ed7e621653c88c482e1d1a2e7f4d/src/interfaces.ts#L61), which are the auth methods that this app supports. Basically, we have SAML, OAuth, and Basic Auth.

Does that help at all?

src/icons/Close.tsx
kristojorg: Good point. I don't want to change other code in this PR but I probably don't need to add this close icon if there is one in the DS.

src/auth/useCredentials.ts
kristojorg: Yea I'll try and refactor this page so it has fewer overlapping/confusing names

src/auth/AuthForm.tsx
kristojorg: You're right, I moved the loading indicator into the basic auth form (it's the only method that has a loading state), but forgot to strip this out.

src/auth/AuthForm.tsx
kristojorg: Just forgot to remove it, thanks

src/auth/AuthForm.tsx
kristojorg: This is because we are rendering the "buttons" UI instead of one individual form. We are just showing a button for each auth type, which will lead to either the external page for that auth or the form. This is the 3rd gif in the description

kristojorg:
Okay I've refactored the `useCredentials` hook to be hopefully much more clear. Take a fresh look at it. 

I also removed the code that supported clever auth sending an error embedded in the browser url. This code branch was never previously accounted for as seen [here](https://github.com/NYPL-Simplified/circulation-patron-web/blob/b08aa99b57a8cf0210964ad90a692897bfd0c2de/src/hooks/useAuth.ts#L19). We simply assumed that it returned credentials or nothing, which isn't technically true. I was never able to get clever to embed an error in the url either, so it looks like this is outdated functionality copy/pasted over from old code. Long story short, I removed it and things are much simpler now

src/icons/Close.tsx
hokei: yup!  https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/icon--icons

kristojorg:
I also replaced the custom `Close` icon with the NYPL design system one

src/components/context/UserContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+      : null,
+    fetchCollection,
+    {
+      shouldRetryOnError: false,
+      revalidateOnFocus: false,
+      revalidateOnReconnect: false,
+      // clear credentials whenever we receive a 401
+      onError: err => {
+        if (err instanceof ServerError && err?.status === 401) {
+          // clearCredentials();
```
EdwinGuzman, line 50: Is this a todo?

EdwinGuzman (approved):
Great, looks good so far.

src/components/context/UserContext.tsx
kristojorg: Yea, I'm keeping track of todos on my own and will file tickets for them, but yes this is one. There is a small bug in SWR that I've filed a ticket for that is keeping us from doing this. All it means is that if you try to sign in with bad credentials, you'll see the invalid credentials error, and then if you close the modal and re-open it you will still see the same error

PR #139, opened 2020-09-16
Remove locally stored book resources when user logs out

M0nica:
- Locally stored books are removed when a user logs out

### DIFF
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -4,9 +4,7 @@ import { useActions } from "opds-web-client/lib/components/context/ActionsContex
 import { useRouter, NextRouter } from "next/router";
 import useLinkUtils from "components/context/LinkUtilsContext";
 import CleverAuthPlugin from "../auth/cleverAuthPlugin";
-import { NEXT_PUBLIC_COMPANION_APP } from "../utils/env";
 
-const IS_OPEN_EBOOKS = Boolean(NEXT_PUBLIC_COMPANION_APP === "openebooks");
 const SAML_AUTH_TYPE = "http://librarysimplified.org/authtype/SAML-2.0";
 
 const CREDENTIALS_NOT_FOUND = "CREDENTIALS_NOT_FOUND";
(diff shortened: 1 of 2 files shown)


src/hooks/useAuth.ts
```diff
@@ -54,7 +54,17 @@ function useAuth() {
   const isSignedIn = !!authState?.credentials;
   const { actions, dispatch } = useActions();
   const { buildMultiLibraryLink } = useLinkUtils();
-  const signOut = () => dispatch(actions.clearAuthCredentials());
+
+  const clearWebpubViewerStorage = () => {
+    const WebpubViewerStorage = window?.indexedDB?.open("WebpubViewerDb", 2);
```
hokei, line 59: is the version always going to be 2?

M0nica: I was mirroring the version from https://github.com/NYPL-Simplified/webpub-viewer/blob/2bf024c63131117cc153fcd365366fd188f242c7/src/BookResourceStore.ts#L22 

It looks like the version number is optional in this instance so I will update accordingly: https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/open

hokei (changes requested):
I don't think that we can just delete the entire webpub-viewer db because CPW is not the only product that uses webpub-viewer.  Eg:  If a user has DRB up in a different tab, we don't want to make them re-download their book because they logged out of open ebooks.

hokei (approved):
Thanks for digging up the Same Origin Policy on IndexedDB: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Basic_Concepts_Behind_IndexedDB

I think this should be ok, then.  
Please make a different ticket for the edge case re: chapters should not be able to be loaded after the user logs out.

M0nica:
> I don't think that we can just delete the entire webpub-viewer db because CPW is not the only product that uses webpub-viewer. Eg: If a user has DRB up in a different tab, we don't want to make them re-download their book because they logged out of open ebooks.

Since IndexedDB storage has a same-origin policy I think we'd only run into issues deleting the webpub-viewer db across libraries would be in the dev environment if the same port is shared. I can add a check so that the deletion only occurs in the open e-books environment and if other libraries that uses webpub-viewer want to configure something differently the logout logic in c-p-w would need to be updated. 

> IndexedDB follows a same-origin policy. So while you can access stored data within a domain, you cannot access data across different domains. Source: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Basic_Concepts_Behind_IndexedDB

PR #142, opened 2020-09-18
adds environment variables to docker container

hokei:


Dockerfile
```diff
@@ -1,5 +1,9 @@
 # build environment
 FROM node:12.2.0-alpine as builder
+# Set env variables
+ENV NEXT_PUBLIC_COMPANION_APP=openebooks
```
hokei, line 4: From what I can understand, `process.env` doesn't read from the nodejs environment variables, which is what's set in ECS.  It reads them from here.  Therefore, the environment variables aren't set unless they're set here.  

I don't know if this is an ok place to put these environment variables - because the beta merge pushes a publish to `nypl/patron-web`, maybe it's ok to have it be Open Ebooks?  If not though, where else can we put these configs?  Do we need to not be reading `env.ts` from `process.env` ?   We should keep this in mind as we put together the configuration plan.

hokei:
@kristojorg I'm not sure if this is a good idea, but the current `beta` is also untenable as any push may cause a new deploy that breaks environment variables.

Dockerfile
kristojorg: So my understanding is that `nypl/patron-web` is to be the community version, and I don't think it should have (or be necessary to have) env vars hard coded that can't be changed when you actually run the docker container. I'm a bit confused what you mean when you say `process.env` doesn't read from the nodejs env vars... isn't `process.env` _the_ nodejs env vars?

kristojorg (changes requested):
I agree we need to resolve the issues with the env vars before merging to master. However, I don't believe this is a suitable way to do it because it will bread non open-eBooks deployments by setting them to use Open eBooks. We need to have a way to set these env vars when running the docker container. Let's workshop this in our meeting today

hokei:
@kristojorg ok I'm stuuuumped.  
When putting everything in `publicRuntimeConfig`, it won't start up (it says `SIMPLIFIED_CATALOG_BASE` isn't defined).  So I moved it back to see if it works.  The thing is, when running with a local env file, it really does look like all the all the variables are set (in the docker desktop inspector) but it still doesn't show up in the browser 

Screen Shot 2020-09-23 at 5 11 52 PM


This, of course, isn't happening when I'm just running it locally.  I'd love any and all insights ...

kristojorg:
Hmmmm okay I can try to give this some time today. But in general, process.env is only server side, so Nextjs _should_ be exposing those variables on window in the browser. Let me give this a look though

kristojorg:
I should test this by pulling the `test` tag and running that right?

hokei:
@kristojorg Yeah, pull the `nypl/patron-web:test` tag

hokei:
I think there's just some nextjs issues with env variables:  See 
https://github.com/vercel/next.js/issues/7713
and https://github.com/vercel/next.js/discussions/15651

I would love to chat tomorrow morning about stuff I found, and also how we might move forward, because I feel like I've spent way outsized amount of time on this already.

kristojorg:
Okay yea so! I read through all the issues and chatter. I did some testing. I think there _is_ a bug in next.js, but also we need to change the way we are doing this anyways. The reason is that in the near future (after the data layer rewrite), we will be generating some of the pages statically at build time (more on this later), which means that the environment variables will be read at build time, which means there can't be one docker image that everybody uses. Everyone needs their own build. 

So this is my solution: 

1. We move the building of docker images out of github actions and into the docker repository itself. We have a `nypl/openebooks` docker hub repository that has the openebooks env vars set up. Alternatively, NYPL can use Amazon ECR if to rebuild if that's easier for them (since it connects with ECS).
1. I will set it up similarly for Lyrasis, building our own image on beta and master tags. 
1. Anybody else looking to use this/run it will need to do the same. It's a bit unfortunate because it's asking people to go through one more step to build their own version of the app, but I think it's necessary because env vars really are set at build time with a static or partially static application.

.github/workflows/publish-beta.yml
```diff
@@ -7,37 +7,46 @@
...
-#         uses: docker/build-push-action@v1
-#         with:
-#           username: ${{ secrets.DOCKER_USERNAME }}
-#           password: ${{ secrets.DOCKER_PASSWORD }}
-#           repository: nypl/patron-web
-#           # tag the image with both beta and beta.short_sha
-#           tags: beta, beta.${{ steps.get_sha.outputs.short_sha }}
-#           add_git_labels: true
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
```
kristojorg, line 11: Is this all new? I don't remember this getting commented out in beta

jest.config.js
```diff
@@ -132,7 +132,7 @@ module.exports = {
   // runner: "jest-runner",
 
   // The paths to modules that run some code to configure or set up the testing environment before each test
-  // setupFiles: ["<rootDir>src/test-utils/index.tsx"],
+  setupFiles: ["<rootDir>/jest.setup.js"],
```
kristojorg, line 135: I don't think we want to change this

.github/workflows/publish-beta.yml
hokei: The checking out submodules is new, but most of this is commented out -- https://github.com/NYPL-Simplified/circulation-patron-web/blob/beta/.github/workflows/publish-beta.yml#L11

jest.setup.js
```diff
@@ -0,0 +1,3 @@
+const mockEnvConfig = {}
+
+jest.mock('next/config', () => () => ({ publicRuntimeConfig: mockEnvConfig }));
```
kristojorg, line 3: You should update `test-utils/setEnv` to do this and set the env var to whatever a tester needs

src/pages/404.tsx
```diff
@@ -1,5 +0,0 @@
-import * as React from "react";
-import ErrorComponent from "../components/Error";
-export default function NotFoundPage() {
-  return <ErrorComponent statusCode={404} />;
-}
```
kristojorg, line 5: Why was this deleted?

src/utils/env.ts
```diff
@@ -1,32 +1,35 @@
...
-export const REACT_AXE = process.env.REACT_AXE;
+export const REGISTRY_BASE = publicRuntimeConfig.REGISTRY_BASE;
+export const CIRCULATION_MANAGER_BASE =
+  publicRuntimeConfig.SIMPLIFIED_CATALOG_BASE;
+export const CONFIG_FILE = publicRuntimeConfig.CONFIG_FILE;
+export const REACT_AXE = publicRuntimeConfig.REACT_AXE;
 // if there is not circ manager base, we are using multi library routes
 export const IS_MULTI_LIBRARY = !CIRCULATION_MANAGER_BASE;
 export const CACHE_EXPIRATION_SECONDS = parseInt(
   process.env.CACHE_EXPIRATION_SECONDS ?? "10",
```
kristojorg, line 21: Should this not also be in the publicRuntimeConfig? I think every env var should be there now

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "importHelpers": true,
     "baseUrl": "src"
   },
-  "exclude": ["node_modules", "config/srv", "jest.config.js"],
+  "exclude": ["node_modules", "config/srv", "jest.config.js", "jest.setup.js"],
```
kristojorg, line 21: Don't think we should do this

kristojorg (changes requested):
A few changes! Does this work now though?

.github/workflows/publish-beta.yml
kristojorg: I see, I guess I missed that PR

src/pages/404.tsx
hokei: gaaaah I totally thought I put this back in.

hokei:
Ran it again and it works!

next.config.js
```diff
@@ -10,6 +10,8 @@ const APP_VERSION = require("./package.json").version;
 
 const config = {
   env: {
+    NEXT_PUBLIC_AXIS_NOW_DECRYPT: process.env.NEXT_PUBLIC_AXIS_NOW_DECRYPT,
+    NEXT_PUBLIC_COMPANION_APP: process.env.NEXT_PUBLIC_COMPANION_APP,
```
kristojorg, line 14: I actually don't think these should be necessary here. The `NEXT_PUBLIC_` prefix does this implicitly, which is why I started migrating to use it

package.json
```diff
@@ -14,6 +14,7 @@
     "dev:axe": "REACT_AXE=true npm run dev",
     "build": "next build",
     "start": "next start",
+    "build-and-start": "npm run build && npm run start",
```
kristojorg, line 17: The one other thing is that previously we got rid of all but runtime files when building the docker container. I don't know if this is still a requirement or not (I can ask lyrasis and maybe you can ask NYPL), but the function of it was to minimize the size of the image being run, because apparently that does have an impact on pricing

kristojorg (changes requested):
Just a couple more things

package.json
kristojorg: What I was originally thinking is that once we confirmed this works, we would make a script file that does:
1. `npm run build`
1. `npm run prune`
1. copy `.next` and `package.json` over to a new directory
1. delete the old directory and everything in it
1. `npm run start` in the new directory

kristojorg: I did confirm that Lyrasis would like this, but that it isn't blocking. You can add it in a follow on PR if you need to get this in ASAP

hokei: let's confirm that the script actually works/deploys first.

kristojorg (approved):
Okay I'll approve if you wouldn't mind adding a ticket to remove unneeded files in the future!

kristojorg:
Also changelog please!

hokei:
OK filed https://jira.nypl.org/browse/SIMPLY-3135

PR #143, opened 2020-09-23
#4 Data Layer - Auth tests

kristojorg:
This PR adds tests for the new auth and loans flow as a follow on to #138


src/auth/AuthForm.tsx
```diff
@@ -217,4 +220,16 @@ const Combobox: React.FC<{
...
+function getIdForMethod(method: AppAuthMethod) {
+  return method.type === OPDS1.SamlAuthType ? method.href : method.type;
+}
+function getMethodForId(
+  authMethods: AppAuthMethod[],
+  id: string
+): AppAuthMethod | undefined {
+  return authMethods.find(method => method.type === id || method.href === id);
+}
+
```
kristojorg, line 234: These tests helped me discover a bug! All the ids for SAML methods were the same, which wasn't good. Now we use the href for saml and type for non saml. We should look at eventually adding an id on the server in case there are multiple with the same type

src/auth/__tests__/AuthForm.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+test("shows form when only one auth method configured", () => {
+  const utils = render(<AuthForm>child</AuthForm>, {
+    library: {
+      ...fixtures.libraryData,
+      authMethods: oneAuthMethod
+    }
+  });
+
+  expect(
+    utils.getByRole("textbox", { name: "Barcode input" })
```
hokei, line 57: can you expect other methods not to be here?

kristojorg: Good point, will do

src/auth/AuthForm.tsx
```diff
@@ -15,7 +15,6 @@ import SamlAuthButton from "auth/SamlAuthButton";
 import CleverButton from "auth/cleverAuthButton";
```
hokei, line 15: nit: capitalize this file name

src/auth/AuthForm.tsx
```diff
@@ -15,7 +15,6 @@ import SamlAuthButton from "auth/SamlAuthButton";
...
 import { AuthFormProvider } from "auth/AuthFormCotext";
```
hokei, line 16: nit: spelling

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+import * as React from "react";
+import { AuthCredentials, BookData, OPDS1 } from "interfaces";
+import { act, fixtures, render, waitFor } from "test-utils";
+import Cookie from "js-cookie";
+import * as router from "next/router";
+import useUser from "components/context/UserContext";
+import * as fetch from "dataflow/opds1/fetch";
+/**
+ * This file tests both UserContext and useCredentials, as
+ * the latter is only used within the former. It doesn't have
```
hokei, line 11: general react question:  Usually when that's the case they should just be the same thing, are they separated out because `useCredentials` is a hook ... ... ?

src/auth/__tests__/BasicAuthForm.test.tsx
```diff
@@ -15,28 +17,10 @@ const method: OPDS1.BasicAuthMethod = {
...
-};
+beforeEach(() => fetchMock.resetMocks());
 
 test("displays form", () => {
-  const utils = render(<BasicAuthForm provider={provider} />, {
-    initialState: merge<State>(fixtures.initialState, {
-      auth: authState
-    })
-  });
+  const utils = render(<BasicAuthForm method={method} />);
```
hokei, line 23: ok let me know if this is all too pedantic,  -- I've mentioned in the past that this piece trips me up every time.  

One thing that would help is fleshing out the comment in `BasicAuthForm.tsx`.  How does it differ from `AuthForm.tsx`?  

The `method={method}` makes me think it's doing something like `<BasicAuthForm method=method.SAML1 />` or something, but that's not what it's doing, right?  `<BasicAuthForm />` should only be used with `BasicAuthMethod` ... right? 

So, for:  
```
export interface BasicAuthMethod extends AuthMethod<typeof BasicAuthType> {
  labels: {
    login: string;
    password: string;
  };
}
```

Is what's being passed into method here ... credentials?  Or is it something else?

hokei:
In general:  All the auths use `AuthFormContext.tsx`, right?  
`<AuthForm>` is a wrapper that contains a sign in for `Basic`, `Clever` and `SAML` methods (whichever is applicable?).  Is there a way to make it clearer that `BasicAuthForm`, `CleverAuthButton` and `SAMLAuthButton` are related to each other and that the relationship between `AuthForm` and `CleverAuthButton` is the same as the relationship between `AuthForm` and `BasicAuthForm`?

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  function Extractor() {
+    const { signOut, loans } = useUser();
+    extractedSignOut = signOut;
+    extractedLoans = loans;
+    return <div>hello</div>;
+  }
+  render(<Extractor />);
+
+  // make sure fetch was called and you have the right data
+  expect(fetch.fetchCollection).toHaveBeenCalledTimes(1);
```
EdwinGuzman, line 130: I'm missing something. `fetch.fetchCollection` gets called when `useUser` is invoked?

src/auth/AuthForm.tsx
EdwinGuzman: Shouldn't it also throw an error? I guess since we're ignoring that tests pass in this PR that this wasn't caught.

EdwinGuzman:
I have been mostly reviewing the code and getting an overview of everything, but I can't say I have a 100% grasp on all the logic or how components relate to each other. So +1 to @hokei 's questions.

src/components/context/__tests__/UserContext.test.tsx
kristojorg: Yea these could be the same thing, I'm just trying to split them up a bit into logical chunks so they're easier to work with. `UserContext` deals with fetching the loans, and it would have just been a huge component if it also dealt with getting the credentials from cookie and syncing that with state. So I split that out into a separate hook, but it's not for reusability since it only ever is used here, it's for encapsulation

src/components/context/__tests__/UserContext.test.tsx
kristojorg: It does, because the `UserContext` component deals with fetching the user's loans. Their loans are just a regular OPDS feed that is protected behind auth, so `UserContext` uses the same `fetchCollection` function to fetch loans as a page will use to fetch a collection for the page. Does that make sense?

kristojorg:
Okay I see I should have been more clear about what's going on here. Let me try to clarify. This is the way authentication and user data is handled in this app now:

- The `UserContext` component is responsible for determining if a user is authenticated, and fetching their loans if they are. It also provides functions to refetch the loans and to update the loans (`setBook`).
- `setBook` is useful when you borrow a book and want to optimistically set it in the users loans before refetching the full loans state from the server.
- We fetch the authentication document at startup, and set the available `authMethods` in the `LibraryContext`. The are `AppAuthMethod` objects which describe the methods available and how to log in with them. 
- Each `OPDS1.AuthMethod` has a type and optional description and links. Specific auth methods can have more properties. For example:
  - The `BasicAuthMethod` has `labels` which is an object that tells you what to call the "username" and "password" fields. For example, many times it is set to `Barcode` and `Pin`. So you'll see the inputs labeled "Barcode" and "Pin" instead of "Username" and "Password".
- The SAML auth method is particular. It will contain a list of individual SAML identity providers. From a UX perspective, these are basically just different methods, so we flatten the nested identity providers out into `ClientSamlProvider`s, which get set in the `LibraryContext` `authMethods` array alongside any others.
- The `AuthForm` component should maybe be called `AuthModal`. It is the modal that pops up to let you sign in. It takes a list of `AppAuthMethods` and decides how to display them to the user depending how many there are. 
- The `AuthFormContext` just passes down a function to let children show the auth form. This is used, for example, when you try to borrow a book but aren't logged in. That `useBorrow` hook will get the `showForm` from context and show the auth modal. 
- The `BasicAuthForm` is the only one of these that is an actual "form", I guess. It's the UI for letting people sign in with basic auth, which takes a username/password combo. Whereas the Clever just redirects you to an external site to do auth, so it just has the `CleverAuthButton`... a button that redirects you properly. 

So what I will do is rename the `AuthForm` to `AuthModal` to try and reduce confusion. Does this help at all?

src/auth/__tests__/BasicAuthForm.test.tsx
kristojorg: Tried to explain below, but in case that is confusing, what's being passed in here are the _labels_ for the login and password field. EG they could be called "Barcode" and "Pin". This is a setting that can be configured in the circulation manager

EdwinGuzman:
Yes, that will help. Thanks for the write up!

kristojorg:
Okay I've renamed `AuthForm` to `AuthModal` and all associated references

src/auth/AuthForm.tsx
kristojorg: Dont actually know how this didn't show an error... maybe the file was misnamed actually. In any case it's now `AuthModalContext` and it's updated

kristojorg:
And fixed the `cleverAuthButton` filename issue

src/auth/__tests__/BasicAuthForm.test.tsx
hokei: Is "Method" just a word with a special meaning in this context, in which the it means "params"?  From what I'm seeing in `interfaces.ts`. AuthMethods are just objects that hold information, like type/description/links/labels, right?

src/auth/__tests__/SamlAuthButton.test.tsx
```diff
@@ -56,7 +35,7 @@ test("redirects to idp", async () => {
   await waitFor(() => {
     expect(window.open).toHaveBeenCalledTimes(1);
     expect(window.open).toHaveBeenCalledWith(
-      "/saml-auth-url&redirect_uri=http%3A%2F%2Ftest-domain.com%2F",
+      "/saml-auth-url/0&redirect_uri=http%3A%2F%2Ftest-domain.com%2F",
```
hokei, line 38: why the 0?

hokei:
OK this looks a lot clearer to me!  I'm happy to move forward if Edwin doesn't have other big concerns!

src/auth/__tests__/SamlAuthButton.test.tsx
kristojorg: This was just what I had in the test fixture... had the index of the idp in the url so I could differentiate

EdwinGuzman:
Yes, same. But it'd be good to find out about the "0" in the url.

src/auth/__tests__/SamlAuthButton.test.tsx
kristojorg: Basically because there are multiple saml idps in the text fixture (or there can be), so I put the number in there to differentiate which url is for which idp. They're generated by a fixture method, so have names like "Saml IDP 0` and urls like `/saml-auth-url/0` or `/saml-auth-url/1` etc

hokei: ah yes I see it it now.  I guess that `auth.ts` change was made in the previous PR?

hokei (approved):

PR #145, opened 2020-09-23
#5 Data Layer - Borrowing + Reserving

kristojorg:
This PR moves borrowing and reserving from redux actions directly into the `useBorrow` hook. A few things to note:

1. If you are not signed in, you are immediately presented a sign in form when trying to borrow (we don't have to wait for the 401 to come back anymore)
1. We now extract error details properly from the problem document sent from the server if a request fails and display them to the user.
1. This includes tests that fully cover this code.


src/components/context/UserContext.tsx
```diff
@@ -61,24 +61,10 @@ export const UserProvider: React.FC = ({ children }) => {
...
-          ...data,
-          books: [...data?.books, book]
-        },
-        false
-      );
-    } else {
-      mutate();
-    }
+    const newData: BookData[] = [...data, book];
+    mutate(newData);
```
kristojorg, line 67: This sets the book returned from the borrow request into the loans cache, and simultaneously triggers a revalidation with the server to make sure our update worked.

src/hooks/useBorrow.ts
```diff
@@ -28,7 +28,27 @@ export default function useBorrow(
...
+      showForm();
+      setError("You must be signed in to borrow this book.");
+      setLoading(false);
+      return;
+    }
+    try {
+      const book = await fetchBook(url, catalogUrl, token);
+      setBook(book);
+    } catch (e) {
+      // TODO: Report error to bug catcher here.
```
kristojorg, line 43: We will do this once error reporting is merged

src/hooks/useBorrow.ts
```diff
@@ -28,7 +28,27 @@ export default function useBorrow(
 
   const borrowOrReserve = async (url: string) => {
     setLoading(true);
-    await dispatch(actions.updateBook(url));
+    setError(undefined);
+    if (!token) {
+      // TODO: register a callback to call if the sign in works
```
kristojorg, line 33: This would be nice to have (to immediately retry your borrow once you sign in), but it isn't necessary immediately.

src/components/context/UserContext.tsx
```diff
@@ -34,12 +34,13 @@ export const UserProvider: React.FC = ({ children }) => {
   const { credentials, setCredentials, clearCredentials } = useCredentials(
     slug
   );
+  console.log(credentials);
```
EdwinGuzman, line 37: Extra console.

src/components/context/UserContext.tsx
EdwinGuzman: What do you mean "to make sure our update worked"? The update here means that it will always refetch loans, right?

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -6,6 +6,8 @@ import Cookie from "js-cookie";
 import * as router from "next/router";
 import useUser from "components/context/UserContext";
 import * as fetch from "dataflow/opds1/fetch";
+import mockAuthStateOnce, { creds } from "test-utils/mockAuthState";
```
EdwinGuzman, line 9: Looks like `mockAuthStateOnce` is not used in this file.

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -50,15 +47,15 @@ beforeEach(() => {
 });
 
 test("fetches loans when credentials are present", () => {
-  mockCookie.get.mockReturnValueOnce(str(credentials));
+  mockAuthenticatedOnce();
   renderUserContext();
 
   expect(fetch.fetchCollection).toHaveBeenCalledTimes(1);
-  expectFetchWithToken(credentials.token, OPDS1.BasicAuthType);
```
EdwinGuzman, line 57: Why is the type no longer relevant?

src/auth/__tests__/AuthForm.test.tsx
```diff
@@ -18,14 +17,20 @@ useDialogSpy.mockReturnValue({
...
 
+/**
+ * We don't have to directly render the auth form because it is already
+ * rendered in the ContextProvider, which our custom `render`
+ * already wraps everything with.
+ */
+
 test("renders header and subheader", () => {
-  const utils = render(<AuthForm>child</AuthForm>);
+  const utils = render(<div>child</div>);
```
hokei, line 27: can we call "render" something less generic, then?  At a glance this reads like black magic

src/components/context/__tests__/UserContext.test.tsx
kristojorg: Ok so `swr` takes the `url`, `token` and `type` as keys. So whenever one of those changes, it calls `fetchCollection` again, and it passes in all three keys. The tricky thing is that `fetchCollection` only takes the `url` and `token`, it doesn't care about the type. This is a normal use of swr where you have something that you data depends on, but your fetcher doesn't actually use it. SWR passes all three keys as arguments, but because it's javascript, you can call a function that takes two arguments with three arguments and everything is kosher.

_However_, I have now wrapped `fetchCollection` in `fetchLoans` inside `UserContext`, and so the inner `fetchCollection` is only ever passed the two params it actually uses anymore. So the mock only gets two params, whereas previously the mock saw all three params even though the function only took two arguments... weird but that's javascript I guess

src/auth/__tests__/AuthForm.test.tsx
kristojorg: I could call it `renderWithContext`, but then it should be changed in every test file. Every single test in the app is currently wrapped in the `ContextProvider`... and we also add other utils to `render`. This is why we import it from `test-utils/index.tsx` instead of `@testing-library/react`. I'm not opposed to changing the name, but if we're gonna do that we should do it in a follow up PR to keep this git diff clean since that will change... lots of files

EdwinGuzman:
It looks god but it seems you merged in merge conflicts.

src/hooks/useBorrow.ts
hokei: didn't error reporting get merged?

src/auth/__tests__/AuthForm.test.tsx
hokei: ok, let's do that at the end then.  I think it would help -- if it's not obvious already I tend to rely on unit tests a lot as documentation haha.

src/hooks/useBorrow.ts
kristojorg: Yep after this PR was made. Since I have so many PRs stacked on top of each other, I'll do this once we get them merged into the datalayer branch

kristojorg:
Thanks for catching that @EdwinGuzman , I've fixed those merge conflicts

src/components/context/UserContext.tsx
kristojorg: Basically what we do is make the update request, and get back a response, and then refetch the loans by calling mutate. This isn't technically necessary because we update them client side with what the server responded with in our update request, but it just makes 100% sure we are looking at the same loans state as the server

kristojorg: It will always refetch loans yes

EdwinGuzman (approved):

PR #147, opened 2020-09-25
#6 Data Layer - Datalayer tests

kristojorg:
This gets mosts of the tests running and fixes most of the typescript issues. The exception here is download/read online logic which is next to come!

image

Test coverage is now at 65%, which will grow quite a lot very soon when we remove the rest of the code in the `owc/` folder:
image


src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -348,82 +225,31 @@ describe("ready to borrow (two links)", () => {
...
-      name: /Loading Label/i
     });
-    expect(borrowButton).toBeInTheDocument();
-    expect(borrowButton).toHaveAttribute("disabled", "");
 
-    const unClickedBorrowButton = await utils.getByText(
-      "Borrow to read online"
-    );
-    expect(unClickedBorrowButton).toBeInTheDocument();
-  });
```
kristojorg, line 403: This test and many similar ones were removed because this logic is now tested in BorrowOrReserve.test.tsx, so those tests don't need to be duplicated here

src/auth/__tests__/cleverAuthButton.test.tsx
```diff
@@ -1,6 +1,11 @@
 import * as React from "react";
-import { fixtures, render } from "test-utils";
+import { fixtures } from "test-utils";
 import CleverButton from "../cleverAuthButton";
+/**
+ * we import reder from testing library because we don't need our
```
EdwinGuzman, line 5: Small typo in `reder`.

src/test-utils/mockUser.ts
```diff
@@ -0,0 +1,24 @@
...
+  status: "authenticated",
+  refetchLoans: jest.fn(),
+  signIn: jest.fn(),
+  signOut: jest.fn(),
+  setBook: jest.fn(),
+  error: undefined,
+  token: "user-token"
+};
+
+export default function mockUser(data: Partial<typeof defaultUser> = {}) {
```
EdwinGuzman, line 18: This is nice, but don't you need to reset the spy?

EdwinGuzman (approved):

src/test-utils/mockUser.ts
kristojorg: We shouldn't have to because we have this set in our `jest.config.js`:
```
 // Automatically clear mock calls and instances between every test
  clearMocks: true
```

PR #150, opened 2020-09-28
#7: Fulfilling books

kristojorg:
This PR handles fulfilling books, moving the logic out of redux / wherever it was scattered across the codebase. Here is an explanation of how this works now:

- The `FulfillmentCard` turns the array of `MediaLink`s into an array of `FulfillmentDetails` via the `getFulfillmentDetails` function. 
- The `getFulfillmentDetails` function is where I am starting to centralize the work of parsing a `type` and `indirectType` combination into a specific method for fulfilling the book. Currently, there are four ways to fulfill a book: read online, read online external, download, and unsupported. These map to specific user experiences. 
- I try to handle the indirection logic inside the `getFulfillmentDetails` function as well, so that our component is just passed exactly the information it needs to perform the fulfillment. One example of this is handling the case of a link's `type` being set to `OPDS1.OPDSEntryMediaType`, which indicates that we need to fetch an opds entry at that url, and then extract a short-lived url from it to fetch the actual resource. Currently, this is only handled for Overdrive read online books, but it should actually be a general solution. We could have the same indirection on any media. This generalized indirection is something we need to continue thinking about. 
- I now parse server errors returned during fulfillment and display them to the user. 
- All tests should now be passing (though we still have work to do to get coverage back up... will come soon)


src/components/__tests__/BookListItem.test.tsx
```diff
@@ -205,7 +205,7 @@ describe("ready to borrow book with multiple borrowUrls", () => {
       {
         url: "/adobe-borrow-link",
         type: "application/atom+xml;type=entry;profile=opds-catalog",
-        indirectType: "application/vnd.adobe.adept+xml"
```
kristojorg, line 208: We were never testing anything to do with the indirect type in these tests, and after making our types more specific, this type no longer qualifies as an indirectType, so I updated it to one that does.

src/components/BorrowOrReserve.tsx
```diff
@@ -4,12 +4,12 @@ import * as React from "react";
 import useBorrow from "hooks/useBorrow";
 import Button from "./Button";
 import { Text } from "./Text";
-import { BookData, FulfillmentLink } from "interfaces";
+import { BookData, MediaLink } from "interfaces";
```
kristojorg, line 7: I used to differentiate between `MediaLink` and `FulfillmentLink`, but that was before I really understood what these are. They're not actually different. So everything is now a `MediaLink`.

kristojorg: I might rename to AcquisitionLink in a future PR because that is more accurate

src/components/bookDetails/FulfillmentCard.tsx
```diff
@@ -250,24 +259,31 @@ const AccessCard: React.FC<{
           <Text>{subtitle}</Text>
         </Stack>
       </Stack>
-      {!isAudiobook && (
+      {!isAudiobook && isFulfillable && (
```
kristojorg, line 262: We now don't show this whole block if there are no supported fulfillment methods.

src/dataflow/download.ts
```diff
@@ -0,0 +1,43 @@
+import fetchWithHeaders from "dataflow/fetch";
```
kristojorg, line 1: This is all refactored from opds-web-client's `fulfillBook` action

src/dataflow/download.ts
```diff
@@ -0,0 +1,43 @@
...
+import download from "downloadjs";
+import { ServerError } from "errors";
+import { OPDS1 } from "interfaces";
+import { generateFilename, typeMap } from "owc/utils/file";
+
+export default async function downloadFile(
```
kristojorg, line 7: This is all refactored from opds-web-client's `fulfillBook` action

src/dataflow/opds1/parse.ts
```diff
@@ -61,14 +61,14 @@ const resolve = (base: string, relative: string) =>
...
     const indirects = link.indirectAcquisitions;
     const first = indirects[0];
-    const indirectType = first?.type as string | undefined;
+    const indirectType = first?.type as OPDS1.AnyBookMediaType | undefined;
     // it is possible that it doesn't exist in the array of indirects
     return {
       url: resolve(feedUrl, link.href),
-      type: link.type as OPDS1.AnyBookMediaType,
+      type: link.type as OPDS1.AnyBookMediaType | OPDS1.IndirectAcquisitionType,
       indirectType
```
kristojorg, line 72: This is confusing, but the way this has been parsed means that when the content is behind some indirection, the ultimate content type is on `indirectType` and the initial type is on `type`

src/hooks/useDownloadButton.ts
```diff
@@ -0,0 +1,37 @@
...
+import useUser from "components/context/UserContext";
+import downloadFile from "dataflow/download";
+import { ServerError } from "errors";
+
+export default function useDownloadButton(
+  details: DownloadDetails,
+  title: string
+) {
+  const [loading, setLoading] = React.useState(false);
+  const [error, setError] = React.useState<string | null>(null);
```
kristojorg, line 12: We now show loading and error states when trying to download! Hooray : )

src/types/opds1.ts
```diff
@@ -32,21 +32,42 @@ export const HTMLMediaType = "text/html";
...
+
+export type IndirectAcquisitionType =
+  | typeof OPDSEntryMediaType
+  | typeof AdeptMediaType
+  | typeof IncorrectAdeptMediaType;
+
+/**
+ * Direct Acquisition Types
+ * These are the final types that resources can be.
+ */
```
kristojorg, line 62: I've started splitting up what was previously one type into the indirect and the direct types that media can have.

src/owc/utils/file.ts
```diff
@@ -1,25 +1,17 @@
...
 export const typeMap: Record<
-  OPDS1.AnyBookMediaType,
+  OPDS1.DownloadMediaType,
   { extension: string; name: string }
 > = {
   "application/vnd.overdrive.circulation.api+json;profile=audiobook": {
     extension: "",
     name: "Overdrive Audiobook"
   },
-  "application/vnd.overdrive.circulation.api+json;profile=ebook": {
```
EdwinGuzman, line 11: Why were all these deleted?

src/utils/fulfill.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+export function fixMimeType(
+  mimeType: OPDS1.IndirectAcquisitionType | OPDS1.AnyBookMediaType
+) {
+  return mimeType === "vnd.adobe/adept+xml"
+    ? "application/vnd.adobe.adept+xml"
+    : mimeType;
+}
+
+export function dedupeLinks(links: MediaLink[]) {
```
EdwinGuzman, line 145: Don't think you're using this function in this file.

EdwinGuzman (approved):
Looks good, just minor comments. Thanks for all the in-depth comments!

src/owc/utils/file.ts
kristojorg: Good question - because I've narrowed the types that book media can be. There are now three types: `IndirectAcquisitionType`, `ReadOnlineMediaType` and `DownloadMediaType`. The deleted ones are either `IndirectAcquisitionType`s or `ReadOnlineMediaType`s. So they wouldn't have an extension for example, and shouldn't ever be passed in to this map. The map is now specifically only for `DownloadMediaType`s

src/utils/fulfill.ts
kristojorg: It's being imported elsewhere : ) . Specifically in the fulfillment card I believe

PR #151, opened 2020-09-28
#8: Fetch Search description at startup

kristojorg:
Previously, we would fetch the data needed to perform a search when displaying a given page. Instead, since the search is library-wide, we simply fetch that data at startup and put it into the `LibraryContext`.

This brings us one step closer to removing redux completely, woohoo!


src/dataflow/opds1/parseSearchData.ts
```diff
@@ -0,0 +1,33 @@
+import { SearchData } from "interfaces";
+import * as xml2js from "xml2js";
+
+const xmlParser = new xml2js.Parser({ xmlns: true });
+
+/** Converts an open search description document into the application's internal
+    representation. */
+export default function parseSearchData(
```
EdwinGuzman, line 8: Is this new or from a previous file?

EdwinGuzman (approved):
Just a minor question

src/dataflow/opds1/parseSearchData.ts
kristojorg: This is from opds-web-client. It used to be the `OpenSearchDataParser` class, with a single method... so I just removed the class wrapper to make it a simple function and moved it into CPW

PR #152, opened 2020-09-28
#9: Collection and books

kristojorg:
This PR accomplishes a few things:

1. Move book and collection fetching logic to SWR (out of redux)
2. Move the recommendations fetching to SWR (out of redux)
3. Make breadcrumbs work without redux (they still don't work well, but work as well as before)
4. Remove all redux code and the remainder of code in the `/owc` folder
5. Parse the `relatedUrl` for a book in `parse` instead of extracting it from `raw` afterwards.
5. Get existing tests to pass

## Collection Fetching

The collection is fetched in the `Collection` component using the url parameters from the browser. It then passes this collection down to other components that need it, like the breadcrumbs.  In the case of a list view, we need to add infinite scrolling pagination, so we use the `useSWRInfinite` hook to do this pagination. We used to have a `useInfiniteScroll` hook, but all the logic has been moved directly inside `InfiniteBookList` now. 

## Book Fetching

This is done similarly to the collection fetching. One notable thing we changed: we have always passed the raw, unparsed `entry` through to the application in the `BookData` interface. This was used to extract other values that the `OPDSDataAdapter` didn't parse for us. Since we now have the parsing logic in this repository, I've updated the `entryToBook` function to extract the `relatedUrl` for fetching recommendations. In the future we will be doing this more so that we don't have to pass a `raw` value around and unsafely dig into it. 

## Testing

I've started mocking `useSWR` in my tests because it turns out it's pretty difficult not to get the cache they are using not to leak between tests. Mocking the return value of the hook itself is much simpler, and still gives us good confidence.


src/__tests__/computeBreadcrumbs.test.tsx
```diff
@@ -32,7 +30,7 @@ describe("computeBreadcrumbs", () => {
       { url: "breadcrumb url", text: "breadcrumb title" },
       { url: collection.url, text: collection.title }
     ];
-    expect(computeBreadcrumbs(data, history)).toEqual(expected);
```
kristojorg, line 35: Using the history in the breadcrumbs didn't work previously. I've removed it here and we will need to make a ticket to eventually make the breadcrumbs better using navigation history.

src/components/BreadcrumbBar.tsx
```diff
@@ -1,18 +1,24 @@
...
+  breadcrumbs?: LinkData[];
+}> = ({ children, className, currentLocation, breadcrumbs }) => {
+  const { catalogUrl, catalogName } = useLibraryContext();
+
   const breadcrumbsWithAtLeastOne =
-    breadcrumbs.length > 0 ? breadcrumbs : [{ text: " ", url: " " }];
-  const lastItem = currentLocation ?? breadcrumbs.pop()?.text;
+    breadcrumbs && breadcrumbs.length > 0
+      ? breadcrumbs
+      : [{ text: catalogName, url: catalogUrl }];
```
kristojorg, line 19: If there are no breadcrumbs, we default to the catalog root

src/components/bookDetails/index.tsx
```diff
@@ -1,50 +1,48 @@
...
+  const { data, error } = useSWR(bookUrl ?? null, fetchBook);
+  const { loans } = useUser();
+  // use the loans version if it exists
+  const book = loans?.find(loanedBook => data?.id === loanedBook.id) ?? data;
 
   if (error) {
-    return <Error error={error} />;
+    if (error instanceof ServerError) {
+      return <Error info={error.info} />;
+    }
```
kristojorg, line 40: We now parse server errors properly, yay!

src/computeBreadcrumbs.tsx
```diff
@@ -40,10 +36,60 @@ const computeBreadcrumbs = (
...
 
 export default computeBreadcrumbs;
+
+/**
+ * Computes breadcrumbs assuming that the OPDS feed is hierarchical - uses
+ * the catalog root link, the parent of the current collection if it's not
+ * the root, and the current collection. The OPDS spec doesn't require a
+ * hierarchy, so this may not make sense for some feeds.
+ * */
+export function hierarchyComputeBreadcrumbs(
```
kristojorg, line 53: This was copied over from opds-web-client

hokei:
Screen Shot 2020-10-02 at 10 53 55 AM
Screen Shot 2020-10-02 at 10 54 40 AM


Was this change intended?

hokei:
Logging in with Clever seems to be broken.  Maybe it happened in this, maybe it happened in the auth work previously.

src/components/WebpubViewer.tsx
```diff
@@ -1,31 +1,24 @@
...
+    const loadDecryptor = async (webpubManifestUrl: any) => {
       const Decryptor = await import(
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         "../../axisnow-access-control-web/src/decryptor"
       );
       if (Decryptor) {
         try {
-          const fulfillmentData = await fetcher.fetch(webpubManifestUrl);
+          const fulfillmentData = await fetch(webpubManifestUrl);
```
hokei, line 21: Something about this change is not passing in the username and password into this fetch.  As a result, `fulfillmentData` gets the "you are not logged in" value, and the book cannot be fulfilled.

src/components/BookList.tsx
```diff
@@ -39,27 +33,77 @@ const ListLoadingIndicator = () => (
...
 
 type BookWithUrl = RequiredKeys<BookData, "url">;
 const hasUrl = (book: BookData): book is BookWithUrl => !!book.url;
 
-export const ListView: React.FC<{
-  books: BookData[];
-}> = ({ books }) => {
-  // this hook will refetch the page when we reach the bottom of the screen
-  const { listRef, isFetchingPage } = useInfiniteScroll();
+export const InfiniteBookList: React.FC<{ firstPageUrl: string }> = ({
```
hokei, line 43: +1 for the new name

src/components/BookList.tsx
```diff
@@ -19,14 +18,9 @@ import useIsBorrowed from "hooks/useIsBorrowed";
...
-
-/**
- * In a collection you can:
- *  - See lanes view
- *  - See List/Gallery view
- *    - Switch between list and gallery in this case
- */
+import { BookData, CollectionData, LaneData, RequiredKeys } from "interfaces";
+import { fetchCollection } from "dataflow/opds1/fetch";
+import { useSWRInfinite } from "swr";
```
hokei, line 23: what does swr stand for?

src/components/BookList.tsx
```diff
@@ -39,27 +33,77 @@ const ListLoadingIndicator = () => (
...
+}> = ({ books, listRef, isFetchingMore = false }) => {
   return (
     <React.Fragment>
       <ul ref={listRef} sx={{ px: 5 }} data-testid="listview-list">
         {books.map(book => (
           <BookListItem key={book.id} book={book} />
         ))}
       </ul>
-      {isFetchingPage && <ListLoadingIndicator />}
+      {isFetchingMore && <ListLoadingIndicator />}
```
hokei, line 106: Is this the same loading indicator on the initial list load, or is that loading indicator somewhere else?

src/components/Collection.tsx
```diff
@@ -1,36 +1,40 @@
...
-  const hasBooks = collectionData?.books && collectionData.books.length > 0;
+  const { data: collection, isValidating } = useSWR(
+    collectionUrl,
+    fetchCollection
+  );
+
+  const isLoading = !collection && isValidating;
 
-  const pageTitle = title ?? `Collection: ${collectionData.title ?? ""}`;
+  const hasLanes = collection?.lanes && collection.lanes.length > 0;
```
hokei, line 33: Is it possible for collection to be null here?  Can we do a check under `useSWR` to throw an error and then eliminate the need to check every time we use collection?

hokei (changes requested):
There's a few things that seem to be not working (most importantly, the fulfill link is being called without credentials)

src/components/ListFilters.tsx
```diff
@@ -1,30 +1,34 @@
...
 import Router from "next/router";
 import useLinkUtils from "./context/LinkUtilsContext";
+import { CollectionData } from "interfaces";
 
 /**
  * This filter depends on the "Sort by" and "Availability" facet groups.
  * They must be named exactly that in the CM in order to show up here.
  */
-const ListFilters: React.FC = () => {
+const ListFilters: React.FC<{ collection: CollectionData }> = ({
```
hokei, line 13: are these being used anywhere?  I'm having trouble finding references to it.

src/components/BookList.tsx
kristojorg: It stands for stale-while-revalidate, which is basically the cache strategy that this hook uses. It serves you the stale data while revalidating with the backend and then updating the UI once the new data comes in. You can see more on: https://swr.vercel.app/

src/components/Collection.tsx
kristojorg: It is possible for it to be `undefined`, when no data has been returned yet. I could return the loading state early in that case, if that's what you mean?

src/components/BookList.tsx
kristojorg: This is supposed to only be shown when loading more... the reason is that the `<Collection>` component makes the initial request for the collection using SWR, and then we only render this infinite list if it has `books` instead of `lanes`. This infinite list then takes the collection url as the first page and requests it again. Since this initial url was already loaded by the `<Collection>` component, SWR returns the cached version immediately, therefore not showing this loading indicator. 

The initial loading indicator is in the `<Collection>` compoent

src/components/ListFilters.tsx
kristojorg: They are not actually right now. I took them out at one point as we migrated to the new designs, but kept them in the codebase as I knew we would want them eventually. These are the filter selectors we will use to add the new filters for Open eBooks.

src/components/WebpubViewer.tsx
kristojorg: Okay yep my mistake. I now pass the token in to this function from the main application so we can use it to fetch with. Previously it was stored internally in the `fetcher` class, which we no longer use.

kristojorg:
Okay, good catches. 

1. I fixed the missing background url on the FirstBook auth button. That was my mistake.
1. I passed the token in to the fetcher, though I haven't been able to test it yet because I don't know how to find a book I can open in the reader. Also, I had to add a new fallback state that is unfinished in case the user isn't signed in and gets to that point. How should we deal with that?

I haven't been able to reproduce the errors you sent me on slack, so let's go through those today

src/components/Collection.tsx
hokei: Yeah, I think it would be nicer (to look at) to return the loading state rather than checking if it's been loaded on every assignment

src/components/ListFilters.tsx
hokei: ok I am putting filters in pretty soon so I was getting thrown by the fact that they were in the code but not showing up -- this is good to know!  Thanks!

hokei:
Oh is the Vercel integration new as of this PR and what is it supposed to do?

hokei:
Navigating to the login modal the first time is showing an error, even though I haven't tried logging in yet
Screen Shot 2020-10-05 at 12 30 23 PM

(the icon on the right of the input is my password manager, so ignore that)

hokei:
in the search results page, does the loading indicator stop showing up if you've seen all the books

src/dataflow/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { ParsedUrlQuery } from "querystring";
+
+/**
+ * Extracts a string parameter from a url query
+ */
+export default function extractParam(
```
hokei, line 6: Any reason why it has to be this way rather than an approach that takes advantage of https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams?  (Feel free to punt if this requires refactoring)

hokei:
there's a lot going on here and I'm not confident I fully understand all of it, but most of it looks reasonable to me -- I'm mostly concerned about the functional issues, and then I think it's more important to merge and move forward and fix bugs when we see them.

src/dataflow/utils.ts
kristojorg: So I may have written a misleading comment. This is to extract a parameter from what next.js gives us. These aren't search parameters so much as dynamic routes (the `library` and `bookUrl` in `/[library]/book/[bookUrl]`). The type that Next.js gives us for those is `string | string[] | undefined`, because they could technically somehow be an array of strings. So this function just takes the query returned by `useRouter` and extracts a specific value and narrows the type to `string | undefined`. TBH I think next.js is misusing `query` here, because they aren't search queries, they're url parameters. But alas

kristojorg:
> Oh is the Vercel integration new as of this PR and what is it supposed to do?

This was an experiment, but we've removed it.

> Navigating to the login modal the first time is showing an error, even though I haven't tried logging in yet

This is a tricky one. It has to do with the fact that I don't clear the credentials when login fails. I couldn't really figure out a solution before but I'll give it a fresh look today.

> in the search results page, does the loading indicator stop showing up if you've seen all the books (It currently seems like it will spin forever)

Good question... I looked and you were right it was a problem I tried to fix, and uncovered another problem with how I was detecting the end of the search results on scroll. I tried to fix it for a while and resorted to removing the auto fetch more functionality in favor of a "View More" button. I asked on slack if that's okay. If not, we should revisit this in a separate ticket I think.

I also merged in the tests so now we just have this one PR to juggle.

hokei:
> I tried to fix it for a while and resorted to removing the auto fetch more functionality in favor of a "View More" button. I asked on slack if that's okay. If not, we should revisit this in a separate ticket I think.

Even tho slack says it's ok, I think let's move this to a new ticket and have the perpetual loading icon just be a known bug for now.

kristojorg:
> Even tho slack says it's ok, I think let's move this to a new ticket and have the perpetual loading icon just be a known bug for now.

I already went ahead and moved to the "View more" button, because the issues were larger than just the loading more icon. It was perpetually fetching actually, which wasn't good. I think the way it works now should be okay

src/components/WebpubViewer.tsx
```diff
@@ -1,71 +1,68 @@
...
-
-    const decryptorParams = await loadDecryptor(fetcher, entryUrl);
-    return await reader(entryUrl, catalogName, decryptorParams);
-  }
+  const loadDecryptorParams = async (webpubManifestUrl: any) => {
+    if (!NEXT_PUBLIC_AXIS_NOW_DECRYPT) return;
+    const response = await fetchWithHeaders(webpubManifestUrl, token);
+    const data = await response.json();
+    // there should never be a status code in the json
+    if (!response.ok || data.status) {
```
hokei, line 22: I'm getting this error when the response is 
{"status": 401, "type": "http://librarysimplified.org/terms/problem/credentials-invalid", "detail": "A valid library card barcode number and PIN are required.", "title": "Invalid credentials"}

which sends this page
Screen Shot 2020-10-06 at 8 57 56 AM

When it would be great to show a nice message to the user

kristojorg:
Okay I added a loading state to the login modal so that when you go to the reader or loans page directly it will show you "Logging in..." until the loans are fetched. We can technically not show this at all and just wait until we get a 401 response from the server, but let's make a separate ticket for that

src/components/WebpubViewer.tsx
kristojorg: This _should_ be caught by an error boundary in production and show a message to the user. Let me test that though. How did you get to this state?

hokei: Was not able to reliably reproduce ... I feel ok punting it until it shows up again.

hokei:
FYI while tabbing, the infinite list still shows.  New ticket.

hokei:
one last thing:  Let's test the fetch more button just to make sure its behavior is documented/won't regress without us knowing.

hokei:
but otherwise, it looks good to me.

kristojorg:
> FYI while tabbing, the infinite list still shows. New ticket.

What do you mean? 

> one last thing: Let's test the fetch more button just to make sure its behavior is documented/won't regress without us knowing.

Will do

hokei:
> What do you mean?

If you tab down the list, you never get to the "see more" button.  Instead, it will continue to load the list as long as you keep tabbing.  This means that the experience for a keyboard user is not the same as the experience for a mouse user.

kristojorg:
Okay I added a test for the "View more" functionality

hokei:
teach me how to write tests this quickly

hokei (approved):
There are a bunch of `firstPageUrl="/fist-page"` typos but let's get this in!

kristojorg:
> If you tab down the list, you never get to the "see more" button. Instead, it will continue to load the list as long as you keep tabbing. This means that the experience for a keyboard user is not the same as the experience for a mouse user.

That is weird... it seems to work on my localhost. I can tab down the list and get to the "View more" button, then press enter and it loads more. Anyway, yea lets open a ticket for what you're seeing.

> teach me how to write tests this quickly

haha copy/paste mostly

> There are a bunch of firstPageUrl="/fist-page" typos but let's get this in!

urgh will fix before merging!

hokei:
> That is weird... it seems to work on my localhost. I can tab down the list and get to the "View more" button, then press enter and it loads more. Anyway, yea lets open a ticket for what you're seeing.

 I will verify this, my localhost might have been in a bad state.  Yay!

PR #153, opened 2020-09-30
#10: Test coverage

kristojorg:
This adds tests to bring us back up to 90% coverage. There are still areas we should add tests for, but this brings us to at least as high of coverage as we were at before. Notable, I brought over tests for the parsing logic from opds-web-client.

Below find a snapshot of our test coverage with a list of files sorted by most uncovered lines:

**Coverage on beta (84%)**
image

**Coverage on this branch (89%)**
image

(they're different reports because my vscode plugin stopped working for some reason)


src/components/__tests__/BookList.test.tsx
```diff
@@ -0,0 +1,135 @@
+import * as React from "react";
+import { render, fixtures } from "test-utils";
+import { BookList, InfiniteBookList } from "../BookList";
+import merge from "deepmerge";
+import { BookData, CollectionData } from "interfaces";
+import { useSWRInfinite } from "swr";
+import { fetchCollection } from "dataflow/opds1/fetch";
+
+const books = fixtures.makeBooks(3);
+
```
kristojorg, line 10: Most of this file came from `ListView.test.tsx`... it was just renamed. I'm not sure why github didn't pick that up. I did add the tests for infinite book list below

hokei:
do you want to put `reader.ts` and `WebpubViewer.tsx` tests as part of this?  (Personally - they're going to mostly be sanity tests, I think, and it'll be much easier for you to set up the mocking than for me because I ... don't know ... jest ... /hides).

kristojorg:
@hokei yep I'll give it a shot : )

kristojorg:
Okay I just added some basic tests for the webpub viewer... mocked out `utils/reader` but tested most other functionality and also updated the code

src/components/WebpubViewer.tsx
```diff
@@ -5,69 +5,64 @@ import { useRouter } from "next/router";
...
-        token
-      );
+    if (token && bookUrl) initializeReader(bookUrl, catalogName, token);
   }, [token, bookUrl, catalogName]);
 
-  if (!token) return <div>You need to be logged in to view this page.</div>;
+  /**
+   * Show the auth modal if the user is not logged in
+   */
+  React.useEffect(() => {
```
hokei, line 48: yaaaaaay this is a much more elegant way of fixing this than I was thinking.

src/components/__tests__/BookList.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    ).toBeInTheDocument();
+  });
+
+  test("shows loading indicator when fetching more", () => {
+    useSWRInfiniteMock.mockReturnValue({
+      size: 1,
+      setSize: jest.fn(),
+      data: []
+    } as any);
+    const utils = render(<InfiniteBookList firstPageUrl="/fist-page" />);
```
hokei, line 131: typo

src/components/__tests__/BookList.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    expect(utils.getByText("Loading ...")).toBeInTheDocument();
```
hokei, line 133: When does the loading indicator stop showing up?  does the loading indicator show up when data is empty?

src/components/__tests__/WebpubViewer.test.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from "react";
+import { render, setEnv } from "../../test-utils";
+import WebpubViewer from "components/WebpubViewer";
+import mockAuthenticatedOnce from "test-utils/mockAuthState";
+import { PageNotFoundError } from "errors";
+
```
hokei, line 6: 🙇

src/components/__tests__/BookList.test.tsx
hokei: It'd be nice to see a "wait for element to be removed" -style test here.

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -63,13 +68,26 @@ test("extracts clever tokens from the url", () => {
   useRouterSpy.mockReturnValue({ replace: mockReplace, query: {} } as any);
   renderUserContext();
 
+  expect(mockSWR).toHaveBeenCalledWith(
+    null,
+    expect.anything(),
```
hokei, line 73: what is the purpose of these guys?

hokei:
Let's merge this into `collection-and-books` so we only have to juggle one branch.

hokei:
Are the tests under `/opds/` copied over?  I don't understand the opds logic super well so it generally looks good but I'm not gonna worry about it if they're copied over

src/components/WebpubViewer.tsx
hokei: However it looks like the user isn't immediately recognized as having been logged in, so the login modal appears for about a second, and then it reloads. This only happens when you refresh the reader page or when you navigate to a reader page directly.  It doesn't happen when you click "read online" from the catalog.

kristojorg:
Okay I'm going to merge this in and then address:

- A test for not showing the loading indicator when you are in fact not loading anymore (you're right this is broken on search results)
- A better way to detect if a user is logged in or not so you don't get the flash of modal

I took as many of the tests from opds-web-client as I could, but most weren't applicable. I did copy over tests for the parsing however, and a few other areas

src/components/context/__tests__/UserContext.test.tsx
kristojorg: This is because I want to expect that the first parameter is null, but I don't care about the rest of the parameters (for this test). If I omit them, though, the expect will fail, so I use this to allow anything to be passed in to the other params.

src/components/__tests__/WebpubViewer.test.tsx
kristojorg: : ) not wonderful but better than nothing

src/components/__tests__/BookList.test.tsx
kristojorg: Yep will add this

src/components/__tests__/BookList.test.tsx
kristojorg: Thank uuu

src/components/WebpubViewer.tsx
kristojorg: Will (try) to fix this!

PR #155, opened 2020-10-09
Config

kristojorg:
This PR does the following

1. App config is now set a YAML file that is loaded from the filesystem or network at build time using webpacks [val-loader](https://webpack.js.org/loaders/val-loader/). This includes the app's `libraries` (see below).
1. App now takes a dictionary to configure different media support (see the [sample config](https://github.com/NYPL-Simplified/circulation-patron-web/blob/config/sample-config.yml).)
1. The book details page now has states for "unsupported", "redirect-and-show", "redirect" and "show" states.
1. Books that have no supported media types are not filtered out when parsing a feed. We currently just show a "Unsupported" state on the book details page.
1. Books are now parsed into more specific types depending on their status: `fulfillable`, `borrowable`, `on-hold`, `reserved`, `reservable` and `unsupported`.
1. We used to choose the first borrow link present on the book as the canonical borrow link, which defined availability etc. Now, we choose the first borrow link which leads to a supported format. This means we don't need to support multiple borrow links for the time being.
1. Updates the book details page design to accommodate new states and designs.
1. Updates the book list view to accommodate new states and designs.
1. This PR removes the concept of "open access" from the app. Open access acquisition options are presented after borrowing (this is the same as before). The only difference is that you don't get a "This book is available to keep forever" message, because that was technically incorrect. This is a desired update.
1. Updates link handling to only allow multi-library routes, consistent with our new handling of library configuration.

## Configuring Libraries

We used to set which libraries the app uses with one of three environment variables (`SIMPLIFIED_CATALOG_BASE`, `LIBRARY_REGISTRY` and `CONFIG_FILE`). There have been some changes here. First, we now ALWAYS set the `CONFIG_FILE` variable, which now refers to the YAML config file holding all the information the app needs to startup properly. There are now only two ways to specify the libraries an app can startup with:

Set a dictionary in the config file:
```yml
libraries:
  lib1: http://circ-manager.com/lib1/groups
  lib2: http://another-cm-backend.com/groups
```

Or set a single URL to a library registry:
```yml
libraries: http://library-registry.com/
```

You will notice there is _no longer any way to set just a single library_. This is intended and has been discussed previously with @leonardr . To use a single library, simply put one library in the `libraries:` key:
```yml
libraries:
  o: http://my-one-backend.com
```

This `o` here will be the "library short name" in the browser, so the home catalog for that library will live at `http://catalog.library.com/o/`, and a book will live at `http://catalog.library.com/o/book/:bookUrl`. This is intended, at least for the time being, as having a single supported URL structure both reduces the complexity of this app, and the complexity of the circulation manager. It also makes it less likely there will be broken urls in the future.

That `o` can be anything you want: `app` or `library` or whatever. The application can also be deployed behind a domain alias to map `http://catalog.library.com/[...anything]` to `http://app-server.com/o/[...anything]` if desired. 

Note that this PR doesn't actually remove all the code to support single library routes yet. This PR removes the ability to configure single library instances, and the link within the app for single library instances. More code will be removed in a follow up PR.

Considerations:

1. The app needs to progress from this place to properly handle multiple borrow links. This means we need to be able to handle the case that a book has different availabilities for different borrow links. We need UX work to do this. At the same time we would need to handle open access content better.

Follow up PRs:

1. I will make a follow on PR to update the list view to show a read/download button if only one fulfillment type is supported for that book.
1. We will need a new PR to remove the remaining single library code.


src/components/BookList.tsx
```diff
@@ -169,182 +165,114 @@ export const BookListItem: React.FC<{
...
-        <>
-          {getCtaButtons(false)}
+      <>
+        <BorrowOrReserve url={book.borrowUrl} isBorrow />
+        <Text
+          variant="text.body.italic"
+          sx={{ fontSize: "-1", color: "ui.gray.dark", my: 1 }}
+        >
+          {availabilityString(book)}
+        </Text>
```
kristojorg, line 178: We no longer have the option to render multiple borrow links. We are just choosing one when parsing the opds entry, but choosing the first one that leads to a supported format according to the config file

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -3,92 +3,46 @@ import { render, fixtures, waitFor } from "test-utils";
...
+  ReservedBook
+} from "interfaces";
 
 function expectViewDetails(utils: ReturnType<typeof render>) {
   const button = utils.getByRole("link", { name: "View Book Details" });
   expect(button).toBeInTheDocument();
   expect(button).toHaveAttribute("href", "/book/http%3A%2F%2Ftest-book-url");
 }
 
-describe("open access book", () => {
```
kristojorg, line 13: This no longer is a state we fall in to

src/hooks/useDownloadButton.ts
```diff
@@ -3,20 +3,23 @@ import { DownloadDetails } from "utils/fulfill";
...
   const [error, setError] = React.useState<string | null>(null);
+  const { catalogUrl } = useLibraryContext();
   const { token } = useUser();
 
   async function download() {
     setLoading(true);
     setError(null);
     try {
-      await downloadFile(details.url, title, details.mediaType, token);
+      const url = await details.getUrl(catalogUrl, token);
```
kristojorg, line 21: This handles the possible indirection for the url. The getUrl function is defined in the `FulfillmentDetails`

src/icons/download.svg
```diff
@@ -1,4 +1,4 @@
-<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<circle cx="12" cy="12" r="11.6471" stroke="black" stroke-width="0.705882"/>
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+<circle cx="12" cy="12" r="11.6471" fill="none" stroke="currentColor" stroke-width="0.705882"/>
```
kristojorg, line 2: There was an error in this I noticed and just decided to quickly fix...

src/interfaces.ts
```diff
@@ -103,52 +127,89 @@ export type BookAvailability =
...
-  publisher?: string;
-  published?: string;
-  categories?: string[];
-  language?: string;
-  relatedUrl: string | null;
-  trackOpenBookUrl: string | null;
-  raw?: any;
-}
+
+export type Book<Status = EmptyObject> = Readonly<
```
kristojorg, line 131: I made this type more specific in a few ways:
1. It how has supertypes `FulfillableBook`, `ReservableBook` etc, which are the primary way we use this type. They're books specifying the specific state they are in.
1. Each supertype defines the data specific to that state instead of optionally including it all the time. Things like "borrowUrl" don't make sense for a book that is ready for reading. 
1. It is now readonly, so you can't accidentally mutate it in the app (there was a case where this happened previously).

src/interfaces.ts
```diff
@@ -103,52 +127,89 @@ export type BookAvailability =
...
+  status: "fulfillable";
+  fulfillmentLinks: FulfillmentLink[];
+  revokeUrl: string | null;
+}>;
+
+export type UnsupportedBook = Book<{
+  status: "unsupported";
+}>;
+
+export type AnyBook =
```
kristojorg, line 198: This is the new way to represent a book in any state

src/test-utils/fixtures/raw-opds-feed.ts
```diff
@@ -40,6 +40,7 @@ This work is available for countries where copyright is Life+70 and in the USA.
 		<link type="image/jpeg" rel="http://opds-spec.org/image" href="../larger.jpg"/>
 		<link type="image/jpeg" rel="http://opds-spec.org/image/thumbnail" href="../thumbnail.jpg"/>
 		<link type="text/html" rel="related" href="http://www.feedbooks.com/book/52" title="Some link for the test"/>
+		<link type="application/atom+xml;type=entry;profile=opds-catalog" title="Full entry" rel="alternate" href="entry.xml"/>
```
kristojorg, line 43: We now assume that books have links to their book details page always (which I believe is true). This fixture had to be updated to include one.

src/test-utils/mockConfig.ts
```diff
@@ -0,0 +1,48 @@
+import * as configModule from "config";
+import { AppConfig } from "interfaces";
+
+export default function mockConfig(custom?: Partial<AppConfig>) {
```
kristojorg, line 4: This lets you mock the config file in tests

src/types/opds1.ts
```diff
@@ -55,11 +59,15 @@ export const OPDSEntryMediaType =
...
 export const IncorrectAdeptMediaType = "vnd.adobe/adept+xml";
+// this is not yet supported, but the description is here:
+// https://github.com/NYPL-Simplified/Simplified/wiki/OPDSForDistributors#bearer-token-propagation
+export const BearerTokenMediaType =
+  "application/vnd.librarysimplified.bearer-token+json";
 
 export type IndirectAcquisitionType =
   | typeof OPDSEntryMediaType
   | typeof AdeptMediaType
-  | typeof IncorrectAdeptMediaType;
```
kristojorg, line 62: We now fix this at parse time, so it's no longer propagated through the app

src/types/opds1.ts
```diff
@@ -90,11 +100,17 @@ export type DownloadMediaType =
   | typeof KepubMediaType
   | typeof PdfMediaType
   | typeof MobiPocketMediaType
-  | typeof Mobi8Mediatype
+  | typeof Mobi8Mediatype;
+
+export type UnsupportedMediaType =
```
kristojorg, line 105: We have no UX we can render for these books right now, even if the config file says "show"

src/utils/__tests__/book.test.ts
```diff
@@ -46,171 +48,28 @@ describe("get authors", () => {
   });
 });
 
-describe("getFulfillmentState", () => {
```
kristojorg, line 49: This is now included in the book, processed during parsing. We no longer have to do it here.

src/utils/file.ts
```diff
@@ -4,10 +4,6 @@ export const typeMap: Record<
   OPDS1.DownloadMediaType,
   { extension: string; name: string }
 > = {
-  "application/vnd.overdrive.circulation.api+json;profile=audiobook": {
-    extension: "",
-    name: "Overdrive Audiobook"
-  },
```
kristojorg, line 10: These removed types never make it here anymore as our type system has gotten more specific

README.md
```diff
@@ -9,98 +9,81 @@
...
+The application will start at the base URL of `localhost:3000`.
 
 ### Running with Decryption
 
-Circulation-Patron-Web supports read online for encrypted books only in the AxisNow format, and if you have access to the [Decryptor](https://github.com/NYPL-Simplified/axisnow-access-control-web)
+This app supports read online for encrypted books only in the AxisNow format, and if you have access to the [Decryptor](https://github.com/NYPL-Simplified/axisnow-access-control-web)
 
 To run with decryption:  
-- Set `NEXT_PUBLIC_AXIS_NOW_DECRYPT=true`
+- Set `axisnow_decrupt=true` in your app's `config.yml` file.
```
hokei, line 78: nit: spelling

README.md
```diff
@@ -9,98 +9,81 @@
...
 - Run `git submodule update`
 - Verify that `axisnow-access-control-web` folder exists.  
 - Run `npm install` as normal.
 
+### ENV Vars and Building
+
+When building for production using `npm run build`, the env vars are set at build time. This means whatever you have in your `.env` or `.env.local` or set in the command line when running `npm run build` will be taken as the env for the app when you run it. Overriding env vars like this `CONFIG_FILE=config.txt npm run start` will not work, you have to set them at build time.
```
hokei, line 85: I know this is for production but I just want to clarify an assumption that I had in case it's not accurate: 
In the dev instance, my mental model was that if `.env.local` existed it would get picked up instead of `.env` ... is that not correct?

next.config.js
```diff
@@ -42,6 +42,12 @@ const config = {
       config.plugins.push(new BugsnagSourceMapUploaderPlugin(config));
     }
 
+    // source the app config file and provide it using val-loader
+    config.module.rules.push({
```
hokei, line 46: why wasn't something like this needed earlier for `cm_libraries.txt` ?

hokei:
have you tried building/running this from the docker container, just in case?

src/components/BookList.tsx
hokei: is it worth going back into opds-web-client and removing the multiple borrow button functionality? (read: filing a ticket for next sprint to do that work)

src/dataflow/opds1/parse.ts
```diff
@@ -221,6 +242,88 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): BookData {
     trackOpenBookUrl: trackOpenBookLink?.href ?? null,
     raw: entry.unparsed
   };
+
+  /**
+   * It's fulfillable if there are fulfillment links,
```
hokei, line 247: can this comment be reworded (or take it out completely, english is less precise than logic, it seems)

hokei: Actually this comment can just be `// It's a fulfillable book`
matching the below comments

kristojorg:
I personally think we need to wait until we get more design guidance from Marcus on how to render the companion app callout and the various states:

image

This one in particular does not look good, and I think needs some though. In general, how to render the companion app redirect is a question mark in the new designs

README.md
kristojorg: Yea that's true in dev and production, the `.env.local` overrides the `.env`, so we most use `.env` to set defaults / as an example.

kristojorg: I believe that `.env` and `.env.local` are merged however, so it's not one or the other. The `.env.local` variables overwrite the ones from `.env`

next.config.js
kristojorg: Good question - it was! The app just had never done it. Instead, ever since I got started, it had loaded this file from filesystem or network on every render, which of course is a performance hit especially in the network case.

src/components/BookList.tsx
kristojorg: I don't personally think so, it wasn't a lot of logic and in my mind opds-web-client is soon to be deprecated in favor of some flavor of this repository's code

sample-config.yml
```diff
@@ -0,0 +1,67 @@
+# APP MEDIA SUPPORT
+# Each entry has an array with "show", "redirect" or "show-and-redirect".
+# 
+# "redirect" tells the app to display a prompt directing the 
+# user to read the book in the companion mobile app (usually SimplyE).
+# 
+# The default for unlisted entries is unsupported. The format will be 
+# completely hidden. Any OPDS Entry with only this one unsupported format
+# will not be visible in the app.
+media_support:
```
EdwinGuzman, line 10: For this, will it matter if a library doesn't support a media type but the app itself does support it because another library supports it?

kristojorg: So those two concepts are now divorced. The app itself will continue to try and support as many media types as possible. The config file will configure which ones are shown on a per-instance basis. Does that answer your question? The only thing is that if the config file tells the app to show something that is internally unsupported, it won't be shown despite the config file setting (because we don't know how).

src/components/__tests__/BookListItem.test.tsx
EdwinGuzman: So how do older open access books get displayed?

src/test-utils/mockConfig.ts
```diff
@@ -0,0 +1,48 @@
+import * as configModule from "config";
```
EdwinGuzman, line 1: Where is the `"config"` file coming from?

EdwinGuzman:
This looks fine to me and it seems like there are still things to do but will be covered later. I don't think you need my approval but I'd wait for Crystal to be good about this update.

src/components/__tests__/BookListItem.test.tsx
kristojorg: They're now considered basically closed access. You have to borrow them before you can get the content, and this was intended (they are considering changing the opds feed to match this functionality too). This is so that the book actually ends up on your bookshelf when you're reading it.

src/test-utils/mockConfig.ts
kristojorg: The config file is generally loaded by webpack at build time, but I want to mock it out in tests. If you're asking about the `from "config"`, the project just has absolute imports turned on for the `src` directory, so this just imports the `src/config/index.ts` module.

kristojorg:
@hokei I have confirmed it works via docker : )

kristojorg:
@hokei I'll look into the wrong borrow button on search results.

The read/download action directly from search results will be in a follow up ticket

kristojorg:
@hokei okay I fixed that issue you were seeing on search results... I needed to use the loaned version of the book in the list if it existed, not the version I got back from the search results.

hokei (approved):

PR #156, opened 2020-10-13
Return and Revoke

kristojorg:
This PR adds the ability to return books and cancel reservations. I also add code to track any errors in the process to Bugsnag.


src/components/CancelOrReturn.tsx
```diff
@@ -0,0 +1,66 @@
...
+  id: string;
+}> = ({ text, loadingText, url, id }) => {
+  const { token, setBook } = useUser();
+  const { catalogUrl } = useLibraryContext();
+  const [loading, setLoading] = React.useState(false);
+  const [error, setError] = React.useState<string | null>(null);
+
+  async function cancelReservation(revokeUrl: string) {
+    if (!token) {
+      setError("You muse be signed in.");
```
kristojorg, line 24: We shouldn't get here, but there is no way to make it impossible from a type perspective right now

hokei: spelling nit

src/components/CancelOrReturn.tsx
```diff
@@ -0,0 +1,67 @@
...
+      return;
+    }
+    setLoading(true);
+    try {
+      const newBook = await fetchBook(revokeUrl, catalogUrl, token);
+      setBook(newBook, id);
+      setLoading(false);
+      // if it works, we need to refetch our loans
```
hokei, line 33: ? what does this mean?  Where does the refetching happen?

hokei:
So I notice return functionality isn't in the NYPL mocks, is this mostly to help us with our testing?  Should we show it by default on pages?  (If it's driven by libraries/etc I'm surprised it doesn't show up at least on the "my loans" page, that's where I'd expect to be able to return books)

src/components/CancelOrReturn.tsx
kristojorg: How r u so good at spelling

src/components/CancelOrReturn.tsx
kristojorg: Good catch, the `setBook` above does that work for us, this comment is outdated. `setBook` optimistically sets the returned book in our loans and then refetches them afterwards just to make sure we are properly sync'd up

kristojorg:
Yea this was requested by Lyrasis and also it was in some mocks from Marcus at some point I thought... idk. I just guessed. But yea ultimately it would be good to have this on the list view in the loans page once we redesign the list page

kristojorg:
I think we should show it though, it's necessary functionality for Open eBooks, I think they just never realized it wasn't already there tbh.

kristojorg:
Okay @hokei this should be ready when you are

src/components/CancelOrReturn.tsx
```diff
@@ -0,0 +1,66 @@
...
+        setError(e.info.detail);
+        return;
+      }
+      setError("An unknown error occurred");
+    }
+  }
+  if (!url) return null;
+  return (
+    <>
+      <Button
```
hokei, line 46: is it difficult to start moving these over to use DS buttons?  Just because then we don't need to worry about keeping them consistent re: font weight, etc.  (Haven't looked at the list refactor branch, but assume that's not there)

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -377,9 +478,7 @@ describe("FulfillableBook", () => {
     const utils = render(<FulfillmentCard book={bookWithRedirect} />);
 
     expect(utils.queryByText("Ready to Read!")).not.toBeInTheDocument();
-    expect(
-      utils.getByText("You're ready to read this book in SimplyE!")
-    ).toBeInTheDocument();
+    expect(utils.getByText("Ready to read in SimplyE!")).toBeInTheDocument();
     expect(utils.getByText("You have this book on loan.")).toBeInTheDocument();
```
hokei, line 482: Shouldn't this also expect the return button?

hokei: If Open Ebooks should explicitly not have a return button, let's test for that.

src/dataflow/opds1/parse.ts
```diff
@@ -113,12 +117,31 @@ function buildFulfillmentLink(feedUrl: string) {
...
   };
 }
 
+/**
+ * A book cannot be returned (even with a revoke url) if:
+ *  - It is locked in to Adobe ACS
+ *  - It is an axisnow book
+ *  - It comes from certain vendors
+ */
+function canReturnFulfillableBook(links: OPDSAcquisitionLink[]): boolean {
```
hokei, line 135: haha guessing this doesn't have at test

src/types/opds1.ts
```diff
@@ -56,17 +56,17 @@ export const OPDSEntryMediaType =
  * decrypted by us, so just download them and let the user open it in an app
  * that can deal with it.
  */
-export const AdeptMediaType = "application/vnd.adobe.adept+xml";
+export const AdobeDrmMediaType = "application/vnd.adobe.adept+xml";
 // there was an issue with incorrect Adept media types being sent.
-export const IncorrectAdeptMediaType = "vnd.adobe/adept+xml";
+export const IncorrectAdobeDrmMediaType = "vnd.adobe/adept+xml";
```
hokei, line 61: wait am I understanding correctly that when something happens with the Adobe DRM, it will send back a resource with this media type?

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -198,11 +221,62 @@ describe("reserved", () => {
...
+    expect(mockFetchBook).toHaveBeenCalledWith(
+      "/revoke",
+      "http://test-cm.com/catalogUrl",
+      "user-token"
+    );
+
+    expect(mockSetBook).toHaveBeenCalledWith(unreservedBook, reservedBook.id);
+  });
+
+  test("handles cancel reservation errors", async () => {
```
hokei, line 251: in general it'd be nice to figure out exactly what errored from the name of the test ... is this test checking that an error while cancelling reservations will return the document to the un-cancelled state?

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -266,6 +340,33 @@ describe("FulfillableBook", () => {
     }
   });
 
+  test("displays return button that calls apropriate url", async () => {
```
hokei, line 343: does there need to be something that checks when the return state errors?

kristojorg: Good point, I can test for that also for that

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
kristojorg: Ah yes, I was testing this in the parser.test.tsx, but I'll also add it here that there should _not_ be a return button here

src/dataflow/opds1/parse.ts
kristojorg: Ok you caught me... it does now though : )

src/components/CancelOrReturn.tsx
hokei: sounds good.  In the screenshots I was showing Marcus it looked like some of the text wasn't bolded, etc.  so that's why I brought it up.  But I'm happy to do it at some other point or whenever it makes life easier.

hokei (changes requested):
Returns always seem to bounce the first time, but work the second time.  (Sent a video in slack)

kristojorg:
@hokei I believe this is only happening with reservations through overdrive, and it's a server issue. If you don't mind, could you test this with another book for now?

kristojorg:
I just debugged and we're opening a ticket to fix this on the server

kristojorg:
@hokei I'd love to get this in if possible! It's currently blocking us from merging the list design PR also. There is a ticket open to fix the bug you discovered in the CM

hokei (approved):

PR #157, opened 2020-10-13
Single library routes

kristojorg:
Removes unneeded code now that we don't support single library routes anymore


src/hooks/useLinkUtils.ts
```diff
@@ -0,0 +1,34 @@
+import useLibraryContext from "components/context/LibraryContext";
+
+const trailingSlashRegex = /\/$/;
+
```
kristojorg, line 4: The logic in this file used to be in the `LinkUtilsContext`. It no longer needs to be provided on context, so I moved it to here in order to reduce the amount of context we are using

src/pages/index.tsx
```diff
@@ -1,17 +1,10 @@
...
+import ErrorComponent from "components/Error";
 
-const CollectionPage: NextPage<AppProps> = ({ library, error }) => {
-  return (
-    <LayoutPage library={library} error={error}>
-      <Collection title={`${library?.catalogName} Home`} />
-    </LayoutPage>
-  );
+const CollectionPage: NextPage<AppProps> = () => {
+  return <ErrorComponent />;
```
kristojorg, line 7: We can no longer view anything at `catalog.library.com/`, so we render the error page which shows links to the available libraries you can view

kristojorg:
Closing this as I've merged it into #158

PR #158, opened 2020-10-13
Performance

kristojorg:
This makes two simple changes that were badly needed, which became apparent when trying the app in production mode. 

1. When we fetch a collection, that collection contains the full book info for each book in the collection, which means we shouldn't really need to fetch the book in order to show you the book details page. This PR simply updates the SWR cache when we fetch a collection to add the books from that collection to the cache.
1. We now use Next.js's [getStaticProps](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) instead of `getServerSideProps` when fetching the data needed for startup on each page. This means that the page will fetch the library data the first time it is ever requested, and then will cache that data for future requests. I set it to revalidate this library data once per hour, since it changes infrequently. 
1. The first time a page is ever requested, it has to get the appropriate information from the server, which might take a second. In this time, we have no library information so we can't render the layout. Instead we render a simple loader. This shouldn't happen often because it is not per-user, but per instance. So the first time a path has _ever_ been loaded. It also used to happen in our previous implementation but the page would just hang until the data was ready, so now at least we show a loader in that time when it does happen

In the future we can fetch the book and collection data statically on the server like we do the library data.


src/pages/[library]/book/[bookUrl].tsx
```diff
@@ -12,6 +12,13 @@ const BookPage: NextPage<AppProps> = ({ library, error }) => {
   );
 };
 
-export const getServerSideProps: GetServerSideProps = withAppProps();
+export const getStaticProps: GetStaticProps = withAppProps();
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  return {
+    paths: [],
```
Comment (kristojorg) [inline src/pages/[library]/book/[bookUrl].tsx:19]: We have the option to statically generate paths at build time, but instead we are just choosing to do it the first time a path is requested, and then from then on out it'll just be static data served to the browser.

kristojorg:
This now includes the #157 work

EdwinGuzman (approved):

PR #160, opened 2020-10-14
List view action

kristojorg:
This adds an action on the list view pages to fulfill a book if there is only one fulfillment possibility, which would be the case in Open eBooks. So it shows a single "Read" button in that case.


src/components/FulfillmentButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+import useLibraryContext from "components/context/LibraryContext";
+import useUser from "components/context/UserContext";
+import downloadFile from "dataflow/download";
+import { ServerError } from "errors";
+
+function useError() {
+  const [error, setError] = React.useState<null | string>(null);
+
+  function handleError(e: any) {
+    // TODO: track error to bugsnag
```
kristojorg, line 27: I made the tracking function in another PR, will update once this is all merged

src/components/FulfillmentButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  return {
+    error,
+    handleError,
+    clearError
+  };
+}
+
+const FulfillmentButton: React.FC<{
```
kristojorg, line 51: This switch used to live in the `FulfillmentCard`, but now it's extracted to be used in the list view also.

hokei: love it.

hokei:
Screen Shot 2020-10-16 at 10 02 53 AM

I know a lot of this is getting rewritten as we speak, but I want to point this out /just in case/ it's a symptom of a bigger underlying issue.  If it's just CSS then I'm happy to ignore it since the list view is about to look super different

src/components/FulfillmentButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+      <Button
+        onClick={download}
+        {...getButtonStyles(isPrimaryAction)}
+        iconLeft={SvgDownload}
+        loading={loading}
+        loadingText="Downloading..."
+      >
+        {buttonLabel}
+      </Button>
+      {error && <Text sx={{ color: "ui.error" }}>{error}</Text>}
```
hokei, line 170: Screen Shot 2020-10-16 at 10 04 16 AM

hokei: More of a thing to check for in the new version, I don't think it's that important to fix here.

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -258,6 +258,38 @@ describe("available to access book", () => {
...
+    });
+    const utils = render(<BookListItem book={book} />);
+    expect(
+      utils.getByRole("button", { name: "Download PDF" })
+    ).toBeInTheDocument();
+  });
+
+  test("doesn't show FulfillmentButton if multiple options", () => {
+    const utils = render(<BookListItem book={downloadableBook} />);
+    expect(utils.queryByText("Download PDF")).not.toBeInTheDocument();
```
hokei, line 290: can we expect the link to book details?

src/components/__tests__/BorrowOrReserve.test.tsx
```diff
@@ -121,7 +121,7 @@ test("catches unrecognized fetch errors", async () => {
   // shows the error, button resets.
   await waitFor(() => {
     expect(
-      utils.getByText("Error: An error occurred while borrowing this book.")
+      utils.getByText("Error: An unknown error occurred.")
```
hokei, line 124: Why the change?

hokei (approved):
Just some CSS issues and nits, you probably want your list work to be on top of this, so I'll approve first.

kristojorg:
> I know a lot of this is getting rewritten as we speak, but I want to point this out /just in case/ it's a symptom of a bigger underlying issue. If it's just CSS then I'm happy to ignore it since the list view is about to look super different

Yep this is basically all rewritten in another branch : ) and shouldn't have this issue anymore

src/components/__tests__/BorrowOrReserve.test.tsx
kristojorg: I extracted the error logic into a hook that is used in multiple places and didn't think it worth it to customize the error message when the error is unknown for each place it's used. Most cases the error will be known (a `ServerError`), and when it's not, you kind of already know the error happened when trying to borrow the book bc you just clicked that and it's right next to the button

src/components/__tests__/BookListItem.test.tsx
kristojorg: We can, but it gets removed in my next PR for this view, so I'm just gonna leave it out for now if that's okay.

PR #161, opened 2020-10-14
Axisnow issue

kristojorg:
We need to move the variable that indicates access to the axisnow decryptor back to an environment variable, and also stub it for when it isn't accessible to avoid webpack errors on build

### DIFF
--- a/src/components/__tests__/WebpubViewer.test.tsx
+++ b/src/components/__tests__/WebpubViewer.test.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import { mockShowAuthModal, render } from "test-utils";
 import WebpubViewer from "components/WebpubViewer";
 import { PageNotFoundError } from "errors";
+import * as env from "utils/env";
 
 jest.mock("utils/reader", () => ({
   __esModule: true,
@@ -42,7 +43,7 @@ test("renders viewer div", () => {
 });
 
 test("fetches params with token if run with NEXT_PUBLIC_AXIS_NOW_DECRYPT", async () => {
-  process.env.NEXT_PUBLIC_AXISNOW_DECRYPT = "true";
+  (env as any).AXISNOW_DECRYPT = "true";
   render(<WebpubViewer />, {
     router: { query: { bookUrl: "http://some-book.com" } }
   });
--- a/src/components/__tests__/WebpubViewer.test.tsx
+++ b/src/components/__tests__/WebpubViewer.test.tsx
@@ -2,7 +2,6 @@ import * as React from "react";
 import { mockShowAuthModal, render } from "test-utils";
 import WebpubViewer from "components/WebpubViewer";
 import { PageNotFoundError } from "errors";
-import mockConfig from "test-utils/mockConfig";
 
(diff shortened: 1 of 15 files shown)


kristojorg:
You can click on the preview link now and see that this works! Yippee : )

hokei:
> You can click on the preview link now and see that this works! Yippee : )

Oh my god this is so nice

hokei:
so the AxisNow stuff can't be in `config.yml`?

hokei:
note to self: submodules are bad figure out what was even that wrong with the npm module

src/utils/reader.ts
```diff
@@ -15,7 +15,7 @@ import {
 } from "library-simplified-webpub-viewer";
 import fetchWithHeaders from "dataflow/fetch";
 import ApplicationError from "errors";
-import { APP_CONFIG } from "config";
+import Decryptor from "AxisNowDecryptor";
```
kristojorg, line 18: We can now simply import from here. If the env var wasn't set to make it available we will get the value `false` instead.

hokei: niiice

src/components/__tests__/WebpubViewer.test.tsx
```diff
@@ -43,7 +42,7 @@ test("renders viewer div", () => {
 });
 
 test("fetches params with token if run with NEXT_PUBLIC_AXIS_NOW_DECRYPT", async () => {
-  mockConfig({ axisNowDecrypt: true });
+  process.env.NEXT_PUBLIC_AXISNOW_DECRYPT = "true";
```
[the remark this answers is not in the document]
hokei: though with the webpack fallback I think ... it's gonna be less likely.

src/utils/reader.ts
hokei: There's a similar import in `WebpubViewer.tsx` that can be changed to be like this too!

hokei:
Looks like there's a failing test but otherwise it looks good to me!

src/components/__tests__/WebpubViewer.test.tsx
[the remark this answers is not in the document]
kristojorg: I don't totally know how right now, but we should open a ticket for those tests in the future... although that is also something that the integration tests Lyrasis is setting up should catch

src/utils/reader.ts
kristojorg: I removed it earlier : )

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -329,7 +331,7 @@ describe("FulfillableBook", () => {
   });
 
   test("internal read online button tracks open_book event", async () => {
-    process.env.NEXT_PUBLIC_AXISNOW_DECRYPT = "true";
+    (env as any).AXISNOW_DECRYPT = "true";
```
hokei, line 334: ah ew fun

hokei (approved):

PR #162, opened 2020-10-14
List filters

kristojorg:
This PR adds support for arbitrary filters configured by the CM in all collection views. 

The designs didn't account for many filters, so I had to guesstimate. Here is what it looks like on mobile:

image

And on desktop:

image


hokei:
can you post a screenshot of this on slack and tag Risa and Marcus?  I just want to make sure they're aware that multiple filters is a thing.

hokei (approved):

PR #168, opened 2020-10-14
See More Block

kristojorg:
This is for the see more block at the end of a lane


hokei:
the right button doesn't seem to scroll to the end of the extra block:  
Screen Shot 2020-10-16 at 8 19 54 AM

src/components/__tests__/Lane.test.tsx
```diff
@@ -46,11 +46,11 @@ test("renders breadcrumbs with 0 books", () => {
...
 
   // it should just show the breadcrumb
-  const breadcrumb = utils.getByText("my lane");
+  const breadcrumb = utils.getByLabelText("Lane Title");
   expect(breadcrumb).toBeInTheDocument();
 
-  // there should be no li elements
-  expect(utils.getByTestId("lane-list")).toBeEmptyDOMElement();
+  // there should be only one See More element
+  expect(utils.getByTestId("lane-list").childNodes).toHaveLength(1);
```
hokei, line 53: So the rest of the cards aren't li ?

src/components/bookDetails/__tests__/Recommendations.test.tsx
```diff
@@ -71,6 +71,34 @@ test("doesn't show recommendations if there are none", () => {
   expect(utils.container).toBeEmptyDOMElement();
 });
 
+test("doesn't show recommendations if the lanes don't have enough books", () => {
```
hokei, line 74: is "enough books" defined by a constant somewhere or is it hardcoded?  If hardcoded, it would be nice to for the test name to show the expected number of books is so that if it's fails the title will be more informative.

hokei (changes requested):
Mostly looks good, my main concern is with the screenshotted issue.

kristojorg:
> the right button doesn't seem to scroll to the end of the extra block:

Question: Had you recently resized the browser when you got to this state? I can't seem to get there without resizing the browser. It occurs when you do that because (I think) our lane scrolling calculations aren't being rerun on browser resize, which I think should be a separate ticket. Can you check if this works if you keep the same size browser from load to when you scroll? (Or maybe you did that any I'm just not able to reproduce)

src/components/__tests__/Lane.test.tsx
kristojorg: This is only when there are 0 books, it will still show the "See More" block

src/components/bookDetails/__tests__/Recommendations.test.tsx
kristojorg: Yep this is >1, I updated the test title : )

src/components/__tests__/Lane.test.tsx
hokei: why should it show the See More block when there are 0 books?  Shouldn't it just not show lanes with no books in them?

kristojorg: Yea I guess you're probably right... although for some reason I didn't add that before

kristojorg: I'll add it

hokei (approved):

PR #169, opened 2020-10-15
Axisnow package

kristojorg:
This imports the axisnow decryptor package as a private npm package. The mock is still provided for when running without `NEXT_PUBLIC_AXISNOW_DECRYPT`, and we also fail at build time now if you try to run the app with decryption but don't have access to the module.

To get this to work, the axisnow package needs to be published to NPM and we need to update the `package.json` here to include it.


src/config/fetch-config.js
```diff
@@ -29,6 +29,7 @@ async function fetchConfigFile(configFileUrl) {
     const parsed = parseConfigText(text);
     return parsed;
   } catch (e) {
+    console.error(e);
```
kristojorg, line 32: Intended

src/utils/reader.ts
```diff
@@ -82,18 +82,12 @@ async function initBookSettings(
...
   const scroller = new ScrollingBookView();
 
-  let decryptor: any = undefined;
-
-  try {
-    decryptor = Decryptor
-      ? await Decryptor.createDecryptor(decryptorParams)
-      : undefined;
-  } catch (e) {
+  if (!Decryptor) {
```
kristojorg, line 85: If we don't have access but someone directly opens a reader page, they will now see this error.

hokei:
we've wanted this for a while but it's gonna take me a bit to figure out where to get private npm publish permissions

kristojorg:
Drats... I don't know how to make npm i ignore a dependency if we don't have access to it. Annoying

hokei:
gah I guess we re-figured out why an npm module was tricky.  I'm sorry I couldn't remember it ...

kristojorg:
Haha no worries but ohhhh well!

kristojorg:
Gonna close this :(

PR #174, opened 2020-10-19
List design

kristojorg:
This PR updates the list views to the new design by Marcus. It includes the return/revoke work , so should be merged as a follow up to that PR. 

There were quite a few states that weren't accounted for in the designs, so I made some (un)educated guesses


hokei:
does it make sense to put it in something like DS's template (Not a component, just CSS classes that you should already have access to) so that it centers when it's big enough? https://nypl.github.io/nypl-design-system/storybook-static/?path=/story/template--template

hokei:
ok I'm gonna come back to this after the other two PRs are merged because it's really hard to look at without.

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -10,9 +10,10 @@ import {
   ReservableBook,
   ReservedBook
 } from "interfaces";
+import { mergeBook, mockSetBook } from "test-utils/fixtures";
 
-function expectViewDetails(utils: ReturnType<typeof render>) {
-  const button = utils.getByRole("link", { name: "View Book Details" });
+function expectReadMore(utils: ReturnType<typeof render>) {
+  const button = utils.getByRole("link", { name: "Read more" });
```
hokei, line 16: this is an `a` element so we should call it a link to avoid confusion

hokei:
Is there a reason why the copy was changed from "Unavailable" to "Unsupported" ? 
Also, is "ready to read!" from somewhere?  How much freedom do we have with copy?

src/components/__tests__/BookListItem.test.tsx
kristojorg: done

kristojorg:
> Is there a reason why the copy was changed from "Unavailable" to "Unsupported" ?

Where do you mean? Unavailable is supposed to mean that all copies of the book are out, so it can't be borrowed only reserved. 

> Also, is "ready to read!" from somewhere? How much freedom do we have with copy?

Haha nope. I didn't see this state in the designs so I interpolated. I think we have some freedom, but it has to also work with `Ready to read in {companionApp}`

src/components/BookList.tsx
```diff
@@ -117,198 +118,139 @@ export const BookListItem: React.FC<{
...
-            isPrimaryAction
-          />
-        )}
-        <NavButton variant="ghost" bookUrl={book.url} iconRight={ArrowForward}>
-          View Book Details
-        </NavButton>
-      </>
+      <NavButton variant="link" bookUrl={book.url} iconRight={ArrowForward}>
+        View Book Details
+      </NavButton>
```
kristojorg, line 249: We show a link to see book details if the book is fulfillable, but there is more than one way to fulfill it.

src/components/BookStatus.tsx
```diff
@@ -0,0 +1,65 @@
...
+      ? "Available to borrow"
+      : status === "reservable"
+      ? "Unavailable"
+      : status === "reserved"
+      ? "Reserved"
+      : status === "on-hold"
+      ? "Ready to Borrow"
+      : status === "fulfillable"
+      ? `Ready to Read${redirectUser ? ` in ${companionApp}` : ""}!`
+      : "Unsupported";
```
kristojorg, line 33: This is now where we set that bold text showing the status of the book. This is used in both the book details and the book list now

src/components/BookList.tsx
```diff
@@ -99,7 +100,7 @@ export const BookList: React.FC<{
   books: AnyBook[];
 }> = ({ books }) => {
   return (
-    <ul sx={{ px: 5 }} data-testid="listview-list">
+    <ul sx={{ px: [3, 5] }} data-testid="listview-list">
```
hokei, line 103: What does `px: [3, 5]` mean?

src/components/BookList.tsx
```diff
@@ -117,198 +118,139 @@ export const BookListItem: React.FC<{
...
 
+const Description: React.FC<{ book: AnyBook; className?: string }> = ({
+  book,
+  className
+}) => {
+  return (
+    <div className={className}>
+      <Text
+        variant="text.body.italic"
+        dangerouslySetInnerHTML={{
```
hokei, line 184: Why does it need to be done this way? Is there something wrong with just setting children as `truncateString(book.summary)` ?

src/components/BookList.tsx
```diff
@@ -117,198 +118,139 @@ export const BookListItem: React.FC<{
...
-          sx={{ fontSize: "-1", color: "ui.gray.dark", my: 1 }}
-        >
-          {availabilityString(book)}
-        </Text>
-        <NavButton variant="ghost" bookUrl={book.url} iconRight={ArrowForward}>
-          View Book Details
-        </NavButton>
-      </>
-    );
+    return <BorrowOrReserve url={book.reserveUrl} isBorrow={false} />;
```
hokei, line 205: For the future:  Do you think it makes sense to separate the borrow and reserve buttons (as well as the cancel and return buttons)?

hokei:
Screen Shot 2020-10-27 at 12 14 48 PM

What do you think about having a way to keep content centered on super wide screens? (New ticket I think)

hokei:
Differences between this and the mock:  

Number of copies should not be in Open Ebooks
The "Available/Ready to read" text should be in primary-color
A non-borrowed book should have both a book details and a borrow button
Title should not be bold

Screen Shot 2020-10-27 at 12 18 22 PM
Screen Shot 2020-10-27 at 12 18 37 PM

hokei:
Looking for string truncation tests, not finding them -- I know this has been around for a while, but 

1) let's not write our own striphtml regex - either put everything in a div and call `textContent` or ... like, do summaries often come with html?  Do we really need to do html parsing at all? 
2) tests for truncateString really help in making sure we know what's going on in edge cases.  Especially since we don't have a book fixture with a super long title, author and summary, let's make sure we have tests for this.  

This can go in another ticket.

src/components/BookStatus.tsx
hokei: Do we anticipate this being configurable in the future?

kristojorg: No I hope not!

src/components/BookList.tsx
kristojorg: It's a theme-ui thing. It means it is 3 until the first app-wide breakpoint, then 5

src/components/BookList.tsx
kristojorg: The string is actually a sanitized html string, so it has to be set this way. Otherwise it would render the html tags as part of the content

src/components/BookList.tsx
kristojorg: Yeaaaa maybe. I was thinking about that. They do share some logic and I think should have the same styling, but yea it might be good to do that in the future

src/components/BookList.tsx
hokei: thanks!

kristojorg:
> What do you think about having a way to keep content centered on super wide screens? (New ticket I think)

Yea down with that

> let's not write our own striphtml regex - either put everything in a div and call textContent or ... like, do summaries often come with html? Do we really need to do html parsing at all?

The summaries do often contain html, almost always in other libraries (I would be surprised if that's not the case in Open eBooks.  The striphtml regex has been carried over from older versions, but I'm down to install an npm package if you'd rather. Right now the html is sanitized by Dompurify when parsing the OPDS entry, but then when showing the summary it should just be a string so that's where I strip the HTML out (it is still in HTML format on the book details page)

> tests for truncateString really help in making sure we know what's going on in edge cases. Especially since we don't have a book fixture with a super long title, author and summary, let's make sure we have tests for this.

How would these tests work? I think you're right it would be great make sure this looks right, but I don't really know how to test this. Any suggestions?

src/components/BookList.tsx
kristojorg: Ah you're right actually, here we can just set it as text because it is the stripped html

hokei: But there shouldn't be html tags after it is stripped?

hokei:
> The summaries do often contain html, almost always in other libraries (I would be surprised if that's not the case in Open eBooks. The striphtml regex has been carried over from older versions, but I'm down to install an npm package if you'd rather. Right now the html is sanitized by Dompurify when parsing the OPDS entry, but then when showing the summary it should just be a string so that's where I strip the HTML out (it is still in HTML format on the book details page)

So the way I've always done it in the past is to put it in a div and call innerText() on the whole div which lets the DOM handle it.  I think, if we know that it doesn't have anything executable, we can go with this approach.

hokei:
> How would these tests work? I think you're right it would be great make sure this looks right, but I don't really know how to test this. Any suggestions?

It looks like truncation works the same in every situation, right?  I think something as simple as a `string.test.ts` that tests different inputs and outputs, like `expect truncateString("biglongword", 2, true) to return "biglongword"` and `expect truncateString("t i n y", 300, false) to return "t i n y"` and ... you get the idea?

hokei (approved):
Not sure if we want to make tickets for the two clean-up tasks now, but they should go somewhere so we don't forget about them ...

PR #175, opened 2020-10-20
List filters delay

kristojorg:
This PR fixes the performance problem we saw when switching between list filters. I'll try to explain the underlying problem...

- The app uses `getStaticProps` for every page to fetch the library data we need. This means that the first time a route is requested, `getStaticProps` is run server-side, and then the page is rendered using that data. For the time being, this only loads the information we need for the library, so basically just the layout (the logo, library name, colors, etc). The actual collection or book data is fetched client-side (for now).
- Once we run `getStaticProps` on the server, the HTML that generates is sent to the client, and then cached. So subsequent requests to that page will be immediately responded to with the HTML already generated and ready on the server. The app will revalidate this cached data according to a schedule we set up in `withAppProps`. 
- So, by default, every page requires the server data from `getStaticProps`, and navigation will be delayed until that data is returned.
- Next.js (for the time being) doesn't know that every page within one library route (`/library/...`) uses the same data, it's not unique to every page. So that means that when you navigate from `/library/collection/collectionUrl` to `/library/collection/collectionUrl2`, the app thinks it needs new server-side data and will wait for `getStaticProps` to return. If that page had already been rendered, it would be cached and therefore not very slow, but it still wouldn't be as fast as if we just reused the old library data and changed the url immediately. 
- Next.js provides a [shallow routing](https://nextjs.org/docs/routing/shallow-routing) option to tell the app to do just that: route to the new page, but don't refetch the server-side data... assume all the data can be fetched client side. 
- It turns out, since all page-specific data is currently fetched client-side, we can use shallow routing everywhere. So I just updated to do that.
- The one place we can't use shallow routing is when we are navigating from the 404 page, because different libraries do actually have different data returned from the server, so we would end up in an error state. 
- I turned on shallow routing everywhere, so we should see increased performance anytime we are navigating between two pages that share the same path (`/collection/:collectionUrl` -> `/collection/:collectionUrl2` or `/book/:bookUrl` -> `/book/:bookUrl2`)

Also, just in case you're wondering, I do plan eventually to turn on full server-side data fetching, so the book details and the collection are fetched server side as well in `getStaticProps`. We just haven't gotten around to it yet.

### DIFF
--- a/src/components/FulfillmentButton.tsx
+++ b/src/components/FulfillmentButton.tsx
@@ -123,7 +123,7 @@ const ReadOnlineInternal: React.FC<{
   const internalLink = buildMultiLibraryLink(details.url);
   function open() {
     track.openBook(trackOpenBookUrl);
-    router.push(internalLink);
+    router.push(internalLink, undefined, { shallow: true });
   }
   return (
     <Button {...getButtonStyles(isPrimaryAction)} onClick={open}>
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -62,7 +62,10 @@ const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
       prefetch,
       replace,
       scroll,
-      shallow,
+      // default to shallow routing because we don't fetch any
+      // page-specific data in getStaticProps. See more:
+      // https://nextjs.org/docs/routing/shallow-routing
+      shallow = true,
       ...rest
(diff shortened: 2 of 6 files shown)


hokei (approved):
I honestly am still a bit iffy on routing, but it works ok now and everything looks ok so let's get it in

kristojorg:
Yea I hear you, it's a bit confusing because the app is both server and client rendered... so kinda complicated sometimes

PR #176, opened 2020-10-20
Stable loans sort

kristojorg:
This simple fix corrects the way that the loans page would resort after a second when you click the page. If two books don't have availability defined, we sort by title, which makes the sort "stable", meaning that it will return the same output every time if it is given the same input.

### DIFF
--- a/src/components/MyBooks.tsx
+++ b/src/components/MyBooks.tsx
@@ -19,11 +19,22 @@ function sortBooksByLoanExpirationDate(books: AnyBook[]) {
   return books.sort((a, b) => {
     const aDate = availableUntil(a);
     const bDate = availableUntil(b);
+    // if there is no availability info for either, compare their titles
+    if (typeof aDate === "string" && typeof bDate === "string") {
+      return compareTitles(a, b);
+    }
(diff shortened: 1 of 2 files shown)


src/components/MyBooks.tsx
```diff
@@ -19,13 +19,24 @@ function sortBooksByLoanExpirationDate(books: AnyBook[]) {
...
     const aDate = availableUntil(a);
     const bDate = availableUntil(b);
+    // if there is no availability info for either, compare their titles
+    if (typeof aDate === "string" && typeof bDate === "string") {
+      return compareTitles(a, b);
+    }
+    // if only one has a defined availability, it goes on top
     if (typeof aDate === "string") return 1;
     if (typeof bDate === "string") return -1;
+    // if both have defined availabilities, sort by date
```
hokei, line 29: What if it's the same date? Will that still be stable?

hokei (changes requested):
see above comment

src/components/MyBooks.tsx
kristojorg: Good point, just added

hokei (approved):

PR #178, opened 2020-10-21
Storybook

kristojorg:
This adds storybook for coordinating design on some of our more complicated components. Initially, we start with the book details page and the book list item component. This is based on the lane-design branch. I also reference the NYPL-DS storybook so it can be browsed alongside our storybook for future use.


src/auth/CleverAuthButton.tsx
```diff
@@ -27,12 +26,15 @@ const CleverButton: React.FC<{ method: OPDS1.CleverAuthMethod }> = ({
...
-        backgroundImage: `url(${imageUrl})`
+        backgroundColor: "#2f67aa"
       }}
       aria-label="Log In with Clever"
     >
-      {!imageUrl ? "Log In With Clever" : ""}
+      {!imageUrl ? (
+        "Log In With Clever"
+      ) : (
+        <img alt="Clever Logo" src={imageUrl} />
```
hokei, line 36: yeah I like this better

src/components/BookList.tsx
```diff
@@ -4,29 +4,32 @@ import * as React from "react";
...
   bookIsReserved,
-  bookIsOnHold,
-  bookIsUnsupported
+  bookIsOnHold
 } from "../utils/book";
 import Lane from "./Lane";
 import Button, { NavButton } from "./Button";
 import LoadingIndicator from "./LoadingIndicator";
 import { H2, Text } from "./Text";
-import * as DS from "@nypl/design-system-react-components";
```
hokei, line 19: I'm not specifically against this, but I'm a bit confused why to include design system in the story book if it's no longer being used and I think it would be useful in the future to talk with Helen about why a project would move off of it.  (Context:  Number of projects using design system is one of the metrics that that team uses, and I think they didn't anticipate for any team that has been using to decide not to, and they should).

kristojorg: This was actually from the list-design PR, when we moved off of the DS.Card to a normal div

kristojorg: In general I support moving more and more towards the DS, I just would rather keep it as a project for after launch except for new components we add that can be imported from the DS

kristojorg:
FYI: I also update dompurify in this PR, because we used to use some weird hacks to make it work in node, while the current version works in node by default

src/components/stories/BookListItem/FulfillableBook/PrefersCompanionApp.stories.tsx
```diff
@@ -0,0 +1,118 @@
+import React from "react";
+import { Story, Meta } from "@storybook/react/types-6-0";
+import {
+  axisnowFulfillmentLink,
+  epubFulfillmentLink,
+  externalReadFulfillmentLink,
+  fulfillableBook,
+  pdfFulfillmentLink,
+  unsupportedBook
+} from "test-utils/fixtures/book";
```
hokei, line 10: really enjoying how these fixtures get reused, super useful

kristojorg: Yea I sort of wish I would have written them all months ago haha

.storybook/config-mock.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (parameters && parameters.config) {
+    APP_CONFIG = {
+      ...defaultMock,
+      ...parameters.config,
+    }
+  }
+  APP_CONFIG.companionApp = globals.companionApp;
+  APP_CONFIG.showMedium = globals.showMedium; 
+  return story();  
+}
```
hokei, line 18: is your code formatter not putting newlines at the end of files?

hokei (approved):
It looks good to me -- I haven't gone in to actually make sure all the stories are written ideally or anything, but as this doesn't affect app functionality, I'm just happy they're using the same fixtures as tests and we can add to it as we need different stuff.

.storybook/config-mock.ts
kristojorg: I've been having intermittent issues with eslint that I can't quite figure out, which might be the cause of this since it runs prettier

PR #181, opened 2020-10-22
Axisnow

kristojorg:
Finally figured it out. We just needed to install the private package as optional so that if it fails the build can still continue... duh. Didn't know that was an option!

This now includes the new private AxisNow Decryptor package! I also improved the error messaging if you try to access the reader but don't have the decryptor. And I made the error page only show the slugs when it has a 404 on "/".

The Open eBooks Preview link now includes reading functionality, so once this is merged, QA can start using it. : )

Note: We no longer need to set any env var to use the decryptor. If the install succeeds, it will be automatically detected and enabled throughout the app.


.github/workflows/ci.yml
```diff
@@ -18,18 +18,29 @@ jobs:
...
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      
+      - name: Install Dependencies
+        run: npm install
```
kristojorg, line 42: This is the now recommended way to install node modules via Github Actions with caching

package.json
```diff
@@ -57,6 +57,9 @@
     "system-font-css": "^2.0.2",
     "theme-ui": "^0.3.1"
   },
+  "optionalDependencies": {
```
hokei, line 60: OH MY GOD please tell me this was buried under a mountain of documentation or something I can't believe it turned out to be something this simple

next.config.js
```diff
@@ -5,15 +5,27 @@ const {
...
   BugsnagSourceMapUploaderPlugin
 } = require("webpack-bugsnag-plugins");
-const path = require("path");
 
 const APP_VERSION = require("./package.json").version;
 
+/**
+ * Set the AXISNOW_DECRYPT variable based on whether the package is available.
+ */
+let AXISNOW_DECRYPT = false;
```
hokei, line 14: Is this a normal thing to do?  put logic in trying to determine what an env variable is?

package.json
kristojorg: Indeed it was... and it doesn't really like work all the time but it works enough. Like you can't install an optional with npm I -o even though it says you can, but whatever it works

next.config.js
kristojorg: I think it's pretty normal. I'd put it in Utica except that it needs to be defined at build time and it is talking about the environment so I figured an env var seems right

kristojorg: I mean it's not literally an env var in the true sense. It's just getting set as a global, not like writing something to the node env or anything

README.md
```diff
@@ -76,10 +75,9 @@ The application will start at the base URL of `localhost:3000`.
 This app supports read online for encrypted books only in the AxisNow format, and if you have access to the [Decryptor](https://github.com/NYPL-Simplified/axisnow-access-control-web)
 
 To run with decryption:  
-- Set `axisnow_decrypt=true` in your app's `config.yml` file.
-- Run `git submodule update`
-- Verify that `axisnow-access-control-web` folder exists.  
+- Run `npm login --registry=https://npm.pkg.github.com`. You will need a Github Personal Access Token to use as your password.
 - Run `npm install` as normal.
+- The app will automatically pick up the installed optional `@nypl-simplified-packages/axisnow-access-control-web` package, and run with decryption enabled.
```
hokei, line 80: So if you're logged in and have access, it will always get the axisnow token?  Is it possible to need to build the app without the axis token, even if theoretically you have access to it?

hokei (approved):
This looks good.  After this deploys to docker, can you pull and run it from the container to make sure that all of that looks good?  (Also, I think it should still work even if we pass in the extra config but I'd try it anyway to see -- I'll make a PR to change that env variable still but we don't need ask Ho-Ling to drop everything and redeploy)

README.md
kristojorg: Hmmm maybe, but only really for dev purposes I think and in that case we would just delete the node_module

PR #182, opened 2020-10-26
Fetch errors

kristojorg:
This PR is in preparation for the login requirements PR. It makes errors more specific and consistent by always using the ProblemDocument format, even when errors originate within this app (as opposed to in the circ manager)


src/dataflow/withAppProps.ts
```diff
@@ -57,20 +52,18 @@ export default function withAppProps(
         props: {
           ...pageResult.props,
           library
-        }
+        },
+        // revalidate library-wide data once per hour per route
```
hokei, line 56: why ... ?

hokei:
so a lot of this looks really reasonable -- was passing errors through CPW and just displaying them always in the plans, or is there a specific login thing that needs it to work this way ?

src/dataflow/withAppProps.ts
kristojorg: Just in case someone has changed something in the CM configuration I guess. Like the colors or the logo, that way it will update eventually without restarting the server

kristojorg:
This is just cleaning some stuff up because we are going to be using these errors more specifically in the login stuff, so I wanted to make sure they all worked the same way. I think passing the CM errors through CPW was always desired, and having specific error messages when necessary is a good thing in general

src/errors.ts
```diff
@@ -46,7 +60,6 @@ function isProblemDocument(
 export class ServerError extends ApplicationError {
   // a default problem document
   url: string;
-  status: number;
   info: OPDS1.ProblemDocument = {
```
hokei, line 63: Still trying to figure out what exactly is sent back from the server - does the server send back an error of format "ProblemDocument" with title and detail?  Is there a better way to pass the server error through (like, is it useful to know that it's a problem here:  https://github.com/NYPL-Simplified/circulation-patron-web/pull/182/files#diff-8dea0ec348e50904c64a6ae0c6ccf215a11277514a8d447daa73eb45fa08bb09R111 vs somewhere else?

hokei (approved):

src/errors.ts
kristojorg: Basically the server sends back a `ProblemDocument` as JSON for most errors, or the `AuthDocument` for 401 errors. What do you mean a better way to pass it through? My idea here is basically... when we call `fetch`, catch any errors and construct a `ServerError` object. If the error returned contains a `ProblemDocument` or `AuthDocument` then set that on the error. If a `ProblemDocument` isn't provided for any reason (the network request itself failed, or the error was a 401), the we construct one ourselves. Throw this new `ServerError`.

Then, we can catch calls to our fetchers and display to the user information from the `ServerError` we receive on failure

PR #186, opened 2020-10-28
Startup fetching / Login requirement

kristojorg:
The goal of this PR is to support the auth protected OPDS feeds in Open eBooks. To do that, I changed a few things:

1. We now start the app by fetching the auth document, and this is the url that needs to be put in the config file, whereas previously it was the catalog root url. This is a breaking change because we will have to update our config files to work with that.
1. We now fetch the "search description" (which is basically a url template) on the client side when the app starts. The url for this is different depending on which specific feed we are viewing, but for the time being it should always return the same thing, so we keep it around and only fetch it once. We had to fetch it client side because it requires authentication now. 
1. We now catch 401s with `SWR` and display the auth modal. In the future we will make this behavior configurable so that we can redirect to the homepage on 401 (for Open eBooks)
1. I made a `useCollection` hook because we were using the same logic to fetch a collection in multiple places: lanes page, infinite book list, recommendations component...
1. I updated the `ApplicationError` so that the callee gets to pass in a custom title and status code if desired. Part of the ongoing evolution of how errors should work in this app


hokei:
does there need to be an update to `community-config.yml` ?

hokei:
looks like the vercel deploys are broken

src/hooks/useCollection.ts
```diff
@@ -0,0 +1,46 @@
...
+            title: "Fetch Collection Error",
+            detail: "An unknown error occurred while fetching the collection"
+          },
+          error
+        )
+    : undefined;
+
+  // extract the books from the collection and set them in the SWR cache
+  // so we don't have to refetch them when you click a book.
+  React.useEffect(() => {
```
hokei, line 41: I think this is causing it such that if you log in, log out and refresh, you will still see the catalog.

hokei: Might be a bigger/more serious bug, after I log out I can still click "my books" but the books themselves don't consider me logged in.  However, staying on this page a bit longer shows me the login modal again.

hokei:
I think you want to check caching issues in general when you're signed in on two browsers/tabs.  For example, signing out on one tab when signed in on another gives a strange state where the app thinks you're simultaneously signed in and signed out
Screen Shot 2020-11-16 at 10 48 31 AM

hokei:
And being signed in on more than one tab seems like a reasonable case because it totally makes sense to have a book open and be reading it on one tab, and browsing other books on a different tab.

src/interfaces.ts
```diff
@@ -224,6 +223,7 @@ export interface CollectionData {
   parentLink?: LinkData | null;
   shelfUrl?: string;
   links?: LinkData[] | null;
+  searchDataUrl: string | null;
```
hokei, line 226: how do you decide between `string | null` vs optional prop?  
(should this be optional to match shelfUrl? weirdly LibraryData.shelfUrl is required but `string | null`

src/pages/_error.tsx
```diff
@@ -3,21 +3,27 @@ import ErrorComponent from "../components/Error";
...
 
 Error.getInitialProps = ({ res, err }) => {
   if (err) track.error(err);
-  const statusCode = res?.statusCode ?? err?.statusCode ?? 404;
+  if (err instanceof ApplicationError) {
+    return {
+      errorInfo: err.info
+    };
+  }
+  const statusCode = res?.statusCode ?? err?.statusCode ?? 500;
```
hokei, line 21: more things that are hard to google, what does the ?? here mean?

src/pages/index.tsx
```diff
@@ -4,7 +4,16 @@ import { AppProps } from "dataflow/withAppProps";
...
 
 const CollectionPage: NextPage<AppProps> = () => {
-  return <ErrorComponent />;
+  return (
+    <ErrorComponent
+      info={{
+        title: "Page Not Found",
+        status: 404,
+        detail:
+          "This app does not have a home page. Url should contain a library slug: https://domain.com/<library>"
```
hokei, line 13: Is this a page that we expect an end user to see?  I think it might be less confusing to end users to remove this second sentence, and go, 
"This app does not have a home page.  Did you mean to visit one of these?" (where that second sentence is passed from the Library error)

hokei: More generally, I think the only situation in which the "did you mean to visit one of these" wording makes sense is when people get to `/index` without meaning to.  I feel like the generic 404 page should say something like "page not found, here are the links to return to catalog" instead of "did you mean to visit"

hokei: Thinking about legitimate situations where people can get to 404s - stuff like they have a book bookmarked and then the book is removed from the collection.  Something about "did you mean to visit" isn't helpful there.

src/test-utils/fixtures/auth.ts
```diff
@@ -62,5 +62,6 @@ export const loans: CollectionData = {
   title: "My Loans",
   lanes: [],
   books: makeFulfillableBooks(1),
-  navigationLinks: []
+  navigationLinks: [],
+  searchDataUrl: "/search-data-ur"
```
hokei, line 66: nit

src/pages/_error.tsx
kristojorg: The `??` is called nullish coalescing, it's basically defining a fallback if the value is `null` or `undefined`. This is similar to `const blah = maybe || default`, but it's safer because it only uses the fallback if the original is `null` or `undefined`, NOT if it is falsy.

kristojorg: And this is two of them chained, so it first looks in `res.statusCode` and then `err.statusCode` and if both of those are undefined then it defaults to 500

src/interfaces.ts
kristojorg: Basically there isn't really a big difference except that an optional doesn't have to be defined at all, and `null` is more specific in my mind. I am, however, slowly migrating these to only use the `| null` version because I know that in the future we will be fetching this data server side and there is a little particularity where Next.js doesn't like `undefined` values in `getStaticProps`. So yea, partially that and partially because I think it's slightly more specific than an optional

src/pages/index.tsx
kristojorg: Yea I tried to make it so that the 404 page only shows the "Did you mean to visit one of these" list when you are on the app root or an invalid library home page so: `/` or `/some-non-library`. It should be working that way here.

And no I don't expect that a user ever should see this page, but if you think we should remove the "Url should contain a library slug" sentence I can do that. I put that there to make it obvious for people new to the project what they did wrong

hokei: If this is only a page that developers or configuration-people can get to, I think it should stay.  

You're saying a normal user should never see this page, even if they accidentally go to "cpwurl.com" instead of "cpwurl.com/app" ?

src/interfaces.ts
hokei: sounds good

src/pages/index.tsx
kristojorg: No they would see it if they go to "cpwurl.com", you're right. Once the Open eBooks homepage is up, that will replace this route so they would see the homepage there. Buuut yea in other instances they would see this route, though I believe that the idea when deploying this for non open ebooks is to have cpwurl.com/lib1 and cpwurl.com/lib2 which are then aliased to catalog.lib1.com and catalog.lib2.com, so in that case the root url shouldn't really be accessible unless you go directly to cpwurl.com, which ideally only people "in the know" would do

src/dataflow/withAppProps.ts
```diff
@@ -58,20 +40,27 @@ export default function withAppProps(
...
+          ? e
+          : new ApplicationError(
+              {
+                title: "App Startup Failure",
+                detail: "Static props could not be fetched.",
+                status: 500
+              },
+              e
+            );
       // show the error page if there was an ApplicationError
```
hokei, line 55: this comment still relevant?  At this point everything is an ApplicationError

src/errors.ts
```diff
@@ -79,33 +82,33 @@ function isProblemDocument(
...
   authDocument?: OPDS1.AuthDocument;
 
   constructor(
     url: string,
     status: number,
     details: OPDS1.ProblemDocument | OPDS1.AuthDocument
   ) {
-    super("Server Error");
+    super(
+      isProblemDocument(details)
```
hokei, line 93: I'd make these if-elses as well or assign them to consts before calling super.

src/pages/index.tsx
hokei: OK if that's the case then I'm ok with it.  I'm just concerned about libraries that ask people to go to cpwurl.com/lib2  because people will definitely try to go to cpwurl.com instead

hokei:
I see references to this in the tests even though I don't think this is part of this ticket - what is a shelf?

hokei:
you also should update `sample-config.yml`, as well as the ##background section on the readme.  
(While you're at updating the readme, the `CONFIG_FILE=config.txt npm run start` part under ENV vars and building should be changed to a yaml

hokei:
also the example test in the readme no longer exists in the app

src/pages/index.tsx
kristojorg: Yea that's true. I think if/when that becomes a problem Lyrasis or whomever should make a decent home page experience where people can find their library, but that's for the future

src/errors.ts
kristojorg: I wanted to do that buuuut I got an eslint error that the super call has to be the first expression in a constructor. I guess I could ignore the eslint error though, what do you think?

kristojorg:
> I see references to this in the tests even though I don't think this is part of this ticket - what is a shelf?

The shelf is your loans. I think that the most common word for it is shelfurl but I've heard loans url also. It's kind of interchangeable to me, but maybe it's supposed to be one vs the other? I'm not totally sure.

I've udated the README and also that crazy ternary.

src/errors.ts
hokei: I think I value knowing what's going (especially when you get one of these pages and don't know how you got here) more than making ESlint happy, so lets ignore the error

hokei (approved):

src/errors.ts
kristojorg: Okay I went ahead and changed this!

PR #187, opened 2020-10-29
Added buildID to next build

hokei:


next.config.js
```diff
@@ -27,6 +27,15 @@ const config = {
     APP_VERSION,
     AXISNOW_DECRYPT
   },
+  generateBuildId: async () => {
+    return await Git.Repository.open(".")
+      .then(repo => {
+        return repo.getHeadCommit();
+      })
+      .then(commit => {
+        return commit.sha();
```
hokei, line 36: I tested it by putting a console log before this, but I'm not really sure how to better test this.

kristojorg (approved):
Okay I tested by building multiple times and verifying that the build id was the same between versions. The only way we will know if that really has solved our problem is to deploy and include bugnsag credentials to track errors

PR #189, opened 2020-11-06
Fix Bugsnag

kristojorg:
1. Update how we source the config file so we can use it in `next.config.js`. It is no longer loaded with val-loader, but fetched and parsed synchronously using `execSync` in `next.config.js`. We then can use it in that file and set it as an env var so it can now be imported from `source/utils/env`. This means we now have access to the `bugsnagApiKey` in `next.config.js` and can use it to upload source maps correctly.
1. Update how we upload source maps to bugsnag.
1. Add `track.error` to send errors to bugsnag and log them to the console (helpful for when viewing AWS logs hopefully).
1. Add helpful context to bugsnag errors, like fetch parameters and the app config file.
1. Add an optional `instance_name` to the config file so that instances can be differentiated in bugsnag.
1. Track errors when fetching library data
1. Track errors when trying to fulfill a book
1. Track errors when trying to borrow a book
1. Track uncaught errors caught by error boundary
1. Update error boundaries to display errors more consistently and approrpriately.


hokei:
how are you verifying that this all is actually being logged?  
Also what is happening here that's changed from the last bugsnag ticket?  

Like:

src/analytics/bugsnag.ts
```diff
@@ -0,0 +1,46 @@
...
+    appVersion: BUILD_ID,
+    plugins: [new BugsnagPluginReact()],
+    appType: IS_SERVER ? "node" : "browser",
+    // the release stage is based on the github branch it is deployed from.
+    releaseStage:
+      GIT_BRANCH === "production"
+        ? "production"
+        : GIT_BRANCH === "qa"
+        ? "qa"
+        : "development",
```
hokei, line 26: Should this be "dev" ?

src/components/bookDetails/index.tsx
```diff
@@ -144,40 +138,4 @@ const SimplyECallout: React.FC<{ className?: "string" }> = ({ className }) => {
   );
 };
 
-const Error: React.FC<{ info?: ProblemDocument }> = ({ info }) => {
```
hokei, line 147: yay!

src/config/fetch-config.js
```diff
@@ -35,10 +42,30 @@ async function fetchConfigFile(configFileUrl) {
...
+    bugsnagApiKey: unparsed.bugsnagApiKey || null,
+    gtmId: unparsed.gtmId || null,
+    companionApp,
+    showMedium
+  };
 }
 
+// get the config and print it to stdout so next.config.js can use it
+getAppConfig(process.env.CONFIG_FILE).then(val => {
+  console.log(val);
```
hokei, line 68: Why is it important to print it?

src/errors.ts
```diff
@@ -63,7 +82,7 @@ export class ServerError extends ApplicationError {
   info: OPDS1.ProblemDocument = {
     detail: "An unknown error server occurred.",
```
hokei, line 83: Is this a fallback or is this a default?  I'm not seeing anything here that's checking for info from the server and passing it, (if it's available) ... that's probably better than this, right?  (Alternatively, if we don't want to expose server error details to the user, is there a way we can log them while still exposing this to the front end?)

src/hooks/useBorrow.ts
```diff
@@ -17,15 +17,14 @@ export default function useBorrow(isBorrow: boolean) {
   const buttonLabel = isBorrow ? "Borrow this book" : "Reserve this book";
 
   const borrowOrReserve = async (url: string) => {
-    setLoading(true);
     clearError();
     if (!token) {
       // TODO: register a callback to call if the sign in works
```
hokei, line 22: Do you remember what this was? (me cheating, this might be useful for a landing page login problem I'm mulling)

src/hooks/useError.ts
```diff
@@ -9,12 +10,11 @@ export default function useError() {
...
   // for network errors
   function handleError(e: any) {
-    // TODO: track error to bugsnag
+    track.error(e);
     if (e instanceof ServerError) {
       setError(`Error: ${e.info.detail}`);
       return;
     }
-    console.error(e);
     setError("Error: An unknown error occurred.");
```
hokei, line 18: What's an example of a network error that wouldn't be an instance of ServerError?  If the internet cuts out or something, that's gonna get caught by the browser, right?

next.config.js
```diff
@@ -52,21 +63,12 @@ const config = {
...
-      config.plugins.push(new BugsnagBuildReporterPlugin(config));
-      config.plugins.push(new BugsnagSourceMapUploaderPlugin(config));
-    }
-
-    // source the app config file and provide it using val-loader
-    config.module.rules.push({
-      test: require.resolve("./src/config/load-config.js"),
-      use: [{ loader: "val-loader" }]
-    });
+    // pass the APP_CONFIG as a global
```
hokei, line 66: can you explain how globals work and why they're desired here?  Isn't it generally better and more explicit to have constants be specifically imported?

src/analytics/bugsnag.ts
kristojorg: It just is a name that gets sent to bugsnag for display in the UI, so it's not a big deal. They use "development" by default so that's what I went with too

hokei:
A general high level question:  It looks like the app is now going:  
Set config in `next.config.js` => read into `env` => use 

Why do variables need to be read into `env` instead of being accessed from `next.config.js` (or accessed from global `APP_CONFIG`? )

src/analytics/bugsnag.ts
hokei: ok! Doesn't bugsnag not log anything if it's development though?

src/config/fetch-config.js
kristojorg: This is because in `next.config.js` I have to fetch the config file, which is asynchronous. In order to fit that into the synchronous way next.config.js is run, I have to do this using `execSync` which spawns a child process and waits at that line until the child process completes, and then takes whatever is on stdout as the resulting value

src/hooks/useBorrow.ts
kristojorg: The callback you mean? This is just indicating that eventually we can make the app re-try the failed fetch if a 401 error occurs and then the user successfully logs in. I don't immediately see how this would help you issue necessarily :/

src/hooks/useError.ts
kristojorg: One would be we try to download a book, get redirected to some S3 bucket somewhere, and that rejects with a 401 or a 404 or 500 or whatever. So it will be an error from a server, but not specifically from the Circulation Manager, which means it won't have the `ProblemDocument` format. The `ServerError` is specifically for errors that are from the CM. So maybe I should rename it `CircManagerError`?

next.config.js
kristojorg: Yea that is true, it is generally better to do that mainly for readability, because it will be more clear where an imported value comes from. Having tons of global things can get really confusing. I would rather have this be imported from somewhere, but sadly we can't really do that in this case. We run this fetch script to get the APP_CONFIG value from the network or file system, and then we need to provide it to the app somehow. Like, it's a dynamic thing that is defined at build time, not something static we can just define in a file and then import. 

One option would be to run the fetch script whenever we need the config file or even once on app startup, but I'd rather not do that because we have it already at build time, so I figure it's not soooo bad to define one global value like this.

kristojorg:
> Why do variables need to be read into env instead of being accessed from next.config.js (or accessed from global APP_CONFIG? )

Can you give me an example of what variables you mean? You mean the env vars like `AXISNOW_DECRYPT`? We could set those globally the same way we do the `APP_CONFIG`, buuut idk then we would have a lot of global variables floating around. I guess maybe we could define the APP_CONFIG on `process.env` just like these other env vars. Do you think that would make more sense?

src/errors.ts
kristojorg: This is just a default/fallback. I set it to whatever the server returns in the constructor down on like 107. I first check if the server returned a problem document and then set `this.info ` to that if it exists

hokei:
> Can you give me an example of what variables you mean?

The ones that are throwing me the most are the github variables -- They're set in `next.config.js` but then imported into `env.ts` to be used in bugsnag logging.  Why is it that bugsnag can't just directly ask `next.config.js` 

https://github.com/NYPL-Simplified/circulation-patron-web/pull/189/files#diff-ff9fe1e089b0f33fe0d0f46047d4ce2d9d95cd3cfb6f7f816bde029e485aba0fR14

kristojorg:
Basically `next.config.js` is a config file that customizes certain settings, like our webpack config. It also controls how nextjs exposes env vars. In order for an env var to be exposed to the app, it either has to be named `NEXT_PUBLIC_xyx` or it has to be defined in the `env` key of the nextjs config in `next.config.js`. They do this because they don't want you to accidentally expose a private env var to the web frontend, like a secret token or something. So you have to explicitly tell nextjs to expose it on the web.

So basically anything you pass in to the `env` key here will be set on `process.env` in the app, so I included the git variables there for use when configuring bugsnag. 

> Why is it that bugsnag can't just directly ask next.config.js

If you mean, why can't we just export the value and import it in `bugsnag.ts`, that's because `next.config.js` isn't something that gets bundled with the app, it's a file that comes _before_ bundling to configure how we bundle the app. So we can't import from it. It would be like trying to import `webpack.config.js` in a normal app and use something that was run in there on the web. Does that answer your question?

src/analytics/bugsnag.ts
kristojorg: No I set it to log in development if the api key is present. To turn it off we just remove the api key from the config file we are using in development

kristojorg:
Okay so, I have done a few things:

1. The `APP_CONFIG` variable is no longer global, it is defined as an env var on process.env just like the other env vars. 
1. I updated the dockerfile so that we could access the git info we need when building the app. 
1. I added some logs when `next.config.js` gets run so that we can see app info easily in the terminal.

hokei (approved):

PR #190, opened 2020-11-09
 Landing page

hokei:
Adds the Open Ebooks landing page 
- The log in buttons are not functional (except for the one on the top right) 
- The top right login button currently doesn't redirect to catalog (we should probably get that functionality in before this gets merged)
- The backgrounds under the books aren't ... quite right.  If someone has a way to do this that doesn't involve breaking the flexbox I'm all ears?  

Main changes:  
- Layout page now can be configured to include header, footer or both.  
- Library data is fetched before login.


src/auth/AuthModal.tsx
```diff
@@ -20,7 +20,10 @@ import ExternalLink from "components/ExternalLink";
 import BasicAuthButton from "auth/BasicAuthButton";
 import LoadingIndicator from "components/LoadingIndicator";
 
-const AuthModal: React.FC = ({ children }) => {
+const AuthModal: React.FC<{ selectedMethod?: AppAuthMethod }> = ({
+  selectedMethod: preSelectedAuth,
```
hokei, line 24: How do I end up passing something in here?  (<AuthModal> itself is called in ContextProvider ... or is this the wrong place to be trying to pass this?  )

hokei:
@kristojorg do you know why the vercel preview is not showing the landing page?  It's checking for `companion_app=openebooks` but I'm wondering if there's something else that just makes it automatically throw a 404 on landing?

src/components/OpenEbooksLanding.tsx
```diff
@@ -0,0 +1,369 @@
...
+    { alt: "book cover text 1", imgHref: "/img/middlegrades-3.jpg" }
+  ],
+  HighSchool: [
+    { alt: "book cover text 1", imgHref: "/img/highschool-1.jpg" },
+    { alt: "book cover text 1", imgHref: "/img/highschool-2.jpg" },
+    { alt: "book cover text 1", imgHref: "/img/highschool-3.jpg" }
+  ]
+};
+
+const OpenEbooksLandingComponent = () => {
```
hokei, line 37: Would love ideas on how to make this component less long and hard to read.  It's mostly using one-off things that don't belong anywhere else in the app, but I'm not sure a 370 line file is the solution to this.

src/components/OpenEbooksLanding.tsx
```diff
@@ -0,0 +1,369 @@
...
+      <div
+        sx={{
+          flex: "1 1 auto",
+          display: "flex",
+          flexDirection: "column"
+        }}
+      >
+        <OpenEbooksHero />
+        <div
+          sx={{
```
hokei, line 57: Things like this, where the container should always have a certain margin, would normally be put in a .wrapper or .container class but I'm not sure what the equivalent is here?

src/components/OpenEbooksLanding.tsx
```diff
@@ -0,0 +1,369 @@
...
+          coverLocation === "right" ? "row" : "row-reverse"
+        ]
+      }}
+    >
+      <Stack
+        direction="column"
+        sx={{
+          minWidth: "300px",
+          flex: "1",
+          mx: [0, 2],
```
hokei, line 321: is there a way to do CSS `last-child-of` here?

hokei:
Also my first time trying to write CSS in JS, so please let me know if there are best practices that I'm just not aware of.  It feels super foreign to me, I'm sure I've messed stuff up!

kristojorg:
Okay this looks good! I'm gonna make a PR to your branch with a few architectural changes, but I also have a couple design things I'll note down as I come across them:

1. The sign in button I think should be white and the header I believe needs some padding
**Design**
image
**This branch**
image

2. I think the Sign in buttons should be left-aligned to their respective columns according to the designs that I see.
3. I think the text in the mobile shoutout section should be middle-aligned in the colored block (vertically)
4. The popular books section looks like it maybe needs one more level of padding to match the designs? Not a big deal tho.
5. The FAQ paragraph I think should have a max-width
6. When I go down to really small widths I see the background not filling the whole row on some sections:
image
7. One mobile the "Popular x grade books" headers should be centered, and the images I think are filling too much space (in fact this might be what's causing the other sections to be too narrow and not fill the full width on mobile. 

Okay I'll make a PR with a couple non-design changes so you can review that

src/components/OpenEbooksLanding.tsx
kristojorg: Honestly I think it is the solution. There just isn't really any other way around it. You could divide it up into components so that it is maybe easier to see what is what, but at the end of the day you need 370 odd lines of custom html one way or another basically

src/components/OpenEbooksLanding.tsx
kristojorg: If it is on-off, then this is the right way to do it. If you're reusing it then you can always:
```
const container = { flex: 2, maxWidth: 1100, mx: "auto"}

<div sx={container} />
or 
<div sx={{...container, bg: "custom-style"}} />
```

kristojorg: If it's something that you only use here, then all you accomplish by extracting it to a class is moving the relevant code for this div into two different places instead of one

src/components/OpenEbooksLanding.tsx
kristojorg: I think that's what `Stack` should be doing by default. But you can use css selectors like that in theme-ui `{"&:last-child-of": { color: "red" } }

src/pages/[library]/book/[bookUrl].tsx
```diff
@@ -5,8 +5,10 @@ import LayoutPage from "components/LayoutPage";
 import { NextPage, GetStaticProps, GetStaticPaths } from "next";
 
 const BookPage: NextPage<AppProps> = ({ library, error }) => {
+  const props = { library: library, error: error };
+
   return (
-    <LayoutPage library={library} error={error}>
+    <LayoutPage props={props}>
```
Comment (kristojorg) [inline src/pages/[library]/book/[bookUrl].tsx:11]: Curious why you did this?

Comment (hokei) [reply src/pages/[library]/book/[bookUrl].tsx]: I don't think I did it on purpose haha

kristojorg:
I think we are okay to merge this into dev before the auth redirect stuff is done. It's broken until we get that auth refactor working anyways, so we just need to hold off from merging to `qa`

hokei:
> When I go down to really small widths I see the background not filling the whole row on some sections: 

I'm not seeing this error, are you on chrome?  

Screen Shot 2020-11-19 at 11 10 30 AM

Wonder if it's using something cached or something ...

kristojorg:
Okay just the stack issue in the logins section and the min-width on the book covers causing issues, but otherwise design looks good to me!

src/components/OpenEbooksLanding.tsx
```diff
@@ -0,0 +1,445 @@
...
+      </ErrorBoundary>
+    </Page>
+  );
+};
+
+const OpenEbooksLandingComponent = () => {
+  const { authMethods } = useLibraryContext();
+  const cleverMethod: CleverAuthMethod = authMethods.find(
+    method => method.type === OPDS1.CleverAuthType
+  ) as CleverAuthMethod;
```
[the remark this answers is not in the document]
hokei: I'm still not sure how to avoid the cast -- `CleverMethod` gets passed into `CleverButton` later, and we know that if "blah.type === cleverMethod" then it must be of that type, right?  I think this is a situation where TS is a bit dense? (I put the check in, though, that seems like a good idea)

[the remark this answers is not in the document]
kristojorg: If you write the check before you use them, then it should narrow the type to be defined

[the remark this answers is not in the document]
kristojorg: Like it should know after that statement that it is defined

kristojorg (approved):
looks good!

PR #191, opened 2020-11-09
Links book covers and titles from list to book page

hokei:


src/theme/variants.ts
```diff
@@ -62,13 +62,23 @@ const variants = {
         fontStyle: "italic"
       }
     },
-    textLink: {
```
kristojorg, line 65: Just to make sure, did you do a search to see if this was used anywhere else previously that we should now change to `link.text`?

src/components/BookList.tsx
```diff
@@ -131,18 +132,26 @@ export const BookListItem: React.FC<{
...
+        </Link>
         <Stack direction="column" sx={{ alignItems: "flex-start" }}>
           <div>
-            <H2 sx={{ mb: 0, variant: "text.body.bold", display: "inline" }}>
-              {truncateString(book.title, 50)}
+            <H2 sx={{ mb: 0, display: "inline" }}>
+              <Link
+                bookUrl={book.url}
+                sx={{ variant: "text.link.bold", color: "brand.primary" }}
+                aria-label={`${book.title}`}
```
kristojorg, line 151: nit: I don't think you need the wrapping ` marks, should just be able to set it to `aria-label={book.title}` right?

kristojorg (approved):
I'll go ahead and improve this with one comment: I used to have the app set up like this actually, and then I started getting errors when running with react-axe telling me that you shouldn't have two links to the same place right next to each other... so I asked in slack and got the recommendation that only one link should be there. At the time I ended up choosing the _whole card_ on the lanes view, and one specific "View Book Details" button on the list view... I think it's worth asking about this, but if you get the green light then go ahead and merge it

kristojorg:
PS: The preview links aren't working right now because I have updated the config file to the format needed for my login branch. Once that gets merged they will start working again : )

src/theme/variants.ts
hokei: yes, I've changed all of the ones that I could find.

hokei:
Hmmm I will run react-axe on this first.  This was a request from Risa, so we will have to figure out how to make it work - it doesn't make sense that two things can't link to the same place ...

hokei:
React-axe doesn't seem to flag this anymore ... Will merge this.

PR #214, opened 2020-11-19
Updates how slug is defined

kristojorg:


src/components/open-ebooks-landing/index.tsx
```diff
@@ -34,6 +41,18 @@ const popularBooks = {
   ]
 };
 
+const LandingPage: NextPage<AppProps> = ({ library, error }) => {
+  return (
+    <Page library={library} error={error}>
+      <ErrorBoundary>
+        <GlobalStyles />
+        <OpenEbooksLandingComponent />
```
hokei, line 49: OH WELL THIS WORKS TOO XD

kristojorg: hehe

src/dataflow/withAppProps.ts
```diff
@@ -19,8 +19,8 @@ export type AppProps = {
 };
 
 export default function withAppProps(
-  defaultLibSlug?: string,
-  pageGetServerSideProps?: GetStaticProps
+  pageGetServerSideProps?: GetStaticProps,
```
hokei, line 22: is there a suggested ordering for these things?

kristojorg: No, I just know that in the future we will use `pageGetServerSideProps` more frequently (on most pages) and the `defaultSlug` only on one page, so I wanted to put the `pageGetServerSideProps` first so that we didn't have to call it like this all the time: `withAppProps(undefined, async () => {...})`

PR #215, opened 2020-11-19
Bugfix/simply 3214 back forward lane

kaplanmichael:
Hi there!  I tried a bit of a different approach to determine when we're at the end of a lane.  Rather than using scroll-related information, which was inconsistent due to the scroll setTimeout, I tried a DOM observable.  Risa confirmed in an earlier chat that each lane will always have a "See More" card in it, so I set up an observable such that when 100% of the "See More" card is visible in the viewport, it will register as having scrolled to the end of the lane.

If you like this approach, we could apply it to the start of the row as well--currently, you have to have scrolled past the first book into order for the back button to enable, which might not be what you want.

I hope it was OK that I added a package?  I wasn't sure if that was cool; let me know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "react-dom": "16.13.1",
     "react-hook-form": "^4.9.2",
     "react-hot-loader": "^4.12.20",
+    "react-intersection-observer": "^8.30.3",
     "react-redux": "^7.1.0",
     "reakit": "^1.2.4",
     "request": "^2.88.2",
--- a/src/components/Lane.tsx
+++ b/src/components/Lane.tsx
@@ -19,7 +19,9 @@ type BookRefs = {
 type CurrentBook = {
   index: number;
   /**
-   * The following dictates whether we snap to a book
+   * The following dictates whether we snap to a book. 
+   * "Snapping" to a book means we scroll the currentBook's
+   * left edge so that it is up against leftmost edge of the lane
    */
   snap: boolean;
 };
--- a/src/components/Lane.tsx
+++ b/src/components/Lane.tsx
@@ -95,7 +95,7 @@ const Lane: React.FC<{
 
   // will be used to set a timeout when the browser is auto scrolling
   const timeoutRef = React.useRef<ReturnType<typeof setTimeout> | null>(null);
-  const browserScrollTime = 3000; // guess how long the browser takes to scroll
+  const browserScrollTime = 1000; // guess how long the browser takes to scroll
 
   /**
    * This effect is used to snap us to a particular book when the
--- a/src/components/Lane.tsx
+++ b/src/components/Lane.tsx
@@ -12,6 +12,7 @@ import Stack from "./Stack";
 import { AnyBook, LaneData } from "interfaces";
 import Link from "components/Link";
 import { Text } from "components/Text";
+import { useInView } from 'react-intersection-observer';
 
 type BookRefs = {
   [id: string]: React.RefObject<HTMLLIElement>;
@@ -66,18 +67,24 @@ const Lane: React.FC<{
   const [isBrowserScrolling, setIsBrowserScrolling] = React.useState<boolean>(
     false
   );
+  const [isAtEnd, setisAtEnd] = React.useState<boolean>(
+    false
+  );
   // we need a ref to the UL element so we can scroll it
   const scrollContainer = React.useRef<HTMLUListElement | null>(null);
 
   // vars for when we are at beginning or end of lane
+  
   const isAtIndexEnd = currentBook.index === filteredBooks.length - 1;
   const isAtScrollEnd = !!(
     scrollContainer.current &&
     scrollContainer.current.scrollLeft ===
       scrollContainer.current.scrollWidth - scrollContainer.current.offsetWidth
   );
-  const isAtEnd = !!(isAtIndexEnd || isAtScrollEnd);
-  const isAtStart = currentBook.index === 0;
+  //const isAtEnd = !!(isAtIndexEnd || isAtScrollEnd);
+  //const isAtEnd = !!(currentBook.index + 1 > filteredBooks.length);
+  //const isAtStart = currentBook.index === 0;
+  const isAtStart = !!(currentBook.index - 1 < 0);
 
   /** Handlers for button clicks */
   const handleRightClick = () => {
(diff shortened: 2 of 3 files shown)


hokei:
question:  Why is it necessary to distinguish between being at scroll end and being at index end?  Can't we insert a "see more" extra card at the end of the returned lane content?

kaplanmichael:
> question: Why is it necessary to distinguish between being at scroll end and being at index end? Can't we insert a "see more" extra card at the end of the returned lane content?

I agree--I don't think it actually _is_ necessary to differentiate those anymore.  I ended up removing that code in one of the cleanup commits in this PR.

src/components/Lane.tsx
```diff
@@ -212,10 +212,15 @@ const Lane: React.FC<{
...
 };
 
-const SeeMoreBlock: React.FC<{ url: string; title: string }> = ({
+const SeeMoreBlock: React.FC<{ url: string; title: string, setIsAtEnd: (arg: boolean)=>void }> = ({
   url,
-  title
+  title,
+  setIsAtEnd
 }) => {
+  const { ref, inView, entry } = useInView({
```
kristojorg, line 220: Personally, I would probably move this hook up to the lane component and then just pass the ref down to the `SeeMoreBlock`. That would also mean you wouldn't need to have the state variable:

```
// in Lane component
const { ref, inView: isAtEnd, entry } = useInView(...);
```

I know that component is pretty huge right now, but it just seems like we are duplicating this state unnecessarily to me and moving this up to Lane would make things simpler ultimately

kristojorg (changes requested):
This looks great! It's a good approach IMO. If I remember correctly, I didn't do this initially because we needed to maintain support for IE11, which doesn't support the IntersectionObserver api (grrr). Things might have changed since then, though. I saw some chatter about other NYPL products no longer supporting IE11. 

I would love to not support IE so I'll ask in standup

kristojorg:
Well it sounds like we don't need IE support so this looks like a great solution!

src/components/Lane.tsx
kaplanmichael: Done!

src/components/Lane.tsx
```diff
@@ -212,10 +221,7 @@ const Lane: React.FC<{
   );
 };
 
-const SeeMoreBlock: React.FC<{ url: string; title: string }> = ({
-  url,
-  title
-}) => {
+const SeeMoreBlock = React.forwardRef<any, SeeMoreBlockProps>((props, ref) => {
```
kristojorg, line 224: This should be `forwardRef<HTMLListElement, SeeMoreBlockProps>` I believe

src/components/Lane.tsx
```diff
@@ -227,9 +233,11 @@ const SeeMoreBlock: React.FC<{ url: string; title: string }> = ({
         mx: 2,
         color: "ui.white"
       }}
+      ref={ref}
+      {...props}
```
kristojorg, line 237: I don't think you need this props spread, but maybe I'm wrong?

src/components/__tests__/Lane.test.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 import { render, fixtures } from "../../test-utils";
 import Lane from "../Lane";
 import { LaneData } from "interfaces";
+import "react-intersection-observer/test-utils";
```
kristojorg, line 5: we need to import this in `src/test-utils/index.tsx` so that it's imported for all test runs, because some tests do render the Lane component internally (such as the book details page, which renders the lane to show recommendations). That _should_ solve the failing tests

package.json
```diff
@@ -17,8 +17,12 @@
...
     "test": "TZ=UTC jest --watch",
     "test:ci": "jest --ci",
-    "lint": "eslint '*/**/*.{ts,tsx}'",
-    "lint:fix": "eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
+    "lint": "run-script-os",
+    "lint:default": "eslint '*/**/*.{ts,tsx}'",
+    "lint:win32": "eslint **/*.{ts,tsx}",
+    "lint:fix": "run-script-os",
+    "lint:fix:default": "eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
+    "lint:fix:win32": "eslint **/*.{js,ts,tsx} --quiet --fix",
```
kristojorg, line 25: If you don't mind, I'd like to find a different way of doing this. After reading that article, it seems like this shouldn't be necessary... maybe we just need to change the glob itself and I think it should always be wrapped in single quotes so that eslint expands it instead of having it expanded by the shell

kristojorg (changes requested):
Just a few things!

src/components/Lane.tsx
kristojorg: Might be `HTMLLIElement` actually... but something like that

kaplanmichael: Good catch! I had meant to go back and change that.  Anyway, it's fixed now!

src/components/Lane.tsx
kaplanmichael: Fixed.

src/components/__tests__/Lane.test.tsx
kaplanmichael: Fixed.

package.json
kaplanmichael: Fixed!

src/components/Lane.tsx
```diff
@@ -212,10 +212,15 @@ const Lane: React.FC<{
...
+const SeeMoreBlock: React.FC<{ url: string; title: string, setIsAtEnd: (arg: boolean)=>void }> = ({
   url,
-  title
+  title,
+  setIsAtEnd
 }) => {
+  const { ref, inView, entry } = useInView({
+    threshold: 1
+  });
+  inView ? setIsAtEnd(true) : setIsAtEnd(false)
```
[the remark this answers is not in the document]
kaplanmichael: @kristojorg now that the code has been changed a bit, is this concern still applicable?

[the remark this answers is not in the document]
kristojorg: nope! all good now

kristojorg (approved):
Looks like this did away with `lint:fix` but that's okay. I don't think it was commonly used and you can do the same thing with `npm run link -- --fix`

hokei (approved):

PR #216, opened 2020-11-23
Login flow

kristojorg:
This PR redoes the login flow so that the selected auth method is stored in the browser URL bar. This will allow us to link to a specific login method directly from the home page (or elsewhere).

It proved very difficult to make this happen while using a modal for auth. Instead, the login page is now its own route. We can check with Risa if this is okay. If not, this was a necessary step on the way to making it a modal again, so that'll just take a bit more work. However, there are some challenges with the modal method surrounding accessibility. For example, modals are only really supposed to open in response to a user action, but the login modal can open without user interaction, such as when they visit a page and fetching the page data fails.

## Important Changes

- There is now a `/login` route which shows the login selector and a `/login/[methodId]` route which shows the handler for that method.
- There is now only one `AuthButton`. This button shows the appropriate styles (like background image etc) and navigates to the specific url for the method.
- There are now "Auth Handlers" instead of "Forms". Only one of these is actually a "Form" (the `BasicAuthHandler`). The `SamlAuthHandler` and `CleverAuthHandler` simply redirect the user to the external url for those auth methods when they're rendered. 
- In order to start a login, you now just redirect the user to the appropriate url. This can be somewhat tricky, so I made a set of utilities in `useLogin` for doing that. 

## Other Notes: 

- I also upgraded us to Next.js 10, which brings some improvements and not really any breaking changes. I needed to do this to let us define urls as objects.
- I had to update storybook for a bugfix and I also removed `val-loader` which we don't use but was still in there.
- There is still a flash of "401 Unauthorized" when you visit a page without being logged in before you actually get redirected to the login page. We can fix this but I think it should be a separate PR.


hokei:
clicking the buttons from the login page doesn't seem to work

hokei:
To help review, I'm gonna change the base to `landing-page` until it gets merged.

kristojorg:
Okay @hokei this should be ready to go

kristojorg:
Okay haha I lied, but now it should be good to go for review provided um yea... provided things work

hokei:
The buttons seem to not be working at all, probably a small typo somewhere?

src/auth/AuthProtectedRoute.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <div
+      sx={{
+        flex: 1,
+        display: "flex",
+        justifyContent: "center",
+        alignItems: "center",
+        flexDirection: "column"
+      }}
+    >
+      <h4>You need to be signed in to view this page.</h4>
```
hokei, line 42: wanted to make sure, this unauthorized page shows up with just this text, right?  Is it possible to instead redirect to the login page?  This would mirror the modal functionality better.

hokei: Testing it with the working buttons - seems like it's already doing this, please ignore me!

hokei:
If you have a direct link to a book detail page can you see it if you're not logged in?
(the behavior I'm seeing is that if you log out from a book detail page you get redirected to that same page, but if you try to get back to catalog from that page you get redirected to login)

hokei:
I'm gonna call this a bug unless it's trivially easy - 
If I log in with clever, navigate to the home page and click "log in with firstbook" it will flash the firstbook login page before redirecting

I think it /should/ probably disable the firstbook login button if I'm already logged in with clever but I have no idea how to even make that work.

hokei:
another task for later:  Can we get the login comboboxes/basic login into storybook? Maybe another good one for @kaplanmichael

src/auth/CleverAuthHandler.tsx
```diff
@@ -0,0 +1,51 @@
...
+  method
+}) => {
+  const { token } = useUser();
+  const { fullSuccessUrl } = useLoginRedirectUrl();
+
+  const authenticateHref = method.links?.find(
+    link => link.rel === "authenticate"
+  )?.href;
+
+  // double encoding is required for unshortened book urls to be redirected to properly
```
hokei, line 22: why?

src/components/ClientOnly.tsx
```diff
@@ -13,3 +13,16 @@ export default function ClientOnly({
 
   return mounted ? <>{children}</> : null;
 }
+
+export function clientOnly<T>(Component: React.ComponentType<T>) {
```
hokei, line 17: what does this do?

hokei:
The login doesn't seem to work in the CPW vercel deploy, any ideas why?  Haven't tested the combobox yet, though I think if we keep going on a separate login page flow, we may just want a list of buttons since it's all on its own page.

hokei (commented):
It mostly looks good, let's just make sure BookDetails doesn't need to be wrapped in Auth and figure out what's going on with the community deploy.

src/auth/AuthProtectedRoute.tsx
kristojorg: : )

src/auth/CleverAuthHandler.tsx
kristojorg: Yea that's a good question. It's always been like this... I tried taking it out once and it messed everything up so here we are haha

src/components/ClientOnly.tsx
kristojorg: This just makes it render nothing on the server, which is necessary when we use things that only work in the browser

kristojorg:
Community deploy is working okay, it's just giving CORS errors, which is expected on this URL (you can disable cors via a browser extension to test if you need to)

src/auth/CleverAuthHandler.tsx
hokei: lol have fun answering that question every time this line gets touched lol

hokei (approved):

PR #217, opened 2020-11-24
Preparing 4.0.0 for release to production

hokei:


hokei:
HOW IS THIS EVEN FAILING.

kristojorg:
I rerean the job and it looks like the failure was just a hiccup. The whole outdated thing is because apparently the previous commits from QA were merged without a fast forward... which created new commits. I need to figure out why that is so it doesn't keep happening. For now the solution is to rebase `qa` on `production` and force push that to here. I just did it locally and there were no code changes, just git refs

kristojorg (approved):

PR #221, opened 2020-11-24
[SIMPLY-3310] Refactor card to using the same sizing and aspect ratio…

kaplanmichael:
…n of the book cover images.  Alphabetically sort CSS-in-JS properties for enhanced readability.

https://jira.nypl.org/browse/SIMPLY-3310


src/components/Lane.tsx
```diff
@@ -289,15 +315,15 @@ const LaneErrorFallback: React.FC<FallbackProps> = () => {
   return (
     <div
       sx={{
-        display: "flex",
-        justifyContent: "center",
         alignItems: "center",
-        height: BOOK_HEIGHT,
-        py: 3,
-        px: 2,
         backgroundColor: lighten("warn", 0.35),
```
hokei, line 319: what autoformatting tools are you using?  It looks like you've got something that automatically alphabetizes these props, which I'd love to make consistent with all of our environments!

kaplanmichael: haha yeah, the alphabetizing of props (and CSS rules) is one of those small things I started doing a few years ago that I found to be really helpful in ways that are hard to measure.  What I am doing currently is using an extension in VSCode called "Sort Lines" which allows me to highlight a block of text, right-click it, and then choose to sort it from the contextual menu.  I haven't looked into it yet, but I'll bet there probably is a way to have the lint --fix do that automatically.

src/components/Lane.tsx
```diff
@@ -219,33 +221,57 @@ const SeeMoreBlock: React.FC<{ url: string; title: string }> = ({
...
         sx={{
-          bg: "brand.primary",
-          display: "flex",
-          justifyContent: "center",
           alignItems: "center",
-          textAlign: "center",
+          backgroundColor: "ui.gray.lightWarm",
+          border: "solid",
+          borderRadius: "card",
+          display: "flex",
```
kristojorg, line 238: Do we need this background color and border style on this element? It seems like the inner Link is what is styling the element the way we ultimately want, so maybe we can get rid of these?

src/components/Lane.tsx
kristojorg: I'm not opposed to this, but I think for the sake of git diffs and PR review we should try to avoid making formatting changes like this in PRs... it just makes it difficult to tell what was actually changed and what was just reordered. If we want to sort things like this then I agree a lint rule would be a good way to go. We currently format on lint with prettier and this could be step in that process as well. We could then add the rule and format everything in one PR

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 4.1.0
 - Add: Landing page for Open Ebooks
 - Fix: Fix our `APP_CONFIG` mocking in storybook so that `showMedium` and `companionApp` can be set via the toolbar once again.
+- Fix: Refactor See More card to use the same sizing and aspect ratio of the book cover images.
```
kristojorg, line 6: All the versions you see in the changelog will already be released... so what we want to do instead is make a new header: `### UNRELEASED CHANGES` at the top of the file and then add this line under that header. When we then go to make a release, we will check those changes to see if it should be patch, minor or major, and then create a new PR that bumps the version number and changelog and we'll merge that through dev->qa->production eventually

kristojorg (commented):
Looks good! Just a couple minor things

src/components/Lane.tsx
kaplanmichael: Agreed.  Yeah, if anything, this type of stuff could be done in a separate "Code Cleanup" PR.

src/components/Lane.tsx
kaplanmichael: Fixed.

CHANGELOG.md
kaplanmichael: Fixed.

kristojorg (approved):
looks good!

PR #222, opened 2020-11-30
Fix: Properly track open_book event

kristojorg:
Closes [SIMPLY-3327](https://jira.nypl.org/browse/SIMPLY-3327) We previously used a `PUT` request to track these, but the CM rejected those. We now send `GET` requests.


hokei:
the way we could have caught this in the past would be to literally go on bugsnag and verify, right?  

(Have you verified that all the other requests are being called properly?) 
(Out of curiosity, why would this have been the only post request we used?)

kristojorg:
Previously this wasn't tracking the error to bugsnag, just logging to the console... so we wouldn't have seen this error in bugsnag. This was a POST request because I thought that's what was needed from the CM to track an open_book event... but it turns out they actually want a GET request. Not idiomatic REST, but whatever that's the way it is for now. 

I didn't specifically verify any other tracking events because this is the only one we sent this way (for now).

kaplanmichael (approved):
LGTM

PR #227, opened 2020-12-04
Simply 3311 image optimization new

kaplanmichael:
https://jira.nypl.org/browse/SIMPLY-3311

Hi, team. I took an initial crack at this.  Let me know if this is kind of what you had in mind.  Thx!

### DIFF
--- a/src/components/BookCover.tsx
+++ b/src/components/BookCover.tsx
@@ -4,6 +4,7 @@ import * as React from "react";
 import { AspectRatio } from "@theme-ui/components";
 import { MediumIcon } from "./MediumIndicator";
 import { AnyBook } from "interfaces";
+import LazyLoadImage from "components/LazyLoadImage";
 
 /**
  * This is meant to be a book cover. Primarily the image and styling,
@@ -51,23 +52,23 @@ const BookCover: React.FC<{
       >
         <MediumIcon book={book} sx={{ height: "30%", fill: "ui.gray.dark" }} />
       </AspectRatio>
-      <img
-        alt={`Cover of book: ${book.title}`}
-        src={imageUrl}
-        onError={handleError}
-        onLoad={handleLoad}
-        sx={{
-          width: "100%",
-          height: "100%",
-          objectFit: "contain",
-          position: "absolute",
-          top: 0,
-          left: 0,
-          right: 0,
-          bottom: 0,
-          opacity: state === "success" ? 1 : 0,
-          transition: "all 0.1s ease-in"
-        }}
+      <LazyLoadImage
+          alt={`Cover of book: ${book.title}`}
+          src={imageUrl}
+          onError={handleError}
+          onLoad={handleLoad}
+          sx={{
+            width: "100%",
+            height: "100%",
+            objectFit: "contain",
+            position: "absolute",
+            top: 0,
+            left: 0,
+            right: 0,
+            bottom: 0,
+            opacity: state === "success" ? 1 : 0,
+            transition: "all 0.1s ease-in"
+          }}      
       />
       {showMedium && (
         <MediumIcon
--- a/src/components/BookCover.tsx
+++ b/src/components/BookCover.tsx
@@ -23,7 +23,6 @@ const BookCover: React.FC<{
 
   const handleError = () => setState("error");
   const handleLoad = () => setState("success");
-  const coverImageref = React.useRef<HTMLImageElement | null>(null);
 
   return (
     <div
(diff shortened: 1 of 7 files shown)


src/components/LazyLoadImage.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { ref, inView } = useInView({
+    threshold: 0,
+    triggerOnce: true
+  });
+  const { src, alt, ...otherProps } = LLImgProps;
+  return (
+    <img
+      ref={ref}
+      src={inView ? src : undefined}
+      data-src={src}
```
kristojorg, line 25: Curious what this attribute does?

src/components/LazyLoadImage.tsx
```diff
@@ -0,0 +1,31 @@
...
+interface LLImgProps extends React.ImgHTMLAttributes<HTMLImageElement> {}
+
+/**
+ * Renders an image tag with the provided props and adds an intersection observer to handle
+ * lazy-loading of the image.  The image tag will have no "src" attribute until at least one
+ * pixel of the image becomes visible in the viewport.
+ * @param LLImgProps - extends from the default img tag props so that we can eventually add the ability for the developer to pass in a custom root element
+ */
+
+const LazyLoadImage: React.FC<LLImgProps> = LLImgProps => {
```
kristojorg, line 15: Can you rename the props parameter so that it is different from the interface name? I think it's fine to just name it `props` or you can even destructure it right there if you'd like:

```ts
const LazyLoadImage: React.FC<LLImgProps> = ({ src, alt, ...otherProps }) => { ... }
```

src/components/LazyLoadImage.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { jsx } from "theme-ui";
+import * as React from "react";
+import { useInView } from "react-intersection-observer";
+
+interface LLImgProps extends React.ImgHTMLAttributes<HTMLImageElement> {}
+
+/**
+ * Renders an image tag with the provided props and adds an intersection observer to handle
+ * lazy-loading of the image.  The image tag will have no "src" attribute until at least one
+ * pixel of the image becomes visible in the viewport.
```
kristojorg, line 11: I think we should change this so that the image begins loading when it is some distance away from the viewport. I'm not sure what the threshold should be exactly, but the idea is that the user rarely sees an image in the process of loading when they're scrolling because they're already loaded

src/components/BookCover.tsx
```diff
@@ -51,7 +52,7 @@ const BookCover: React.FC<{
       >
         <MediumIcon book={book} sx={{ height: "30%", fill: "ui.gray.dark" }} />
       </AspectRatio>
-      <img
+      <LazyLoadImage
```
kristojorg, line 55: nit: Personally I would name this `LazyImage` because otherwise it sounds like a verb, which doesn't seem right as a component name

kristojorg (changes requested):
Looks great! And I think is going to be very helpful! Just a few things

kristojorg:
I just tested it, and it looks like this one change reduces the amount of image data loaded on the catalog root page from 1.3MB to 250kB! That's huge!

kristojorg:
I would also put a test and mock `useInView` to make sure that when it returns true, the source is in fact added to the image (and conversely it is undefined otherwise).

kristojorg:
Total size of network requests decreased by half from ~3MB to ~1.5MB.

src/components/LazyLoadImage.tsx
kaplanmichael: Good catch.  That was a remnant of an earlier implementation I had tried and eventually changed.  I've removed that property now.

src/components/LazyLoadImage.tsx
kaplanmichael: Great idea! Done!

kaplanmichael:
> Total size of network requests decreased by half from ~3MB to ~1.5MB.

...And this is just the beginning [of ways we can optimize] !!

src/components/BookCover.tsx
kaplanmichael: Fixed!

src/components/LazyLoadImage.tsx
kaplanmichael: Yes that makes complete sense.  I've gone ahead and made this fix for when the user is scrolling vertically.  In order to also do this for horizontal scrolling, however, I would need to refactor a bit.  Let me know if it makes sense to do that here or in a different PR.

kristojorg: I think it's fine to leave the horizontal scrolling work to a separate ticket for now. We might actually just leave it and see if people like it as it before refactoring to make that change.

kristojorg (approved):
Looks good! We can merge once you fix the lint error 

PS. I set my editor (VSCode) to run format on save using eslint, which automatically fixes these prettier errors for me

src/components/LazyLoadImage.tsx
kaplanmichael: 👍

kaplanmichael:
> I would also put a test and mock `useInView` to make sure that when it returns true, the source is in fact added to the image (and conversely it is undefined otherwise).

Done!

src/components/__tests__/LazyImage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { render, screen } from "@testing-library/react";
+
+import LazyImage from "../LazyImage";
+import { useInView } from "react-intersection-observer";
+import { mockAllIsIntersecting } from "react-intersection-observer/test-utils";
+
+const HookComponent = ({ options }) => {
+  const [ref, inView] = useInView(options);
+  return <div ref={ref}>{inView.toString()}</div>;
+};
```
kristojorg, line 13: You'll need to type the parameters to this function, probably this will work:
```
const HookComponent = ({ options }): Parameters<typeof useInView> => { ... }
```

kristojorg (approved):
Tests look great! Just the one typefix and we're good!

PR #228, opened 2020-12-07
Open eBooks Custom Login Page

kristojorg:
This is a custom `LoginPicker` for Open eBooks that shows some documentation of the two available login methods (taken from the home page).


src/auth/LoginPicker.tsx
```diff
@@ -0,0 +1,129 @@
+import * as React from "react";
```
kristojorg, line 1: This file is simply extracted from the home page

src/auth/OpenEbooksLoginPicker.tsx
```diff
@@ -0,0 +1,97 @@
+/** @jsx jsx */
```
kristojorg, line 1: This file is extracted from `/components/OpenEbooksLanding.tsx`

src/auth/LoginPicker.tsx
```diff
@@ -0,0 +1,128 @@
...
+      >
+        {authMethods?.map(method => (
+          <option key={method.id} value={method.id}>
+            {method.description}
+          </option>
+        ))}
+      </Select>
+      <AuthButton method={selectedMethod} />
+    </div>
+  );
```
kaplanmichael, line 127: If I click one of the buttons, proceed to its login form, and then click the browser's back button, I see a version of the login page where the buttons do not appear properly and any subsequent activity will fail:
image

image

kaplanmichael (changes requested):

src/auth/LoginPicker.tsx
kristojorg: This is a symptom of the ip address circ manager error that is being resolved elsewhere

src/auth/LoginPicker.tsx
```diff
@@ -0,0 +1,128 @@
...
+ * Renders buttons that allow selecting between auth providers.
+ */
+const Buttons: React.FC<{
+  authMethods: AppAuthMethod[];
+}> = ({ authMethods }) => {
+  return (
+    <Stack direction="column" aria-label="Available authentication methods">
+      {authMethods.map(method => (
+        <AuthButton key={method.id} method={method} />
+      ))}
```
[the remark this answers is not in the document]
kristojorg: Okay so the buttons and the logo were actually meant to be left-aligned according to the design (at this width). So I've updated to do that and I think it looks better. I also fixed some other design stuff like you pointed out. Should be more in line with the design specs now

src/auth/LoginPicker.tsx
kaplanmichael: Can't wait! :)

src/auth/LoginPicker.tsx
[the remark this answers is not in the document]
kaplanmichael: Looks great now!

kaplanmichael (approved):

PR #229, opened 2020-12-07
Re-enable strict type checking

kristojorg:
This is a typescript option that was disabled when we began pulling in code from opds-web-client. This PR re-enables it to bring runtime safety guarantees for the future.


hokei (approved):

hokei:
Can would it help to add strict type checking to axisnow-access-control-web?

kristojorg:
@hokei we can, but it's not holding us back here at all

PR #231, opened 2020-12-07
Lazy React AXE

kristojorg:
When testing @kaplanmichael 's recent lazy image PR, I noticed we were still loading React-AXE in production... even though it wasn't being run. This was bloating our first load JS from 223kB to 351kB, an almost 60% increase! This happened because of how we checked for development. We need to keep it simple for webpack's tree shaking, and this change fixes it.


hokei:
Oh my god!  😬 
(PS I'm good with this but I'll leave it to Michael to approve)

src/pages/_app.tsx
```diff
@@ -1,8 +1,7 @@
 import * as React from "react";
 import { AppProps, NextWebVitalsMetric } from "next/app";
-import { IS_SERVER, IS_DEVELOPMENT, REACT_AXE } from "../utils/env";
+import { IS_SERVER, REACT_AXE } from "../utils/env";
```
kaplanmichael, line 3: Looks good.  I did notice that this package is now deprecated on NPM though.  Is there an alternative or an updated version we should be using?

kaplanmichael (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Add: Button to go back to auth selection from basic auth if multiple methods are present.
 - Fix: Refactor See More card to use the same sizing and aspect ratio of the book cover images.
 - Fix: Fix delay between when you reach the end of a lane and when the forward/back arrows turn gray (disabled).
+- Fix: Don't download react-axe in production. Shaves 60% off the first load bundle.
```
kaplanmichael, line 8: Seems to be a merge conflict here.

src/pages/_app.tsx
kristojorg: Yea good point, I believe it just moved to a different name but I will install the updated version

kristojorg:
Installed the updated version @kaplanmichael thanks for finding that

kaplanmichael (approved):

PR #233, opened 2020-12-15
Feature/simply 3152 breadcrumb update

kaplanmichael:
https://jira.nypl.org/browse/SIMPLY-3152


kristojorg:
Okay so I just pushed a couple fixes here. There were two problems I could identify:

### Mutating Props

In `BreadcrumbsBar.tsx` we (read _I_) mutated the value of the `breadcrumbs` prop by calling `.pop()` on it. This was a reference to the same value stored all the way up in the SWR cache of the network data. So it made it look like suddenly the server returned one less breadcrumb on the next render. Very strange behavior ensued. 

The safest thing to do would be to mark all of our internal data structures as `Readonly` in `interfaces.ts`. This will make typescript throw an error if we try to mutate one of those values. We may want to investigate doing that to avoid errors like this in the future, but it would be a pretty large typechange, so maybe left to another PR if we want to do that.

As a patch, I marked the `breadcrumbs` prop to `BreadcrumbBar` as `Readonly<LinkData[]>`, thus disallowing mutations specifically in that component. This properly caused an error when I called `breadcrumbs.pop()`, and so I created a local copy of the array that we can freely mutate without changing the original value. 

### Placement of `BreadcrumbsProvider`

So this one requires a bit more knowledge of next.js... the problem was that even after setting the breadcrumbs to something, when I navigated pages it would somehow get reset to an empty array. It turns out that nextjs unmounts and remounts in full-page transitions between urls. They've discussed changing this, but that's how it works right now. This means that the context was lost since it was unmounted and then remounted with a fresh empty array default value. 

To fix these sorts of problems, they provide `pages/_app.tsx`, which is a component that wraps every page and stays mounted throughout the app lifecycle. Thus I moved `BreadcrumbProvider` into there, and it started working (persisting the value across page transitions). 

### Other

I also made a couple changes in `BreadcrumbsContext`, fixing the types and providing a default empty array value for the breadcrumbs. 

You'll want to verify this works and write some tests for it

.gitignore
```diff
@@ -32,6 +32,7 @@ wallaby.js
 /config/
 /patron-web-configs/
 .vscode/
+src/config/
```
kristojorg, line 35: This shouldn't be gitignored because we need the contents of that folder. Place your config file in `/config` (outside of src) and it will be automatically ignored (line 32 above)

src/components/Collection.tsx
```diff
@@ -11,21 +11,34 @@ import computeBreadcrumbs from "computeBreadcrumbs";
...
   const hasBooks = collection?.books && collection.books.length > 0;
   const pageTitle = isLoading ? "" : title ?? collection?.title ?? "Collection";
 
-  const breadcrumbs = computeBreadcrumbs(collection);
+  //||||||||||||||||||||||||||||||||||||||||||||||||||||||
+  const breadcrumbItems = React.useMemo(() => computeBreadcrumbs(collection), [
+    collection
+  ]);
+
+  const { breadcrumbs, setBreadcrumbs } = useBreadcrumbContext();
```
kristojorg, line 32: I would try to think of new names for these things so future readers can understand the difference... maybe `collectionBreadcrumbs` and `savedBreadcrumbs` and `updatedSavedBreadcrumbs` or something like that

.gitignore
kaplanmichael: Fixed.

src/components/Collection.tsx
kaplanmichael: Good idea; done!

kristojorg:
To fix the windows/mac env difference: 

1. Notice that in `jest.config.js` there is a `setupFiles` key and a `setupFilesAfterEnv` key. The first is a path to a file to set env vars before running the test suite. There is a file in there we already use to mock the `APP_CONFIG` variable. 
2. In that file (`src/test-utils/set-env.ts`), you can define your env var: `process.env.TZ = "UTC"`

That means it doesn't need to be set at the command line when running the tests. Let me know if that works

kristojorg:
The two failing tests are both valid:

1. The structure of the breadcrumbs seems to have changed, so it's not able to find one with the right aria label.
2. The other one where it gets two calls to `mocedSWR` is more complicated. That's happening because by setting state when the collection changed, the component rerenders and we now have two calls to `mockedSWR`, one for each render. I think in this case it's fine actually to remove the expectation on the number of calls. Just so long as it gets called with the right arguments is all we really care about in that instance.

kaplanmichael:
> To fix the windows/mac env difference:
> 
> 1. Notice that in `jest.config.js` there is a `setupFiles` key and a `setupFilesAfterEnv` key. The first is a path to a file to set env vars before running the test suite. There is a file in there we already use to mock the `APP_CONFIG` variable.
> 2. In that file (`src/test-utils/set-env.ts`), you can define your env var: `process.env.TZ = "UTC"`
> 
> That means it doesn't need to be set at the command line when running the tests. Let me know if that works

I actually tried that last week but it didn't work :(

kaplanmichael:
> > To fix the windows/mac env difference:
> > 
> > 1. Notice that in `jest.config.js` there is a `setupFiles` key and a `setupFilesAfterEnv` key. The first is a path to a file to set env vars before running the test suite. There is a file in there we already use to mock the `APP_CONFIG` variable.
> > 2. In that file (`src/test-utils/set-env.ts`), you can define your env var: `process.env.TZ = "UTC"`
> > 
> > That means it doesn't need to be set at the command line when running the tests. Let me know if that works
> 
> I actually tried that last week but it didn't work :(

The solution we ended up going with makes the above obsolete, so this is resolved.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+- Refactor: Include browsing history in breadcrumbs
```
kristojorg, line 4: nit: This should be an `Add` I think since it's a new feature

src/components/Collection.tsx
```diff
@@ -11,21 +11,32 @@ import computeBreadcrumbs from "computeBreadcrumbs";
...
 import ErrorComponent from "components/Error";
+import useBreadcrumbContext from "../components/context/BreadcrumbContext";
 
 export const Collection: React.FC<{
   title?: string;
 }> = ({ title }) => {
   const { collection, collectionUrl, isValidating, error } = useCollection();
 
-  if (error) return <ErrorComponent info={error?.info} />;
-
```
kristojorg, line 21: Was this something you did or I did? I can't remember the context of why here if I did it haha

kristojorg (changes requested):
Looks good to me except that I feel like we should add a test for this functionality... maybe a test that when you render a collection it sets the breadcrumbs context, and a test that when there is something in the context it gets used by the book details page

kristojorg:
Hey any updates on this guy?

CHANGELOG.md
kaplanmichael: Fixed.

CHANGELOG.md
```diff
@@ -15,23 +18,28 @@
 - Fix: Properly track AxisNow decryption errors to Bugsnag.
 
 ### 4.1.0
+
```
kristojorg, line 21: Do you have some kind of formatter turned on for markdown files?

kaplanmichael: Not intentionally.  I have "Format on Save" active per your instructions from a few weeks ago, but I didn't see a way to specify which _types_ of files get caught up in that...

kristojorg (approved):
LGTM!

PR #234, opened 2020-12-16
Release 4.1.0 to production

kristojorg:
This PR releases version 4.1.0 to production. It will immediately take effect on beta.openebooks.us.

# Changes

1. New dedicated login page. No more login modal.
2. New Open eBooks home page (will not affect deployments that don't have the `openebooks` key in their config file).


kaplanmichael (approved):
Looks good!

PR #236, opened 2020-12-16
4.2.0 to qa

kristojorg:
This releases `v4.2.0` to QA.

### 4.2.0

- Fix: The `open_book` event is now tracked by sending a `GET` request to the CM url instead of a `PUT` request, which was not accepted.
- Add: Button to go back to auth selection from basic auth if multiple methods are present.
- Fix: Refactor See More card to use the same sizing and aspect ratio of the book cover images.
- Fix: Fix delay between when you reach the end of a lane and when the forward/back arrows turn gray (disabled).
- Add: Custom Open eBooks Login Picker page to show helpful documentation when choosing a login method.
- Fix: Don't download react-axe in production. Shaves 60% off the first load bundle.
- Add: Lazy-loading of book cover images: A book cover image source will now only download once at least one pixel of the image's placeholder becomes visible in the viewport.
- Refactor: Re-enabled Typescript's strict type checking to provide better runtime type safety guarantees, and keep type safety from deteriorating hrough development.
- Fix: Properly track AxisNow decryption errors to Bugsnag.


kaplanmichael (approved):
LGTM

PR #237, opened 2020-12-16
Readme updates

kristojorg:
I am just updating the readme and removing the `community-config.yml` in favor of the `sample-config.yml`. The community config will be stored elsewhere so it can be updated without having to be tied to the release cycle of this repo.


kristojorg:
Closing in favor of #238

PR #239, opened 2020-12-22
[SIMPLY-3318] Add catalog button on openE home page.

kaplanmichael:
https://jira.nypl.org/browse/SIMPLY-3318

Figma:  https://www.figma.com/file/vcWnIIbpNtZE7M5n2lfegC/Open-E-Books?node-id=733%3A3296

**Note: See JIRA story comments for clarification about design.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ### UNRELEASED CHANGES
 - Add: Add catalog button on openE home page
-- Add: "theme" prop to SignOut component so that its color scheme can be configured at page level
 
 ### 4.2.0
 - Fix: The `open_book` event is now tracked by sending a `GET` request to the CM url instead of a `PUT` request, which was not accepted.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+- Add: Add catalog button on openE home page
+- Add: "theme" prop to SignOut component so that its color scheme can be configured at page level
 
 ### 4.2.0
 - Fix: The `open_book` event is now tracked by sending a `GET` request to the CM url instead of a `PUT` request, which was not accepted.
--- a/src/components/OpenEbooksLanding.tsx
+++ b/src/components/OpenEbooksLanding.tsx
@@ -247,7 +247,6 @@ const OpenEbooksHero: React.FC = () => {
                 href={{
                   pathname: `/${APP_CONFIG.openebooks?.defaultLibrary}`
                 }}
-                title="Catalog"
                 sx={{ mr: 1 }}
               >
                 Catalog
--- a/src/components/OpenEbooksLanding.tsx
+++ b/src/components/OpenEbooksLanding.tsx
@@ -6,7 +6,7 @@ import Button, { NavButton, AnchorButton } from "./Button";
 import Stack from "./Stack";
 import useLibraryContext from "./context/LibraryContext";
 import useUser from "./context/UserContext";
-import SignOut from "./SignOut";
+import { SignOut } from "./SignOut";
 import SvgChevronRight from "icons/ExpandMore";
 import { GetStaticProps, NextPage } from "next";
 import withAppProps, { AppProps } from "dataflow/withAppProps";
@@ -250,7 +250,7 @@ const OpenEbooksHero: React.FC = () => {
               >
                 Catalog
               </AnchorButton>
-              <SignOut />
+              <SignOut theme="light" />
             </>
           ) : (
             <NavButton
--- a/src/components/OpenEbooksLanding.tsx
+++ b/src/components/OpenEbooksLanding.tsx
@@ -6,7 +6,7 @@ import Button, { NavButton } from "./Button";
 import Stack from "./Stack";
 import useLibraryContext from "./context/LibraryContext";
 import useUser from "./context/UserContext";
-import { SignOut } from "./SignOut";
+import SignOut from "./SignOut";
 import SvgChevronRight from "icons/ExpandMore";
 import { GetStaticProps, NextPage } from "next";
 import withAppProps, { AppProps } from "dataflow/withAppProps";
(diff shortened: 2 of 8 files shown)


kaplanmichael:
I tried adding a test to verify the functionality of the theme prop, but had no luck.  If I'm not mistaken, it seems related to us using CSS variables for our colors.  Let me know if you can think of a way to achieve this.

src/components/SignOut.tsx
```diff
@@ -7,7 +7,13 @@ import Button from "./Button";
...
 import useUser from "components/context/UserContext";
 
-export default function SignOut() {
+interface SignOutProps {
+  theme?: "light" | "dark";
+}
+
+export const SignOut: React.FC<SignOutProps> = ({
+  theme = "dark"
+}: SignOutProps) => {
```
kristojorg, line 16: If we're going to do this, I would rather add it as a color to the `Button` component, as in make the `Button` set it's own `color` to white when it's background is black. Can be done in `components/button/styles` I think

src/components/OpenEbooksLanding.tsx
```diff
@@ -240,7 +240,18 @@ const OpenEbooksHero: React.FC = () => {
       >
         <div sx={{ display: "flex", margin: 3, justifyContent: "flex-end" }}>
           {isAuthenticated ? (
-            <SignOut />
+            <>
+              <AnchorButton
```
kristojorg, line 244: The style of this button is a little off on hover. I think similar to my suggestion with `SignOut` we should make the button handle a white `color` prop in `components/button/styles`

src/components/OpenEbooksLanding.tsx
```diff
@@ -240,7 +240,18 @@ const OpenEbooksHero: React.FC = () => {
...
+                variant="ghost"
+                color="ui.white"
+                href={libraryLinks.libraryWebsite?.href}
```
kristojorg, line 247: So this sort of works, but it is an absolute url to the production version of the app. I think it would be better to use a `NavButton` and then you link to "/app" by getting the default library from the app config: `href={APP_CONFIG.openebooks.defaultLibrary}`

src/components/OpenEbooksLanding.tsx
```diff
@@ -240,7 +240,18 @@ const OpenEbooksHero: React.FC = () => {
...
+                title="Catalog"
```
kristojorg, line 248: What does this prop do?

kristojorg (changes requested):
Looks good, just a couple changes

src/components/OpenEbooksLanding.tsx
kaplanmichael: Done!

src/components/OpenEbooksLanding.tsx
kaplanmichael: Removed!

src/components/SignOut.tsx
kaplanmichael: OK.  Have a look at the latest; is that more what you had in mind?

src/components/OpenEbooksLanding.tsx
kaplanmichael: OK. Have a look at the latest; is that more what you had in mind?

src/components/OpenEbooksLanding.tsx
```diff
@@ -240,7 +240,19 @@ const OpenEbooksHero: React.FC = () => {
...
         <div sx={{ display: "flex", margin: 3, justifyContent: "flex-end" }}>
           {isAuthenticated ? (
-            <SignOut />
+            <>
+              <NavButton
+                variant="ghost"
+                color="ui.white"
+                href={{
+                  pathname: `/${APP_CONFIG.openebooks?.defaultLibrary}`
+                }}
```
kristojorg, line 249: This should work, but I think you can also just say ``href={`/${APP_CONFIG.openebooks?.defaultLibrary`}`` without the need to make it a URL Object

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+- Add: Add catalog button on openE home page
+- Add: "theme" prop to SignOut component so that its color scheme can be configured at page level
```
kristojorg, line 5: Can remove this now

kristojorg (approved):
Looks good!

kristojorg:
Just need to fix that failing test

kristojorg:
It looks like you just need to update the snapshots. Run `npm run test` and then press `u` and then commit and push

kaplanmichael:
> It looks like you just need to update the snapshots. Run `npm run test` and then press `u` and then commit and push

Done!  (There was no 'u' option, so I ran it from the prompt instead:  "npm run jest --updateSnapshot".  )

CHANGELOG.md
kaplanmichael: Done!

src/components/OpenEbooksLanding.tsx
kaplanmichael: If I do that, it returns the following:   
image

kristojorg: Ohhhh good point that's because the `NavButton` is using our custom `Link` which is meant to handle routing _within_ a single library. So you can actually just set `href="/"` which is internally interpreted to be the home page of the library catalog. Maybe I should make this more clear now that we do actually have something at "/", whereas previously we didn't and only cared about the "/[library]" routes

kaplanmichael: Cool, that seemed to work. Should I also change the path of the "log in" button as well?  currently, in the code, it reads  `href={`/${slug}/login`}` but I could change it to ` href="/login"}`.

kristojorg: Yea good catch, I think that'll be clearer overall, thanks!

CHANGELOG.md
kristojorg: mmmm I'm still seeing it  : )

src/components/OpenEbooksLanding.tsx
kaplanmichael: Done!

PR #240, opened 2021-01-05
Fix invalid redirect URI

kristojorg:
This is an attempt to fix the invalid redirect uri we would sometimes send during clever auth. The issue was caused by the fact that sometimes we would redirect before the client finishes rendering, which means that Next.js's `asPath` would have a full url in it instead of just the _path_. I put in a couple guards for this case but have not been able to fully test it.


kaplanmichael (approved):
LGTM!

PR #241, opened 2021-01-06
Bugfix/simply 3331 wide viewport issue

kaplanmichael:
https://jira.nypl.org/browse/SIMPLY-3331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ### UNRELEASED CHANGES
 
+Fix: Improve the appearance of home page book thumbnails for larger viewports
+
 ### 4.2.0
+
 - Fix: The `open_book` event is now tracked by sending a `GET` request to the CM url instead of a `PUT` request, which was not accepted.
 - Add: Button to go back to auth selection from basic auth if multiple methods are present.
 - Fix: Refactor See More card to use the same sizing and aspect ratio of the book cover images.
@@ -14,23 +17,28 @@
 - Fix: Properly track AxisNow decryption errors to Bugsnag.
 
 ### 4.1.0
+
 - Add: Landing page for Open Ebooks
 - Fix: Fix our `APP_CONFIG` mocking in storybook so that `showMedium` and `companionApp` can be set via the toolbar once again.
-- Refactor: Made the login state a full page so that it can be navigated to directly. 
+- Refactor: Made the login state a full page so that it can be navigated to directly.
 
 ### 4.0.0
+
 - BREAKING CHANGE: Start app by fetching auth document first instead of catalog root. This means that the `libraries` key in the config file needs to be set with auth document urls instead of catalog root urls.
 - Feature: Support auth-protected collections.
 - Fix: Update webpub viewer.
 
 ### 3.1.3
+
 - Fix: Update the production docker publish Github Action to v2, allowing for use of private Github package registry for AxisNow Decryptor.
 - Fix: Book covers and titles in BookList now link to the book page
 
 ### v3.1.2
-- Bump to v3.1.2 in order to tricker Github Actions run. No changes, but renamed branches to `dev` `qa` and `production` from `beta` and `master`. 
+
+- Bump to v3.1.2 in order to tricker Github Actions run. No changes, but renamed branches to `dev` `qa` and `production` from `beta` and `master`.
 
 ### v3.1.1
+
 - Fix: Update github workflow files to properly publish new branches (production, qa, dev).
 
 ### v3.1.0
@@ -43,7 +51,6 @@
 - Fix: Deterministic `buildId` based on git commit SHA in order to allow running multiple instances of the application behind a load balancer.
 - Fix: Fetch the config file synchronously at startup so that the bugsnag api key can be used during the webpack setup.
 - Improvement: Application now tracks errors better by parsing them more precisely, and then recording them to bugsnag and console.
-
 
 ### v3.0.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,10 @@
 ### UNRELEASED CHANGES
 - Add: Add catalog button on openE home page
 
+Fix: Improve the appearance of home page book thumbnails for larger viewports
```
kristojorg, line 6: I think this needs a bullet point and shouldn't have a space between it and the preceding

src/components/OpenEbooksLanding.tsx
```diff
@@ -334,24 +334,19 @@ const PopularBookSection: React.FC<{
...
-                  flex: "[1, 0, 1]",
-                  mx: [0, 2],
-                  my: [2, 2, 0],
-                  paddingLeft: 1,
-                  backgroundColor: "brand.secondary",
-                  maxWidth: "30%",
-                  height: "100%"
+                  width: "100%",
+                  minWidth: "75px",
+                  flex: [1, 1, "auto"],
```
kristojorg, line 342: Is this meant to be `flex: 1 1 auto;` in the css or are you meaning to use these media queries?

kristojorg (changes requested):
I don't think this is quite right... 

This is the original:
image


And this is your new version:
image


I believe the books are being stretched right now when they shouldn't be

kristojorg:
I can't quite tell what the solution is meant to be but I think this isn't right as is

kristojorg:
I think it would make sense to give a max width to the whole section maybe so it just doesn't expand wide enough to look weird

CHANGELOG.md
kaplanmichael: Fixed!

src/components/OpenEbooksLanding.tsx
kaplanmichael: Rewrote that code block so this no longer applies.

kaplanmichael:
We determined that the issues described above were due to the way Safari's implementation of certain Flexbox properties differs from that of other browsers.  Should be OK now.

dev-server.js
```diff
@@ -6,18 +6,33 @@ const fs = require("fs");
...
 const app = next({ dev });
 const handle = app.getRequestHandler();
+const port = 3000;
 
 const httpsOptions = {
   key: fs.readFileSync("./localhost.key"),
   cert: fs.readFileSync("./localhost.crt")
 };
 
+var ipAddr = Object.values(require("os").networkInterfaces()).reduce(
```
kristojorg, line 16: A comment on what this is doing would be appreciated

src/components/OpenEbooksLanding.tsx
```diff
@@ -295,68 +285,61 @@ const PopularBookSection: React.FC<{
   coverLocation?: "left" | "right";
 }> = ({ children, books, coverLocation }) => {
   return (
-    <div
+    <section
+      className="popular-books-section"
```
kristojorg, line 289: What is the purpose of this className?

src/components/OpenEbooksLanding.tsx
```diff
@@ -295,68 +285,61 @@ const PopularBookSection: React.FC<{
...
       sx={{
-        m: 2,
+        px: "8.5%",
+        my: ["20px", "20px", "35px"],
```
kristojorg, line 292: I would rather use theme scale values than hardcoded pixel values if possible. But if those aren't specific enough then we can use this

src/components/OpenEbooksLanding.tsx
```diff
@@ -295,68 +285,61 @@ const PopularBookSection: React.FC<{
...
         display: "flex",
         flexDirection: [
           "column",
           "column",
           coverLocation === "right" ? "row" : "row-reverse"
-        ]
+        ],
+        ":first-of-type": {
+          marginTop: ["28px", "28px", "70px"]
```
kristojorg, line 300: Same here

kristojorg (commented):
Some questions, and I would rather not use hard coded pixel values unless it's necessary

kristojorg:
Otherwise this looks good to me.

src/components/OpenEbooksLanding.tsx
kaplanmichael: I felt that it makes the code slightly more readable, but I'm cool with zapping it if you'd prefer not having stuff like that in the code.  LMK.

kristojorg: I think we should preserve classNames as styling, but you can add a comment if you'd like! It does seem pretty readable to me simply because the component is already named

dev-server.js
kaplanmichael: Done!

src/components/OpenEbooksLanding.tsx
kaplanmichael: I was referring to the readability of the _rendered_ code, but you've sold me on stripping it out :)  Should be good now!

src/components/OpenEbooksLanding.tsx
kaplanmichael: Fixed!

src/components/OpenEbooksLanding.tsx
kaplanmichael: Fixed!

kristojorg (approved):
Okay this looks good to me!

kaplanmichael:
> Okay this looks good to me!

@kristojorg Awesome, thanks!  Should I go ahead and merge?  How should I route the JIRA ticket?

kristojorg:
Okay I squashed and merged (which puts all the commits from this PR into a single commit on the dev branch to keep things clean). You can move the JIRA ticket to ready for QA!

PR #244, opened 2021-02-03
[SIMPLY-3496] 401 page is flashed during login redirection in CPW

kaplanmichael:
Display a page loader instead of an error screen when a 401 is caught.

https://jira.nypl.org/browse/SIMPLY-3496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,16 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+
 - Documentation: update README.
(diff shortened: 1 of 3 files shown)


kristojorg (changes requested):
Would be great to add a simple test to Error.test.tsx for the 401 case. Otherwise this looks great and works perfectly.

kaplanmichael:
> Would be great to add a simple test to Error.test.tsx for the 401 case. Otherwise this looks great and works perfectly.

Done!

kristojorg:
Okay this looks good. I'll merge.

kristojorg (approved):

PR #246, opened 2021-02-16
[SIMPLY-3479] Stuck in "Logging in" state

kristojorg:
This ticket addresses [SIMPLY-3479](https://jira.nypl.org/browse/SIMPLY-3479). I now clear the browser auth cookie when the server responds with a 401 to the /loans route request. This should resolve the issue where users need to clear cache in order to avoid a "stuck in login" state.


kaplanmichael (approved):
LGTM!  Great to finally get this into the codebase!

PR #252, opened 2021-03-08
[SIMPLY-3466] Login Error

kristojorg:
There was an undefined error when logging in via clever sometimes. https://jira.nypl.org/browse/SIMPLY-3466 .

I'm not sure exactly what the underlying bug was, but it had to do with how we were removing the hash parameter from the url after logging in. In the end I decided to swap use of `window.history.replaceState` instead of `router.replace` from next.js because I saw that line in the stack trace of the error, which somehow was only happening in production. This seems to have resolved the issue.


kaplanmichael (approved):
Loox good.

PR #255, opened 2021-04-14
contributors

kristojorg:
This is a draft PR to resolve https://jira.nypl.org/browse/SIMPLY-3653 where contributors are not showing up right in list view or book details view.

### DIFF
--- a/src/components/BookList.tsx
+++ b/src/components/BookList.tsx
@@ -121,9 +121,6 @@ export const BookListItem: React.FC<{
   // uses contributors if there are no authors
   const authors = getAuthors(book);
 
-  if (authors[0] === undefined) {
-    console.log(book);
-  }
   return (
     <li
       sx={{
--- a/src/components/BookList.tsx
+++ b/src/components/BookList.tsx
@@ -117,6 +117,13 @@ export const BookListItem: React.FC<{
   // if the book exists in loans, use that version
   const loanedBook = loans?.find(loan => loan.id === collectionBook.id);
   const book = loanedBook ?? collectionBook;
+
+  // uses contributors if there are no authors
+  const authors = getAuthors(book);
+
+  if (authors[0] === undefined) {
+    console.log(book);
+  }
   return (
     <li
       sx={{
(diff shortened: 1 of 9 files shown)


src/components/BookList.tsx
```diff
@@ -110,6 +110,18 @@ export const BookList: React.FC<{
   );
 };
 
+function getAuthorsString(book: AnyBook): string {
+  const { authors } = book;
+  if (!authors) return "Unknown Author";
+  const authorsArray = getAuthors(book, 2);
+
+  if (authors.length > 2) {
+    authorsArray.push(`& ${authors?.length - 2} more`);
```
kaplanmichael, line 119: Does `authors` need the question mark here?  If there aren't any authors that line of code I think won't be executed in the first place.

src/components/__tests__/BookList.test.tsx
```diff
@@ -50,7 +50,7 @@ test("truncates authors", () => {
   );
   const utils = render(<BookList books={[longBook]} />);
 
-  expect(utils.getByText("one, two & 3 more"));
+  expect(utils.getByText("one, two, & 3 more"));
```
kaplanmichael, line 53: This Oxford comma just made my day.  This PR now gets 100 bonus points!

src/components/BookList.tsx
```diff
@@ -119,7 +131,7 @@ export const BookListItem: React.FC<{
   const book = loanedBook ?? collectionBook;
 
   // uses contributors if there are no authors
```
kaplanmichael, line 133: Is this comment accurate? Looking at `getAuthorsString`, I didn't see anything related to contributors (it seems like that's all handled in the `parse` method.

kaplanmichael (changes requested):
Looks good. Just a few quick questions:

src/components/BookList.tsx
kristojorg: Good point thanks

src/components/__tests__/BookList.test.tsx
kristojorg: Hahah it was a necessary fix

src/components/BookList.tsx
kristojorg: Yep it's accurate, it's withing the `getAuthors` method which is called inside `getAuthorsString`. I moved them next to each other in `utils/book.tsx` more obvious.

PR #261, opened 2021-07-08
[WIP] Cypress

hokei:


kristojorg:
I think that ultimately we probably only want to run this workflow on `deployment_status` events, since those _should_ happen with every push to a PR anyway, otherwise I think it will be run 3 times on every push. 

Also, one more thing. Maybe we should just apply for the OSS plan of this: https://www.cypress.io/dashboard which appears to do all of this more easily and also records the test runs so you can quickly see failures. Just an option

kristojorg:
I'd also like to note for the QA org that I have installed react-testing-library for cypress so they can use those queries to interact with the dom, which implicitly gives us more confidence about A11y

kristojorg:
Ok I added a single passing test case and it appears to be running 😎 ... now off to actually start vacation !

hokei:
yeah I put it on `push` to help us test the PR, but it looks OK now so I'll change it to `deployment_status` and merge.

hokei:
yeah I put it on `push` to help us test the PR, but it looks OK now so I'll change it to `deployment_status` and merge.

PR #262, opened 2021-07-27
Create First Cypress Tests (Jira OE-166)

jcsmit17:


package.json
```diff
@@ -21,7 +21,9 @@
     "lint:fix": "eslint \"*/**/*.{ts,tsx}\" --fix",
     "generate-icons": "npx @svgr/cli --template src/icons/_template.js --typescript --icon --ext tsx --svg-props 'fill=currentColor' -d src/icons src/icons",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "cypress:chrome": "./node_modules/.bin/cypress run --browser chrome cypress",
+    "cypress:headless": "./node_modules/.bin/cypress run cypress"
```
kristojorg, line 26: you don't need the `./node_modules/.bin/cypress` here, you can just write `cypress run ...`

cypress/integration/firstbook_spec.js
```diff
@@ -0,0 +1,31 @@
...
+    cy.get('#password').type('9864');
+    cy.get('.css-td3bz2').click();
+    cy.get('#search-bar', {timeout: universal_timeout}).should('be.visible');
+    cy.get('#search-bar').click();
+    cy.get('#search-bar').type('einstein');
+    cy.get('#__next > div > header > div > form > button', {timeout: universal_timeout}).click();
+    cy.get('#cpw-content > div > ul > li:nth-child(2) > div.css-5489ev > div > div:nth-child(1) > h2 > a', {timeout: universal_timeout}).click();
+    cy.get('.css-db5xrm:nth-child(2) .css-1ii5jqc', {timeout: universal_timeout}).click();
+    cy.get('#cpw-content > section > div.css-1pioiqw > div > div.css-865ptw > div.css-1r8p8yq > div > div.css-b7iuw7 > button', {timeout: universal_timeout}).click();
+})
```
kristojorg, line 18: We should not be performing a manual login before every test. We should test the login flow independently, and then for other tests just provide the auth token directly. [Here is an example](https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/logging-in__basic-auth/cypress/integration/spec.js) from the cypress website showing how to do that using basic auth, which would work for firstbook.

I highly suggest [this youtube video](https://www.youtube.com/watch?v=5XQOK0v_YRE) about why it's recommended to do it this way.

cypress/integration/clever_spec.js
```diff
@@ -0,0 +1,36 @@
...
+        cy.get('#react-autowhatever-1--item-0 > div > div > div', {timeout: universal_timeout}).click();
+        cy.get('#username', {timeout: universal_timeout}).type('44628');
+        cy.get('#password', {timeout: universal_timeout}).type('44628');
+        cy.get('.UsernamePasswordForm--loginButton', {timeout: universal_timeout}).click();
+        cy.get('#search-bar', {timeout: universal_timeout}).type('einstein');
+        cy.get('.css-evf56', {timeout: universal_timeout}).click();
+        cy.get('.css-1ii5jqc', {timeout: universal_timeout}).click();
+        cy.get('.css-1omctl1:nth-child(1)', {timeout: universal_timeout}).click();
+        // Verify behavior
+        cy.get('#viewer > main > div.page-container', {timeout: universal_timeout}).should('be.visible');
```
kristojorg, line 21: What does this verify?

kristojorg (changes requested):
We will need to change all queries from queries based on dom-structure, class name, id, or other non user-facing elements to ones that query elements the way users would: by their text content or aria-label or otherwise. We can do this using [cypress-testing-library](https://testing-library.com/docs/cypress-testing-library/intro/). I have already instantiated it in the cypress setup and used it in my sample test I believe. The reasons are as follows:

- Class names are generated at build type via a hash function, thus will change anytime any style changes. This would make tests fail almost every build.
- DOM structure and class names or IDs are all likely to change as the app continues to develop, even if the underlying functionality stays the same. The test should pass if the functionality is the same, even if we add new dom markup or move things around. 
- Most importantly, the presence of a certain class names or ids or dom children is not really what we care about. We care that there is a button called "login" on the page, and a field labeled "username" that can me typed into, etc. Further, we care that these things are accessible to people using screen readers, which means that the elements should have the apropriate aria-label, etc. Querying the dom this way is what cypress-testing-library helps you do. 

Writing the tests using this strategy should both be easier (you don't have to dig through for class names and dom structure etc), less flaky, and better verify the functionality we care about

cypress/integration/firstbook_spec.js
kristojorg: This is another [example using JWT](https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/logging-in__jwt/cypress/integration/spec.js), which is actually more like what we would need to do, since the page itself is not protected with basic auth, just the server endpoints. We need to save the auth token into the localStorage the same way the app does it before the tests, that way it is there when the app looks for it and we can proceed as logged in. 

Performing a manual login on every test will add complexity to every test, potential for flakiness, and will slow them down significantly. Let's try using the localStorage way. I would suggest loggin in manually and looking at `localStorage` to see the format it is stored in in, and then you can perform a login programmatically and get an auth token to save there in `beforeEach`

cypress/integration/clever_spec.js
```diff
@@ -0,0 +1,36 @@
...
+
+beforeEach(() => {
+    cy.visit('https://qa-web.openebooks.us/');
+    cy.get('.css-1cnys34').click();
+})
+
+describe('Clever', () => {
+    it('opens book in e-reader', () => {
+        // Log in for the first time
+        cy.get('.Autosuggest--textInput', {timeout: universal_timeout}).type('573a4487a920880100000aba');
```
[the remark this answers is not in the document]
jcsmit17: I may be able to trim it down or provide a cleaner way to wait for page loading, but this is the rough version to ensure the page is loaded before looking for elements.

cypress/integration/clever_spec.js
jcsmit17: This is verifying that e-reader is present in browser.

cypress/integration/clever_spec.js
```diff
@@ -0,0 +1,36 @@
...
+        cy.get('#search-bar', {timeout: universal_timeout}).type('einstein');
+        cy.get('.css-evf56', {timeout: universal_timeout}).click();
+        cy.get('.css-1ii5jqc', {timeout: universal_timeout}).click();
+        cy.get('.css-1omctl1:nth-child(1)', {timeout: universal_timeout}).click();
+        // Verify behavior
+        cy.get('.next > .flip-page-icon', {timeout: universal_timeout}).click();
+        cy.wait(30000);
+        cy.get('.prev > .flip-page-icon', {timeout: universal_timeout}).click();
+        cy.wait(30000);
+        cy.get('#viewer > main > div.page-container', {timeout: universal_timeout}).should('be.visible');
```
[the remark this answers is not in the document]
jcsmit17: This is verifying that the e-reader without error after flipping pages.

.gitignore
```diff
@@ -35,4 +35,10 @@ wallaby.js
 .vercel
 
 # storybook
-storybook-static
\ No newline at end of file
+storybook-static
+
+# cypress testing
+cypress/plugins
```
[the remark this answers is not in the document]
jcsmit17: Good catch. I'll fix in next patch.

kristojorg:
Closing in favor of #264

PR #264, opened 2021-08-05
SFR-1239 & SFR-1241 Create Cypress Tests for browsing and checking age appropriateness

samanthaandrews:
- Cypress tests for [all access browsing](https://jira.nypl.org/browse/SFR-1239) & [checking age appropriateness](https://jira.nypl.org/browse/SFR-1241) in CPW.
- Also includes test for the "Report a problem" modal dialog.
- Fixes a couple accessibility labels.
- Created a custom Cypress action for logging in programmatically before each test.

FYI:
* I created the following task (with subtasks) for making sure we get things running with GitHub Actions & GitHub Secrets: https://jira.nypl.org/browse/SFR-1253
* I still need to file a couple bug tickets for items I found. I'll link them here later this evening.
* Storybook is still failing (but we knew that). I'll create a ticket for this too, unless you've already filed one @hokei ?


cypress.json
```diff
@@ -1,3 +1,13 @@
...
+  "baseUrl": "http://localhost:3000",
+  "env": {
+    "OPENEBOOKS_ALL_ACCESS_USER_USERNAME": "A3IAYP9NDS",
+    "OPENEBOOKS_ALL_ACCESS_USER_PW": "",
+    "OPENEBOOKS_HIGH_SCHOOL_USER_USERNAME": "HMN4IIRXX5",
+    "OPENEBOOKS_HIGH_SCHOOL_USER_PW": "",
+    "OPENEBOOKS_MIDDLE_GRADES_USER_USERNAME": "MX8GNTIL6I",
+    "OPENEBOOKS_MIDDLE_GRADES_USER_PW": "",
+    "OPENEBOOKS_EARLY_GRADES_USER_USERNAME": "EMNK6TDG4R",
+    "OPENEBOOKS_EARLY_GRADES_USER_PW": ""
```
samanthaandrews, line 11: These came from this [doc](https://docs.google.com/spreadsheets/d/1D0153QostVdk6YWMSZ4Vv_aBa2TxpKJ8vW7R8biiZoA/edit#gid=317403225) that @kristojorg sent, but it seems like we should have dedicated test accounts for these Cypress tests and make note that these will be running on a daily or hourly chron job?

cypress/integration/open-ebooks/age-appropriateness.ts
```diff
@@ -0,0 +1,369 @@
...
+    cy.visit(HIGH_SCHOOL_AUTHOR_RECOMMENDATIONS_PATH);
+    cy.findByText("404 Error: No such lane.").should("not.exist");
+    cy.visit(HIGH_SCHOOL_DETAIL_BOOK_PATH_3);
+    cy.findByText("404 Error: No such lane.").should("not.exist");
+
+    // Verify you cannot navigate to collections or books outside your age class
+    cy.visit(MIDDLE_GRADES_COMICS_COLLECTION_PATH);
+    cy.findByText("404 Error: No such lane.").should("exist");
+    // cy.visit(MIDDLE_GRADES_DETAIL_BOOK_PATH);
+    // cy.findByText("404 Error: No such lane.").should("exist");
```
samanthaandrews, line 159: You will see these commented out assertions throughout this file, and that is because I believe these are bugs that I found! Basically, if you try to navigate directly to a path that you should not have access to, you can still see the content. For example, if I'm logged in as an "Early Grades" user and I navigate to the URL for the book _Betrayed_ by Jennifer Rush, I can view that page and do not get a 404 page. FWIW, there is a 451 server error that displays in the console.

I will create a bug for this issue.
Screen Shot 2021-08-05 at 11 06 58 AM

cypress/integration/open-ebooks/report-problem.ts
```diff
@@ -0,0 +1,31 @@
...
+      .should("exist")
+      .click();
+    cy.findByRole("dialog", { name: "Report a problem" }).should("exist");
+    cy.findByRole("heading", { name: "Report a problem" }).should("exist");
+    cy.findByRole("button", { name: "Cancel" }).should("exist");
+    cy.findByRole("option", { name: "Cannot Fulfill Loan" }).should("exist");
+    cy.findByRole("textbox", { name: "Details" })
+      .should("exist")
+      .type("frontend cypress test");
+    cy.findByRole("button", { name: "Submit" }).should("exist").click();
```
samanthaandrews, line 22: I am actually submitting this data, even though it's a test. I'm not sure who these reports are sent to, but perhaps this is annoying for them??

cypress/plugins/index.ts
```diff
@@ -11,12 +11,20 @@
 
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
+const injectDevServer = require("@cypress/react/plugins/next");
```
samanthaandrews, line 14: I thought that this would resolve issues with importing the `generateToken()` function into our tests, but it didn't, so I'm thinking I should just remove it and its associated packages.

cypress/support/commands.ts
```diff
@@ -0,0 +1,75 @@
...
+type UserType =
+  | "ALL_ACCESS_USER"
+  | "HIGH_SCHOOL_USER"
+  | "MIDDLE_GRADES_USER"
+  | "EARLY_GRADES_USER";
+
+function generateToken(username: string, password: string) {
+  const btoaStr = btoa(`${username}:${password}`);
+  return `Basic ${btoaStr}`;
+}
```
samanthaandrews, line 51: This is the function that I'm unable to import into any Cypress file without ultimately receiving this error. I am not really sure why Cypress needs access to this config only when importing an external function??
Screen Shot 2021-08-03 at 12 38 37 PM

cypress/support/utils.ts
```diff
@@ -0,0 +1,29 @@
+export const BASIC_LOGIN = "http://opds-spec.org/auth/basic";
```
samanthaandrews, line 1: I put all of the "magic" strings in one file. Hoping this would make the tests easier to change if we needed to do so.

src/interfaces.ts
```diff
@@ -5,7 +5,8 @@
  * Currently only used for support of a Library Registry, which is
  * an OPDS 2 Feed of OPDS 2 Catalogs from which we extract the catalog root url
  */
-export * as OPDS2 from "types/opds2";
+import * as OPDS2 from "types/opds2";
+export { OPDS2 };
```
samanthaandrews, line 9: Needed to do this to fix the errors when importing the `generateToken()` function in Cypress, which then led me to the missing `process.env.APP_CONFIG` error above 🙄

package.json
```diff
@@ -119,7 +122,9 @@
     "react-is": "^16.13.1",
     "react-test-renderer": "^16.13.1",
     "typescript": "^4.3.4",
-    "webpack-bugsnag-plugins": "^1.4.3"
+    "webpack": "^4.44.1",
+    "webpack-bugsnag-plugins": "^1.4.3",
+    "webpack-dev-server": "^3.11.2"
```
samanthaandrews, line 127: These were the changes made to attempt to fix the import of `generateToken()` as recommended by the Cypress docs for ["Component Testing"](https://docs.cypress.io/guides/component-testing/framework-configuration) (but we aren't even doing component testing in Cypress). However, I get the same error with or without these additional packages which is why I'm thinking I should revert these changes.

samanthaandrews: Removed.

cypress/plugins/index.ts
samanthaandrews: Removed.

cypress.json
hokei: yes.  I think we can make a note in the google doc that this is what they're being used for, and make sure Miguel is aware of this.

cypress/integration/open-ebooks/age-appropriateness.ts
```diff
@@ -0,0 +1,369 @@
...
+  it("can view books and collections for high school grades", () => {
+    cy.visit(APP_PATH);
+
+    // Verify collection title
+    cy.findByRole("heading", { name: "Open eBooks (QA Server)" }).should(
+      "exist"
+    );
+
+    // Verify all access lanes are not present
+    cy.findByRole("heading", { name: "High School collection" }).should(
```
hokei, line 81: wait isn't this a high school user?  Why would "high school collection" not exist?

cypress/integration/open-ebooks/age-appropriateness.ts
```diff
@@ -0,0 +1,369 @@
...
+      "exist"
+    );
+    cy.findByRole("heading", { name: "Romance collection" }).should("exist");
+    cy.findByRole("heading", { name: "Fiction collection" }).should("exist");
+    cy.findByRole("heading", { name: "Nonfiction collection" }).should("exist");
+    cy.findByRole("heading", { name: "All High School collection" }).should(
+      "exist"
+    );
+
+    // Verify early and middle grade lanes are not present
```
hokei, line 118: Somehow I was under the impression that high school students can view early/middle grade books, but not vice versa.  
Can we verify this?

cypress/integration/open-ebooks/browse.ts
```diff
@@ -0,0 +1,133 @@
...
+    // Find a book to click on
+    cy.findByRole("listitem", { name: "Book: Heart of a Champion" })
+      .should("exist")
+      .within(() => {
+        cy.findByRole("heading", { name: "Heart of a Champion" }).should(
+          "exist"
+        );
+        cy.findByText("Deuker, Carl").should("exist");
+        cy.findByText("Available to borrow").should("exist");
+        cy.findByText("9999 out of 9999 copies available.").should("exist");
```
hokei, line 58: This I think might become frustratingly brittle if this book is used as a test from anyone else on QA.

cypress/integration/open-ebooks/report-problem.ts
hokei: you should ask in the #open-ebooks slack, I think Miguel should know (or should find the answer)  

I also think if there's a way to call a mock in this submit I'd prefer to do that?  You know, to prevent an [HBO Max ](https://twitter.com/hbomaxhelp/status/1405712235108917249)kind of problem

src/interfaces.ts
hokei: this is weird because AFAIK CPW doesn't support a library registry but ... I think we can deal with this some other time

hokei:
What's here looks good, but I'm concerned about the fact that it's all using live data.  Right now, also, beta.openebooks.us is literally pointing to the qa.circulation so it's using live _production_ data.  I'm not sure when we're going to get a real QA instance, and I'm not sure if the data on this QA instance would be guaranteed, if we got it.  

Is there a way we can run these extensive tests using mocks, and then run a simpler set of "are the services up" tests using live data?  Or is there some reading I can do on why we don't want to mock the OPDS2 response?

cypress.json
kristojorg: Are we passing the passwords in via env vars? If so I think we should also pass in the username via env var, just so the two are set in the same place and easier to update in the future

cypress/integration/open-ebooks/age-appropriateness.ts
kristojorg: Yep this is a good find! Here is the deal- the circulation manager (backend) was not really built initially for this sort of authentication and user-role scheme, so the implementation we have is a bit naive. I could be wrong, but I think that individual books will always be "unprotected" which means they're public essentially and can be viewed by anyone, logged in or not, and no matter of age - if they navigate directly to them. So it may be that this is intended behavior, and we should actually have a test that it is true and add a comment as to why.

However, that said, I haven't seen this console error before, so maybe it's new and we should actually be blocking logged in users from viewing these books? 

@leonardr can you give any insight into the protections for an individual OPDS entry?

cypress/integration/open-ebooks/report-problem.ts
kristojorg: Lol yep we should at least ask, and possible just mock out that endpoint I agree

cypress/support/commands.ts
kristojorg: I think this is because somewhere along the chain of that import we have some code that uses `process.env.APP_CONFIG`, which I set up for us in the actual app via next.config.js. I think we'll probably need to set it to some mock value for the cypress runtime as well.

cypress/support/utils.ts
kristojorg: This _should_ exist as a constant in the app somewhere already... but I guess we still aren't able to import the app code huh... anyway once we fix that we should see if we can import this string

src/interfaces.ts
kristojorg: It does support a library registry actually. The "community demo" uses the NYPL library registry to deploy an instance for every library in the registry

cypress/integration/open-ebooks/age-appropriateness.ts
kristojorg: That's not the case I don't believe, but agree we should verify. My understanding is that each age group can see only their age group.

kristojorg:
@hokei - We do have a qa instance, and `qa-web.openebooks.us` is pointing to the qa circulation manager.

I also think it is good that this is using live data, and I think it should. That's why they're end to end tests as far as I understand. 

At some point, I wouldn't be opposed to having the same set of tests run against both live and mocked endpoints, but I think the starting point should be using live qa server, because we want to know if every _actually_ works from end to end. If something changes on the backend and that breaks our app, we want to know before someone randomly finds/reports it.

cypress/integration/open-ebooks/age-appropriateness.ts
```diff
@@ -0,0 +1,369 @@
...
+    cy.findByRole("heading", { name: "Middle Grades collection" }).should(
+      "exist"
+    );
+    cy.findByRole("heading", { name: "Early Grades collection" }).should(
+      "exist"
+    );
+
+    // Verify you can visit all 3 lanes
+    cy.visit(HIGH_SCHOOL_COLLECTION_PATH);
+    cy.findByText("404 Error: No such lane.").should("not.exist");
```
kristojorg, line 45: I think I would generally try to confirm you can visit by making assertions in the affirmative. If we happen to change this error message these tests could pass erroneously.

cypress/integration/open-ebooks/age-appropriateness.ts
kristojorg: They don't see the heading, it just shows them their sublanes. You only see the sub-lane headings. In the all-access case, the "high school collection" is a sub lane itself. Does that make sense? I think we could add a comment to clarify this though

cypress/integration/open-ebooks/browse.ts
kristojorg: That's a good point. Maybe we could just test that "out of 9999 copies available" is there

cypress/integration/open-ebooks/report-problem.ts
```diff
@@ -0,0 +1,31 @@
...
+    cy.loginByApi("ALL_ACCESS_USER");
+  });
+
+  it("submits an issue through the modal dialog", () => {
+    cy.visit(HIGH_SCHOOL_DETAIL_BOOK_PATH_1);
+    cy.findByRole("button", { name: "Report a problem" })
+      .should("exist")
+      .click();
+    cy.findByRole("dialog", { name: "Report a problem" }).should("exist");
+    cy.findByRole("heading", { name: "Report a problem" }).should("exist");
```
kristojorg, line 16: thought: do these exist before you click the button? I can't remember if this HTML is always rendered, but only visible after clicking button, or if it's actually not there until you click. Maybe you can verify?

kristojorg (changes requested):
This looks great, super thorough! I'm hoping that the qa-server collection is somewhat stable and we don't need to go updating the books and stuff in these tests often. We'll see though. If not, then we should talk to leonard about having some standard collections and books that stay the same for testing purposes added to the qa server. 

Just a couple comments

kristojorg:
I can merge this without storybook working, and we can make storybook work next. However, I do think we should try to make the e2e actions work before merging maybe. Open to merging into dev first though and doing that in another PR.

cypress/integration/open-ebooks/age-appropriateness.ts
hokei: Oh yes, this makes sense, but I think a comment would alleviate some confusion

hokei:
> At some point, I wouldn't be opposed to having the same set of tests run against both live and mocked endpoints, but I think the starting point should be using live qa server, because we want to know if every actually works from end to end. If something changes on the backend and that breaks our app, we want to know before someone randomly finds/reports it.

 I just think, if we're generating reports and stuff based on this, I don't want to be on the hook for server errors, or, even worse, just data getting updated.  This is a bigger conversation and I think we will continue it in a bunch of different venues - for now, this is fine.  
 
 It'll also be useful to get it out and seeing whether data actually does end up getting updated frequently 
When Leonard comes back we could also see about getting test users set up, with test lanes and collections, so that the data is stable.

cypress/integration/open-ebooks/report-problem.ts
samanthaandrews: Good point. I will check on that.

cypress/integration/open-ebooks/browse.ts
samanthaandrews: You're right. I think this might be a good candidate for using mocked data instead of live.

cypress/integration/open-ebooks/age-appropriateness.ts
samanthaandrews: I'll check with Miguel now

cypress/integration/open-ebooks/age-appropriateness.ts
samanthaandrews: These lanes are "All access" lanes that are only visible to an "All access" user, so the separation-by-grade lanes will not be present. I will add a comment to clarify.

samanthaandrews:
> I can merge this without storybook working, and we can make storybook work next. However, I do think we should try to make the e2e actions work before merging maybe. Open to merging into dev first though and doing that in another PR.

@kristojorg @hokei My plan after reading through your comments (thanks, btw!) is to follow the Cypress docs recommendation and create mock data for these tests, while still having one e2e test that uses real data and runs through the app as a real user would.

I'd prefer to merge this PR into dev after making that change, in addition to the other changes requested. And then I will get to work on  https://jira.nypl.org/browse/SFR-1253, which includes several sub-tasks with getting Cypress configured to work with GitHub Actions & GitHub Secrets. Specifically, I'll work on https://jira.nypl.org/browse/SFR-1254 and https://jira.nypl.org/browse/SFR-1257. Feels like getting this config set up and running how we want it is a separate ticket.

cypress/integration/open-ebooks/age-appropriateness.ts
samanthaandrews: Miguel confirmed the behavior that @kristojorg described, each age group can see only their age group.

cypress.json
samanthaandrews: This is the doc where I found the test account info: https://docs.google.com/spreadsheets/d/1D0153QostVdk6YWMSZ4Vv_aBa2TxpKJ8vW7R8biiZoA/edit#gid=317403225

I requested edit access last week, but have not yet received it. Anyone able to access it and mark the accounts we're currently using?

cypress.json
```diff
@@ -1,3 +1,13 @@
...
+  "baseUrl": "http://localhost:3000",
+  "env": {
+    "OPENEBOOKS_ALL_ACCESS_USER_USERNAME": "",
+    "OPENEBOOKS_ALL_ACCESS_USER_PW": "",
+    "OPENEBOOKS_HIGH_SCHOOL_USER_USERNAME": "",
+    "OPENEBOOKS_HIGH_SCHOOL_USER_PW": "",
+    "OPENEBOOKS_MIDDLE_GRADES_USER_USERNAME": "",
+    "OPENEBOOKS_MIDDLE_GRADES_USER_PW": "",
+    "OPENEBOOKS_EARLY_GRADES_USER_USERNAME": "",
+    "OPENEBOOKS_EARLY_GRADES_USER_PW": ""
```
samanthaandrews, line 11: @kristojorg Moved all these values into environment variables. I still think having the key in the cypress config is valuable, so folks who are new to the repo understand that these are being passed in _somewhwere_ and it's not happening by magic.

cypress/integration/open-ebooks/browse.ts
```diff
@@ -0,0 +1,187 @@
...
+    cy.findByRole("heading", { name: "Staff Picks collection" }).should(
+      "exist"
+    );
+
+    // Click on another see more
+    cy.findByRole("link", { name: "See more: Staff Picks collection" }).click();
+    cy.intercept("GET", `${SERVER_URL}/feed/407?entrypoint=Book`, {
+      fixture: "open-ebooks/high-school/staff-picks.html"
+    }).as("staffPicks");
+    cy.wait("@staffPicks");
```
samanthaandrews, line 44: @hokei @kristojorg I experimented with mocking out _some_ of this data (I think in 3 or 4 different places in this file), mostly where it seemed to get really specific about one particular book. Otherwise, I kept communication with the server. Lmk what you think.

cypress/integration/open-ebooks/read-book.ts
```diff
@@ -0,0 +1,17 @@
...
+  });
+  it("Read Online button exists and works", () => {
+    cy.visit(LOANS_PATH);
+    cy.intercept("GET", `${SERVER_URL}/loans`, {
+      fixture: "open-ebooks/all-access/loans.html"
+    }).as("loans");
+    cy.wait("@loans");
+    cy.findAllByText("Ready to Read!").should("have.length", 3);
+  });
+});
```
samanthaandrews, line 15: Here I was just experimenting with `cy.intercept()` and fixtures.

cypress/integration/open-ebooks/report-problem.ts
```diff
@@ -0,0 +1,55 @@
...
+      .type("frontend cypress test");
+    cy.findByRole("button", { name: "Submit" }).should("exist").click();
+    cy.intercept(
+      "POST",
+      `${SERVER_URL}/works/Axis%20360%20ID/0013217610/report`,
+      {
+        fixture: "open-ebooks/report-problem-form.json"
+      }
+    ).as("submission");
+    cy.wait("@submission");
```
samanthaandrews, line 46: @kristojorg @hokei I did successfully mock this server request 🎉

hokei: 🎉

cypress/integration/open-ebooks/age-appropriateness.ts
```diff
@@ -0,0 +1,373 @@
...
+    cy.visit(MIDDLE_GRADES_COMICS_COLLECTION_PATH);
+    cy.findByRole("heading", { name: "Comics" }).should("exist");
+    cy.visit(MIDDLE_GRADES_DETAIL_BOOK_PATH_ABBY_CARNELIA);
+    cy.findByRole("heading", {
+      name: "Abby Carnelia's One and Only Magical Power"
+    }).should("exist");
+    cy.visit(MIDDLE_GRADES_STAFF_PICKS_COLLECTION_PATH);
+    cy.findByRole("heading", { name: "Staff Picks" }).should("exist");
+
+    // FIXME: Verify you cannot navigate to collections or books outside your age class
```
hokei, line 265: did we decide to make these test assert that you can navigate to them, as documentation that this is currently the functionality?  I'm reluctant to make and link a ticket because that's how stuff disappears into the Jira void, but at least we have it here?

hokei:
This is looking really good!  I'll leave it to Kristo for final sign off but great job!

cypress/integration/open-ebooks/age-appropriateness.ts
samanthaandrews: I also am reluctant to make a ticket until we know the answer to Kristo's question above... https://github.com/NYPL-Simplified/circulation-patron-web/pull/264#discussion_r686731553

I'll make a note to circle back on this with Leonard (I think he is on vacation) and either make a ticket to block these assets based on user type OR make these tests assert the behavior that Kristo described. It's strange because some of these tests I wrote passed, which means that they indeed blocked resources. However, they were often flaky.

cypress/integration/open-ebooks/age-appropriateness.ts
```diff
@@ -0,0 +1,373 @@
...
+    cy.visit(HIGH_SCHOOL_ROMANCE_COLLECTION_PATH);
+    cy.findByRole("heading", { name: "Romance" }).should("exist");
+    cy.visit(HIGH_SCHOOL_AUTHOR_RECOMMENDATIONS_PATH_JENNIFER_RUSH);
+    cy.findByRole("heading", { name: "Rush, Jennifer" }).should("exist");
+    cy.visit(HIGH_SCHOOL_DETAIL_BOOK_PATH_ALTERED);
+    cy.findByRole("heading", { name: "Altered" }).should("exist");
+
+    // FIXME: Verify you cannot navigate to collections or books outside your age class
+    // cy.visit(MIDDLE_GRADES_COMICS_COLLECTION_PATH);
+    // cy.findByText("404 Error: No such lane.").should("exist");
```
kristojorg, line 159: Is it also true that you can currently directly navigate to collections outside your age class?

cypress/integration/open-ebooks/age-appropriateness.ts
kristojorg: So we do need a JIRA ticket in the staging sprint (I think you made one yesterday) for:
- Make sure patrons cannot directly visit books outside their age class (send token on book entry requests)

One thing we want to make sure of when we do this - we currently get OPDS entries in full when we request a collection, and we save those OPDS entries into the cache keyed by the entry url, so that if we then request the entry, it will immediately be available because we have it in the cache from the collection. We want this pattern to still work when we are also sending tokens for the entry requests. We can chat more about it when it comes time

kristojorg (approved):
This looks fantastic. I think all we need to do is make a few tickets coming out of this:
- Make sure to protect the book details page from age-inapropriate requests. But still preserve the cacheing behavior we have setup so that we can save the entries we get in a collection request and not have to re-fetch them when we navigate to a book.
- Make the tests able to run as e2e and integration (with server mocks). Or some other similar setup so we can have both e2e and integration.
- Replace the duplicated `generateToken` with one imported from the app. This will require fixing the App Setup Error we get.
- Get GH Actions set up (I know you're already on this)
- Fix storybook build (idk if there is a ticket for this already)

If you can make sure we have those tickets @samanthaandrews in the staging sprint that would be awesome.

LGTM! 🥳

kristojorg:
Maybe also make an update to the changelog so that it passes that test and then I'll merge despite the storybook failure

cypress/integration/open-ebooks/age-appropriateness.ts
leonardr: I'm not sure exactly what you're asking so I'll restate the rules. If your account has an external_type ("E") that is associated with a root lane ("Early Grades") and you try to access a lane that is neither that root lane nor a recursive child of that root lane ("Middle Grades / Comics") you should get a 404.

kristojorg: Sorry I was actually asking sam if that behavior was working or not. We currently have a known bug where we do not send the token with OPDS Entry requests, which means that authenticated patrons are erroneously able to view any book if they directly navigate to it with a link. 

The fact that this test is commented out for the _collection_ as well as the book, makes me wonder if we are also erroneously able to directly navigate to collections outside your root lane while logged in. This would be unexpected and another (different) bug if true

samanthaandrews: Collection lanes seem to be behaving as expected, i.e. getting blocked with a 404 for users outside their age class. For some reason, though, these tests were still flaky even for collection lanes. I'll make sure they're working as expected in the ticket that addresses the blocking resources at a book level.

PR #268, opened 2021-08-18
Updating Cypress scripts & adding info about running Cypress to the README

samanthaandrews:
In an effort to make smaller, easier to read PRs, here is a quick change I made that I figured should be made separately from the GitHub actions config I have been working on. This makes it easier to run cypress tests locally without having to pass a long string of environment variables. Instead, we are encouraged to add usernames and passwords to the `.env.local` file that is not committed to the repo.

This PR is not associated with a JIRA ticket. Let me know if that's ok.

Expect more PRs like this 🎉


README.md
```diff
@@ -230,6 +230,26 @@ test("fetches search description", async () => {
 });
 
 ```
+## Cypress Tests
+
+In order to run Cypress tests locally, you'll need to add a set of specific usernames and passwords to a `cypress.env.json` file. This file is not committed to the repository to keep these passwords secure. Usernames and passwords are marked in the [Contributed test library cards](https://docs.google.com/spreadsheets/d/1D0153QostVdk6YWMSZ4Vv_aBa2TxpKJ8vW7R8biiZoA/edit#gid=317403225) document.
```
samanthaandrews, line 235: Is it ok that I added a link to this document in the README?

package.json
```diff
@@ -17,7 +17,12 @@
     "build-and-start": "npm run build && npm run start",
     "test": "jest --watch",
     "test:ci": "jest --ci",
-    "test:integration": "cypress open",
+    "cypress:open": "cypress open",
+    "cypress:run": "cypress run",
```
samanthaandrews, line 21: I added these "run" commands because that's what is used in the CI, and the tests I had written are flaky in this mode. So now I'm fixing some tests to be more stable using the `cypress run` command. It seems that using `cypress open` is more forgiving.

CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 ## CHANGELOG
-- Add: Set up Cypress tests for browsing and ensuring age appropriateness
+- Add: Update scripts for Cypress and add Cypress instructions to the README
```
kristojorg, line 2: We don't need to have this here, all entries just go under `###UNRELEASED CHANGES`

README.md
kristojorg: No, you can reference it by name but I think we will have outside contributors asking for permission to view this doc if we do this. I would also be specific about the fact that the cypress tests are currently only configured for the open ebooks setup...

package.json
kristojorg: very strange, but good to know!

kristojorg (approved):
Love this! My one question is: if we simply put those variables in `.env.local`, would they be picked up by cypress? We already have that env file for environment variables that isn't checked in, would it be simpler to instruct people to write them there?

If that wouldn't work, I'm approving : )

README.md
samanthaandrews: Ok, I removed it 👍

samanthaandrews:
Ok, environment variables can be added to .env.local instead of cypress.env.json. Merging now!

PR #270, opened 2021-08-20
SFR-1253: Adding GitHub Actions for Cypress integration tests and end to end tests

samanthaandrews:
Adds two different GitHub Actions workflow files:
1. `oe-e2e-cypress.yml` only runs tests in the `server-e2e` directory & on a daily cron schedule (8 AM)
2. `oe-integration-cypress.yml` only runs tests in the `integration` directory every time a PR is created or updated

Currently these run tests in a chrome browser, on a desktop sized-screen and on a mobile sized-screen. There is an issue with Firefox that is currently throwing an error when I run these tests (I think it is related to this https://github.com/github/fetch/issues/310).


.github/workflows/oe-e2e-cypress.yml
```diff
@@ -0,0 +1,118 @@
...
+        run: npm i cypress
+
+      - name: Chrome desktop tests
+        uses: cypress-io/github-action@v2
+        with:
+          install: false
+          browser: chrome
+          spec: cypress/tests/open-ebooks/server-e2e/**/*
+        env:
+          CYPRESS_BASE_URL: https://beta.openebooks.us
```
samanthaandrews, line 45: I'm not sure if we should be testing on this prod config in the E2E tests. Maybe would be better to test on the latest dev deployment? Do we have a deployment configured for that for Open EBooks?

hokei: I think the reason to do it on prod is that we currently have no monitoring whatsoever, so no one will tell us if beta.openebooks.us is down until someone emails someone?

hokei: Is there a place these results get reported?

samanthaandrews: Once merged, I should receive an email with the results of the daily cron job since I'm the one who initially created the workflow. Here's some [info](https://docs.github.com/en/actions/guides/about-continuous-integration#notifications-for-workflow-runs) about when the notifications would be sent to a different user.

The goal is to get these posting to a QA slack channel, but I'd like to monitor the results for a few days before doing so, just to make sure things are running smoothly. I also need to figure out which slack channel to send them to. I'll ask in Standup tomorrow!

kristojorg: @hokei I don't think these will work on prod because we won't have the stable "test collection". We could eventually set up some basic tests to run on prod that login and simply read a book or something

.github/workflows/oe-e2e-cypress.yml
```diff
@@ -0,0 +1,92 @@
...
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: |
+          echo "$GITHUB_CONTEXT"
+
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+          token: ${{ secrets.AXIS_NOW_ACCESS_TOKEN }}
```
kristojorg: I may be wrong but I actually don't think we need either of these anymore. We used to use a submodule for the axis now decryptor but moved away from that, and since we aren't actually installing everything I don't know we need the token either

PR #275, opened 2021-08-31
SFR-1267: Fix Storybook build

samanthaandrews:
- Upgrades storybook to the latest version
- Adds `core-js` as a dependency
- Removing webpack config aliases for `core-js` modules and features

Following recommendations from this storybook issues thread: https://github.com/storybookjs/storybook/issues/11255


Toxiapo (approved):
Looks good to me!

PR #276, opened 2021-09-01
Fixing cypress example .env.local in README

samanthaandrews:
@Toxiapo noticed that I had some incorrect info in the cypress tests section of the README. This PR fixes those.


README.md
```diff
@@ -234,21 +234,21 @@ test("fetches search description", async () => {
...
-    "OPENEBOOKS_ALL_ACCESS_USER_USERNAME": "",
-    "OPENEBOOKS_ALL_ACCESS_USER_PW": "",
-    "OPENEBOOKS_HIGH_SCHOOL_USER_USERNAME": "",
-    "OPENEBOOKS_HIGH_SCHOOL_USER_PW": "",
-    "OPENEBOOKS_MIDDLE_GRADES_USER_USERNAME": "",
-    "OPENEBOOKS_MIDDLE_GRADES_USER_PW": "",
-    "OPENEBOOKS_EARLY_GRADES_USER_USERNAME": "",
-    "OPENEBOOKS_EARLY_GRADES_USER_PW": ""
-}
+CONFIG_FILE=config/opene.yml
```
Toxiapo, line 242: Should we add a description to tell the user to add this new config file?

Toxiapo (approved):
Just a small comment that can be ignored if you think it's fine, otherwise LGTM!

README.md
samanthaandrews: I think that part is described in a section that comes before, `Configure Multiple Config Files` which also has the example open ebooks config file there.

PR #280, opened 2021-09-07
SFR-1268: Block direct navigation to books above a users' age class

samanthaandrews:
We weren't properly passing in all 3 arguments (`bookUrl: string`, `catalogUrl: string`, and `token?: string`) to fetch books, which is why this bug appeared to be happening. Now, when an early grades user is logged in and they try to directly navigate to a high school book (`/app/collection/https%3A%2F%2Fqa-circulation.openebooks.us%2FUSOEI%2Fworks%2FAxis%2520360%2520ID%2F0013215327`), they are given a page that displays the 451 error that comes from the server.

Screen Shot 2021-09-07 at 2 15 17 PM

Users who are not logged in at all are still able to access books directly, which I believe is the desired behavior. Correct me if I'm wrong!

This PR:
- passes the correct arguments to fetch books
- passes the arguments to cache books using the `mutate()` function from `swr`
- updates Cypress tests to reflect the desired behavior

Jira issue: [https://jira.nypl.org/browse/SFR-1268](https://jira.nypl.org/browse/SFR-1268)


src/utils/cache.ts
```diff
@@ -6,19 +6,25 @@ import { mutate } from "swr";
...
 
-async function cacheBooks(books: AnyBook[]) {
+async function cacheBooks(
+  books: AnyBook[],
+  catalogUrl: string,
+  token?: string
+) {
   books.forEach(book => {
-    mutate(book.url, book, false);
+    mutate([book.url, catalogUrl, token], book, false);
```
samanthaandrews, line 28: Is there a way to confirm/test that this is behaving how I expect?

kristojorg: Good question... For a manual test what I would do is load a collection and then turn off internet and click on a book. You should be able to see the book details still

cypress/tests/open-ebooks/server-e2e/age-appropriateness.ts
```diff
@@ -146,6 +146,15 @@ describe("High school access", () => {
     cy.findByRole("heading", { name: "Altered" }).should("exist");
   });
 
+  it("Can navigate directly to books below their age class", () => {
```
kristojorg, line 149: I actually didn't know this was the intended functionality. @leonardr can you confirm that high school users should be able to navigate to middle grades books and early grades books?

src/components/bookDetails/index.tsx
```diff
@@ -21,22 +21,48 @@ import extractParam from "dataflow/utils";
...
+    if (error instanceof ServerError) {
+      fetchBookError = error;
+    } else {
+      fetchBookError = new ApplicationError(
+        {
+          title: "Fetch Book Error",
+          detail: "An unknown error ocurred while fetching the book"
+        },
+        error
+      );
```
kristojorg, line 63: Were you seeing this come up ever? I think generic errors should already be caught higher up the tree but I could be wrong

kristojorg (approved):
Looks good to me! Just the one question about the custom error, wondering if we need that or if we should handle all fetch errors like that in a more generic way?

cypress/tests/open-ebooks/server-e2e/age-appropriateness.ts
leonardr: No, an authenticated user whose patron type has an associated root lane 1) can only access their root lane and lanes underneath it, 2) will not see links that take them above their root lane.

However, an _unauthenticated_ user can access any lane, and that's probably what's going on here -- the client is dropping the credentials.

Here's an example: the "E" patron can access lane 44 (the Spanish-language lane underneath "Early Grades") but the "M" patron cannot:

curl https://E31T7SX02T:[pw]@circulation.openebooks.us/USOEI/feed/44 
`200`

curl https://MLY968Q99G:[pw]@circulation.openebooks.us/USOEI/feed/44 
`404`

An unathenticated patron can see the feed with no problem:

curl https://circulation.openebooks.us/USOEI/feed/44 
`200`

I feel like I've written this before, but we did it this way so we could implement a _public_ web catalog that allowed teachers to browse the collection before signing up. That's not how beta.openebooks.us works, so we could potentially implement a stricter authentication regime on the circulation manager side -- but that would be extra work and I'm still not convinced the lack of a public catalog wasn't an oversight.

samanthaandrews: Linking to this thread for posterity: https://nypl.slack.com/archives/C0153AFA1RS/p1631115972013400

Thanks so much for your explanations @leonardr . This makes sense to me.

src/components/bookDetails/index.tsx
samanthaandrews: I wasn't seeing this at all, so I'm thinking I can just remove it!

src/utils/cache.ts
samanthaandrews: Hmm this is not working but it is also not working in prod. I'm thinking I will create another bug ticket to investigate caching behavior for books.

samanthaandrews: https://jira.nypl.org/browse/SFR-1292

kristojorg: Good point, thanks for the ticket

PR #282, opened 2021-09-16
SFR-1269: Import generateToken function

samanthaandrews:
Cypress was unable to import the `generateToken()` function from `auth/useCredentials.ts` because that file references `utils/env.ts` which references process.env.APP_CONFIG. We were hitting the AppSetupError from that file:  `process.env.APP_CONFIG is undefined`

To fix this, I moved the `generateToken()` function into its own file where there are no references to `process.env`. I realize this isn't a complete fix, but it seems good enough for now. Didn't want to spend too much time on this until it becomes clear that we absolutely must give Cypress access to process.env values in the code... which was proving to be more difficult than it should be.

I also replaced the BASIC_LOGIN string in the Cypress utils file with the BasicAuthType value that already exists in `types/opds1`


src/test-utils/fixtures/auth.ts
```diff
@@ -6,17 +6,15 @@ import {
 } from "interfaces";
 import { OPDS1 } from "interfaces";
 import { makeFulfillableBooks } from "test-utils/fixtures/book";
-
-export const basicAuthId = "http://opds-spec.org/auth/basic";
-export const samlAuthId = "http://librarysimplified.org/authtype/SAML-2.0";
```
samanthaandrews, line 11: Removed these because the `samlAuthId` was not being referenced anywhere in the file and imported BasicAuthType value from "types/opds1"

hokei (approved):

PR #283, opened 2021-09-20
New Web Reader

kristojorg:
**NOTE:** To view the new web reader, open a book normally and then change the url from `/read` (old reader) to `/read-v2` manually. 

TODO

- [x] Handle errors
  - Handled auth errors and added error boundary
- [x] Make a `/read-v2/` url for new reader
- [x] Figure out why manifest info is fetched ~~twice~~ three times
   - Update: it is fetched thrice because the client fetches it to set up a decryptor for the reader, and the reader fetches it at least once to display the book. The third time is still a mystery. I added some cache headers that should make it faster, but this should be split out into another ticket. The WebReader should take an already fetched JSON manifest.
- [x] ~~Possibly fetch fulfillment info and manifest in `getStaticProps`~~ would block rendering too long. Best bet is to display a better loader
- [x] Add "back to Open eBooks" button
- [ ] Get WebWorker working (separate PR?)


public/css/ReadiumCSS-default.css
```diff
@@ -0,0 +1,169 @@
+/* Readium CSS
```
hokei, line 1: Can these be exported with the web reader? Seems like everything that's using the reader will need these.

public/fonts/opendyslexic/OFL.txt
```diff
@@ -0,0 +1,94 @@
+Copyright (c) 2019-07-29, Abbie Gonzalez (https://abbiecod.es|support@abbiecod.es),
```
hokei, line 1: same with this - can it be something exported from the webreader as a "sensible default"?

src/pages/[library]/read-v2/[bookUrl].tsx
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * - get the book url
+   * - get the fulfillment info
+   * - create the manifest url
+   * - set up decryptor worker
+   */
+
+  const { token } = useUser();
+  const { query } = useRouter();
+  const [getContent, setGetContent] = React.useState<null | GetContent>(null);
```
Comment (hokei) [inline src/pages/[library]/read-v2/[bookUrl].tsx:55]: nit: there's gotta be a better name for this

public/css/ReadiumCSS-default.css
kristojorg: Yea so these are exported with the web reader. The hard part here is that both of these are "injected" into the reader iframe, as in the iframe has a `<style href="url">` tag where we can control the `url`. Thus we need to provide the web reader with a link to a version of this file. Exporting it from our package doesn't really help us here, and I'm not sure what the ideal solution would be. I tried importing it in my reader page, but it got applied to the host app and not to the iframe, which is obviously not what we're after. 

Some app frameworks have a way to import a file as a url as in `import hrefForStyles from "@nypl/web-reader/html-style.css"` would give us an href we can pass to the web reader... but I don't see that in Next.js. In fact I needed this for something else previously and commented on a issue but it doesn't seem resolved: https://github.com/vercel/next.js/discussions/12844

So for now I think this is the best solution sadly. The other solutions I can think of are:
- Use `unpkg` as the default in the web-reader with a url like this: https://unpkg.com/@nypl/web-reader@0.2.0-alpha.2/dist/html-styles.css . Obviously then we (and any consumers using the default) are relying on a third party cdn for our css. Maybe fine? Idk
- Host the styles somewhere ourselves alongside the web-reader repo (don't really think this is worth the work).

hokei: OK, I think the important thing here is to make sure that the instructions for installing it are super clear.  We can definitely work on this after integrating.

hokei: Super naiive question:  If these files are in `node_modules/web-reader` why can't we make the package manager identify and use the files from there, copying them to dist, etc? 

(either way, new ticket)

kristojorg: > Super naiive question: If these files are in node_modules/web-reader why can't we make the package manager identify and use the files from there, copying them to dist, etc?

In some package managers we can. Maybe there is a way to config next.js to do this, but I haven't seen it. In next.js we can just import something, which will include it in our bundle, but we can't get a url to that file sadly

src/pages/[library]/read-v2/[bookUrl].tsx
```diff
@@ -0,0 +1,123 @@
...
+import extractParam from "dataflow/utils";
+import { useRouter } from "next/router";
+import useSWR from "swr";
+import useUser from "components/context/UserContext";
+import { fetchJson } from "dataflow/fetch";
+import { PageLoader } from "components/LoadingIndicator";
+import { GetContent, WebpubManifest } from "@nypl/web-reader/dist/types/types";
+
+//@TODO - fix font imports in the readium css
+import "@nypl/web-reader/dist/pdf-styles.css";
```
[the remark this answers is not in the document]
Comment (kristojorg) [reply src/pages/[library]/read-v2/[bookUrl].tsx:17]: Yea true we don't need that here, I did that for testing purposes. Will remove.

src/pages/[library]/read-v2/[bookUrl].tsx
```diff
@@ -0,0 +1,142 @@
...
+          alignItems: "center",
+          justifyContent: "center",
+          height: "100vh"
+        }}
+      >
+        <PageLoader />
+      </div>
+    );
+  }
+  return <WebReader webpubManifestUrl={manifestUrl} getContent={decryptor} />;
```
Comment (hokei) [inline src/pages/[library]/read-v2/[bookUrl].tsx:130]: Are we doing the import instead of the hook because it's easier?  Or is there a reason to use this instead of the hook?

Comment (kristojorg) [reply src/pages/[library]/read-v2/[bookUrl].tsx:130]: You only need to use the hook if you want to render your own UI. If you're fine with the default reader UI then you can just import the component and use it this way

Comment (kristojorg) [reply src/pages/[library]/read-v2/[bookUrl].tsx:130]: The hook basically gives you the content and all the apis to change the content like `goForward` etc. You could then use those apis with your own wrapper and buttons and stuff, if you want

kristojorg:
@samanthaandrews Once we merge your web reader tests, I'd like to update this PR to make those tests run against both `/read` and `/read-v2` so we can see if/how they fail on the new web reader

hokei:
I'm trying to read a book on the vercel preview deploy and getting an axis error: 
Screen Shot 2021-09-24 at 8 53 59 AM
g

samanthaandrews:
Running locally I believe I'm seeing the same error as @hokei, but also I might not have some setting configured correctly? Crystal do you think I need to set up something in my repo to give me access to the private package?

Screen Shot 2021-09-24 at 11 21 19 AM

I am also noticing that when I navigate to my loans, the "Read" button is missing. I am thinking these two issue are related? No errors in the console or network tab on this page.

Screen Shot 2021-09-24 at 11 23 44 AM

hokei:
Looks like there are issues with image rendering
Screen Shot 2021-09-28 at 11 57 25 AM


This book is - Navajo Code Talkers
by Aaseng, Nathan

I'll try to figure out if this is this particular book or if other books have this.

src/pages/api/axisnow/[isbn]/[vaultId].ts
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable camelcase */
+import Decryptor from "@nypl-simplified-packages/axisnow-access-control-web";
+import { RemoteFetcher, Epub } from "@nypl/epub-to-webpub";
+import { NextApiHandler } from "next";
+import extractParam from "dataflow/utils";
+import track from "analytics/track";
+
+/**
+ * This is a handler for Open eBooks Axisnow encrypted EPUBS.
+ * eg: http://localhost:3000/api/axisnow/9781467784870/6734F7F5-C48F-4A38-9AE5-9DF4ADCFBF0A
```
Comment (hokei) [inline src/pages/api/axisnow/[isbn]/[vaultId].ts:10]: is this a real vault id and isbn?  If yes, can we replace with just `http://localhost:3000/api/axisnow/[isbn]/[vaultId]` like in this path?

hokei (approved):
This looks good to me.  Let's get this into QA.  Great work!

PR #284, opened 2021-09-21
SFR-1305 & SFR-1283 Adding web reader integration tests and fixing 502 errors

samanthaandrews:
- Adding web reader integration tests
- Adding an extra step to the `loginByApi` Cypress command (stubbing out the `/loans` request and waiting for it to finish) , which mitigates the 502 errors we were seeing
- Change `oe-integration-cypress.yml` GH Action to run on all cypress tests, both `server-e2e` and `integration`, now that the 502 errors have been solved
- Reorganizes & removes duplicate Cypress tests that were created because we were splitting up tests into two different CI workflows

What the directory structure look like (i'm finally happy with this 😅)
Screen Shot 2021-09-21 at 3 32 20 PM


.github/workflows/oe-integration-cypress.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           record: true
           parallel: true
           group: "Integration - Chrome"
-          spec: cypress/tests/open-ebooks/integration/**/*
+          spec: cypress/tests/open-ebooks/**/*
```
samanthaandrews, line 43: This is the change that makes it so all tests in the `cypress/tests` directory are being run on every PR deployment. This allowed me to remove some tests that we were duplicating. I didn't change anything in the `oe-e2e-cypress.yml` workflow since I figured those same e2e tests should still be the only ones running on the daily schedule.

cypress/support/commands.ts
```diff
@@ -48,6 +51,10 @@ type UserType =
 Cypress.Commands.add(
   "loginByApi",
   (userType: UserType, appType: AppType = "OPENEBOOKS") => {
+    cy.intercept("GET", `${SERVER_URL}/loans`, {
+      statusCode: 200,
+      fixture: "open-ebooks/all-access/loans.html"
+    }).as("loans");
```
samanthaandrews, line 57: Mocking out this data and waiting for it on line 78 is what has fixed the 502 errors. This means we are returning mocked loans data for tests that use the `loginByApi` function. I think this is ok since we have actual e2e tests for login with Clever and FirstBook that don't use this command.

cypress/tests/open-ebooks/server-e2e/age-appropriateness/all-access.ts
```diff
@@ -19,11 +19,6 @@ describe("All access", () => {
   it("Landing page displays all 3 access lanes", () => {
     cy.visit(APP_PATH);
 
-    cy.log("verify collection title");
-    cy.get("main")
-      .findByRole("heading", { name: "Open eBooks (QA Server) Home" })
-      .should("exist");
-
```
samanthaandrews, line 26: Moved these checks into the `loginByApi` command.ts file above.

cypress/tests/open-ebooks/server-e2e/browse/browse.ts
```diff
@@ -38,10 +33,10 @@ describe("All-access browsing", () => {
     cy.get("[data-testid=listview-list]").find("li").as("bookList");
     cy.get("@bookList").should("have.length.at.least", 1);
 
-    cy.findAllByText("Available to borrow").should("have.lengthOf", 5);
+    cy.findAllByText("Available to borrow").should("have.lengthOf", 4);
```
samanthaandrews, line 36: I checked out a book in this list since it was one from our special "integration testing lane" so I had to adjust the tests to reflect that the list had been updated.

cypress/tests/open-ebooks/server-e2e/web-reader/navigation.ts
```diff
@@ -0,0 +1,44 @@
...
+    cy.log("paginated loads as default");
+    cy.findByRole("menuitem", { name: "Paginated mode enabled" }).should(
+      "exist"
+    );
+    cy.log("close settings card");
+    cy.findByRole("button", { name: "Settings" }).click();
+
+    cy.log("click next page button");
+    cy.get("div[id='next-page-btn']").find("button").should("exist").dblclick();
+    cy.wait(5000);
```
samanthaandrews, line 21: You'll see a few wait() calls in this file. Honestly it seemed to be the only thing that would make these tests behave as expected.

cypress/support/commands.ts
```diff
@@ -67,5 +74,51 @@ Cypress.Commands.add(
...
+    cy.get("main")
+      .findByRole("heading", { name: "Open eBooks (QA Server) Home" })
+      .should("exist");
   }
 );
+
+Cypress.Commands.add("openLoanedEPub", (url: string) => {
+  cy.visit(url);
+  cy.findByRole("heading", { name: "Loading..." }).should("exist");
+  cy.wait(10000);
```
samanthaandrews, line 88: This is where I am waiting for the iframe to load. I tried many different variations of this... waiting for the loading spinner image (sometimes it appeared and sometimes it didn't), waiting for the package.opf requests, but nothing was as reliable as a plain ol' wait :/

cypress/tests/open-ebooks/server-e2e/web-reader/navigation.ts
kristojorg: Ok good to know, I guess I'll follow a similar strategy in the web reader repo. Bummer that there isn't a better solution, but whatever for now it's fine. Why do we need the double click though?

cypress/support/commands.ts
samanthaandrews: Yes, sort of. The source of the 502 errors that we were seeing in Cypress were happening as a result of setting the cookie and then going right into starting tests and making assertions without waiting for the authentication document to come back from the backend (kind of a race condition in a sense). By checking that the "loans" request has finished, we are now testing a more realistic app behavior. I have never actually seen these 502 errors in CPW, _except_ for when open ebooks backend was actually down, which seems appropriate. The 502 errors were consistently happening in Cypress because the automation was just going a bit too quickly.

I'd argue that in this case, the fixture is ok since we're not trying to test authentication/login with this step. This just ensures that the body of each test runs more reliably and quickly. We do have true e2e login tests with FirstBook that will fail if something is actually wrong with the backend or a third party.

.github/workflows/oe-e2e-cypress.yml
```diff
@@ -87,14 +87,14 @@ jobs:
      
       - name: Upload failure screenshots
         uses: actions/upload-artifact@v2
-        if: failure()
+        if: ${{ failure() }}
```
samanthaandrews, line 90: i think both of these syntax choices are correct, but wanted to keep it consistent

cypress/support/commands.ts
```diff
@@ -67,5 +76,56 @@ Cypress.Commands.add(
...
     cy.getCookie(COOKIE_KEY).should("exist");
+    cy.visit(APP_PATH);
+    cy.findByRole("heading", { name: "Loading..." }).should("be.visible");
+    cy.wait("@loans").then(interception => {
+      assert.isNotNull(interception?.response?.body, "loans API call has data");
+    });
+    cy.get("main")
+      .findByRole("heading", {
+        name: "Open eBooks (QA Server) Home",
+        timeout: 15000
```
samanthaandrews, line 87: After reading more Cypress docs, I changed all of my arbitrary `wait()` calls into timeouts. This means that Cypress will keep trying to find the element for the duration of the timeout period. Once it finds it, the test continues. It fails if the element is not found. Which means that the test likely continues before the timeout period finishes, which seems like a tiny but good optimization.

cypress/tests/open-ebooks/server-e2e/web-reader/navigation.ts
Toxiapo: I ran into a similar issue where sometimes a single click will not register. What I did was to check the loading screen right after the click, that seems working fine for me.

cypress/support/commands.ts
leonardr: If this helps, you can fetch the link with rel="http://librarysimplified.org/terms/rel/user-profile" from the authentication document and make an authenticated request to that URL (/patrons/me, basically) to _quickly_ determine whether a patron is authenticated. Making an authenticated request to the bookshelf feed (/loans, basically) can take several seconds since it has to call out to external vendors.

cypress/tests/open-ebooks/server-e2e/web-reader/navigation.ts
samanthaandrews: Ugh it did not work with the existing web reader 💩

cypress/support/commands.ts
samanthaandrews: Changed these to use timeout instead.

samanthaandrews:
I feel like Cypress was gaslighting me all day... but these tests are finally in a good place I think. The only flakiness I am still seeing is that sometimes after performing the login steps and load the app, the app gets stuck at the loading spinner. Once again, not seeing this behavior when actually using the app, just occasionally in Cypress.

cypress/tests/open-ebooks/server-e2e/web-reader/navigation.ts
kristojorg: Haha rawr iframes and web readers. Ok well we can keep it and just update it in the tests of the new web reader. It would be awesome if we could figure out the source of the flakiness with these iframes but I can't for the life of me get to the bottom of it

kristojorg:
> Once again, not seeing this behavior when actually using the app, just occasionally in Cypress.

I still wonder what this could be... maybe we only see it on cypress because we aren't doing those exact steps often enough to reproduce it manually? 

Anyways, this looks good to me. I'll agree on the fixture in the e2e test since we have other tests for the login flow as you said.

cypress/tests/open-ebooks/server-e2e/login/firstbook-login.ts
```diff
@@ -53,25 +64,29 @@ describe("FirstBook login", () => {
...
-      .should("exist")
-      .click();
-    cy.wait("@auth").its("response.statusCode").should("eq", 302);
-    cy.location("pathname").should("eq", LOGIN_PATH);
-    cy.findByRole("listitem", { name: "Current location: Login" }).should(
-      "exist"
-    );
-  });
-});
+// FIXME: I have no idea why this test fails ONLY in the CI
```
kristojorg, line 67: We should make a ticket to address this I guess. Really strange, all the flakiness sounds annoying sorry : /

cypress/tests/open-ebooks/server-e2e/web-reader/color-theme.ts
```diff
@@ -0,0 +1,34 @@
+import { ALMOST_MIDNIGHT_LOAN } from "../../../../support/utils";
+
+const findColorThemeId = (id: string) => {
+  cy.getIframeHead().find("style").should("have.id", id);
+};
+
+// FIXME: This test is flaky in CI. Fails in the login beforeEach step.
```
kristojorg, line 7: We should make a ticket for this too. Is it still true? I wonder why this would be failing

cypress/tests/open-ebooks/server-e2e/web-reader/font-settings.ts
```diff
@@ -0,0 +1,60 @@
...
+    cy.findByRole("menuitem", { name: "Sans-serif font disabled" })
+      .should("exist")
+      .click();
+    findFontId("sans-font-internal");
+
+    cy.log("can change to dyslexia friendly font");
+    cy.findByRole("menuitem", { name: "Dyslexia Friendly font disabled" })
+      .should("exist")
+      .click();
+    findFontId("dyslexia-font-stylesheet");
```
kristojorg, line 37: Do you think we should use `have.css` with `font-family` to make sure this is changing the font family or is that not necessary?

kristojorg (approved):
Just a couple questions, otherwise this looks good to me

cypress/tests/open-ebooks/server-e2e/login/firstbook-login.ts
samanthaandrews: Wait I actually figured out that this is failing in the beforeEach() hook, in the step where we are loading the app, and sometimes it gets stuck loading. So I'm going to remove these comments 👍

cypress/tests/open-ebooks/server-e2e/web-reader/color-theme.ts
samanthaandrews: Same here. Failing occasionally because of the same beforeEach() hook where we get stuck loading.

cypress/tests/open-ebooks/server-e2e/web-reader/font-settings.ts
samanthaandrews: yeah I'll add this

PR #285, opened 2021-09-22
Setup WebWorker Decryptor

kristojorg:
This PR swaps us to a WebWorker-based decryptor in the web-reader. I had to add a webpack plugin to make this work and handle the server-side case where we can't use workers. This should make reading faster!


samanthaandrews (approved):
LGTM

PR #289, opened 2021-09-30
SFR-1330 Adding more detailed Cypress tests for the new web reader

samanthaandrews:
I added a few TODO and FIXME comments based on the issues I encountered. I think this is a good first pass at a solid test suite, but we should add to these tests and change them slightly once we fix the bugs.

Tests two different books right now:
- One Amazing Thing
- Almost Midnight

Future updates should add two problematic books to the tests:
- Franklin
- Among the Hidden


cypress/tests/open-ebooks/server-e2e/web-reader-v2/open-books.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    cy.log("navigate to loans and click a book to read");
+    cy.visit(LOANS_PATH);
+    cy.wait("@loans");
+    cy.findByRole("link", { name: "Almost Midnight" }).should("exist").click();
+    cy.findByRole("button", { name: "Read" }).should("exist").click();
+
+    cy.log("open web reader");
+    cy.findByRole("img", { name: "Loading..." }).should("exist");
+    cy.findByRole("document", { name: "", timeout: 10000 }).should("exist");
```
kristojorg, line 30: Just curious, is this finding the iframe's document?

cypress/tests/open-ebooks/server-e2e/web-reader-v2/read-books.ts
```diff
@@ -0,0 +1,95 @@
...
+  V2_IFRAME_SELECTOR
+} from "../../../../support/utils";
+
+describe("Users can read/navigate loaned books", () => {
+  beforeEach(() => {
+    cy.loginByApi("ALL_ACCESS_USER");
+    cy.loadApp();
+  });
+
+  //TODO: Let's change this test to instead open the Among the Hidden book since that one was having loading & TOC issues
```
kristojorg, line 13: I think for the future, it would be awesome if we can make a function that tests any book in a generic way at least. I imagine we will run in to a handful of books that don't work, and we could just add a test for them like this:
```ts
describe('problematic books load and navigate', () => {
  it('Among the Hidden', () => { testBook(AMONG_THE_HIDDEN_PATH) })
  // others
}
```

idk just a thought. Maybe we don't need to do it for so many books and maybe it would be terribly slow, but it's an idea.

kristojorg (approved):
Looks great, lgtm!

cypress/tests/open-ebooks/server-e2e/web-reader-v2/open-books.ts
samanthaandrews: Yes... but now that you pointed it out I think it's not that useful so I removed it ha.

cypress/tests/open-ebooks/server-e2e/web-reader-v2/read-books.ts
samanthaandrews: I agree completely. We could create a generic Cypress command for it. I'll likely come back to it later this week once your fixes have been merged and implement!

samanthaandrews:
It looks like changes to the web-reader css have caused books to not load in this pr. I think we need to wait for changes from https://github.com/NYPL-Simplified/web-reader/pull/72 in order to get these tests working again

Screen Shot 2021-10-04 at 2 06 51 PM
.

PR #290, opened 2021-10-01
SFR-1323 Add ServiceWorker

kristojorg:
This adds a service worker to CPW:

1. I use next-pwa to create a manifest of precacheable files during build. This is then used in `src/ServiceWorker/index.ts`. 
2. I write my own ServiceWorker/index.ts which precaches the app files and then initializes the web reader event handlers.
3. I use `usePublicationSW` to send a message to the SW telling it to cache the publication we are opening.

**TODO**

- [x] Make next-pwa build to the correct directory
- [x] Disable logging

**EDIT**

Unfortunately, it wasn't possible to get `next-pwa` to build the service worker into a static directory in the build folder, only in the `/public` folder. Thus I had to add these files in `/public` to the gitignore so that they aren't checked in and changing all the time. A different solution would be nice, but I don't currently have one.


kristojorg:
There are currently two major performance bottlenecks when opening a book causing it to take around 10s:
1. Fulfilling the book takes 3-5s on my machine. This is required before we can fetch the manifest
2. Generating the webpub manifest using `epub-to-webpub` in the api route takes approx 2-3s, and that endpoint is called at least 2x. 

This PR will assist with #2, making it so that manifest is cached after the first time it is requested, thus cutting out ~5s of load time after the first load. But there is still obviously _much_ improvement that needs to be made if we expect people to use this.

next.config.js
```diff
@@ -30,7 +31,9 @@ const RELEASE_STAGE =
     ? "qa"
     : "development";
 
-const BUILD_ID = `${APP_VERSION}-${GIT_BRANCH}.${GIT_COMMIT_SHA}`;
+const BUILD_ID = encodeURIComponent(
+  `${APP_VERSION}-${GIT_BRANCH}.${GIT_COMMIT_SHA}`
+);
```
kristojorg, line 36: I spent a couple hours debugging 404s only to realize that the build id on my branch had a `/` in it and was used in the url of some files, which caused problems! argh

samanthaandrews: uuugh yes i've run into this too. maybe we should add a note to the README about not using `/` in branch names?

src/utils/reader.ts
```diff
@@ -50,19 +49,19 @@ async function initBookSettings(
...
+  //   serviceWorkerUrl: new URL("/sw.js", window.location.origin),
+  //   staticFileUrls: [
+  //     new URL(window.location.href),
+  //     new URL("index.html", window.location.href),
+  //     new URL("main.css", window.location.href),
+  //     new URL("require.js", window.location.href),
+  //     new URL("fetch.js", window.location.href),
+  //     new URL("webpub-viewer.js", window.location.href)
+  //   ]
+  // });
```
samanthaandrews, line 64: do we need this?

next.config.js
kristojorg: I think it should be solved now! I added this so that the build ID gets url-encoded and thus the `/` turns intoa `%2` or something.

src/utils/reader.ts
kristojorg: I think the old one should still work without its service worker. We sadly can't have two in the same app though, so I had to disable it.

next.config.js
samanthaandrews: ahh i see now. awesome!

samanthaandrews (approved):
thanks for walking us through this. there are some merge conflicts to resolve but otherwise looks good.

PR #291, opened 2021-10-02
Fix Loading Indicator

kristojorg:
This fixes the loading indicator by importing reader css to style it, and it also removes the necessity to copy the readium css files to our public folder by using a webpack loader to load a url for them and then passes that to the web reader as an injectable.


src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -23,6 +22,9 @@ import ApplicationError from "errors";
 import useLibraryContext from "components/context/LibraryContext";
 import Link from "components/Link";
 import { ChevronRight } from "icons";
+import "@nypl/web-reader/dist/esm/index.css";
```
Comment (kristojorg) [inline src/pages/[library]/read/[bookUrl].tsx:25]: This imports the css for styling the non-iframe portion of the reader.

src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -23,6 +22,9 @@ import ApplicationError from "errors";
...
+// This is how we can import a css file as a url. It's complex, but necessary
+import cssInjectableUrl from '!file-loader?{"publicPath":"/_next/static","outputPath":"static"}!extract-loader!css-loader!@nypl/web-reader/dist/injectable-html-styles.css';
```
Comment (kristojorg) [inline src/pages/[library]/read/[bookUrl].tsx:27]: This imports css as a url which we can then inject into the iframe as a link.

Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:27]: @hokei this might be useful in DRB as well

Comment (kristojorg) [reply src/pages/[library]/read/[bookUrl].tsx:27]: I presume we could find some similar pattern for the fonts too.

samanthaandrews (approved):
Curious to see how this will impact the cypress tests that are also in review. We will see!

kristojorg:
> Curious to see how this will impact the cypress tests that are also in review. We will see!

Hopefully not at all 🤞

PR #302, opened 2021-10-22
Update and improve Cypress server e2e tests

samanthaandrews:
While running the daily Cron jobs, a few of the e2e tests have had issues. Screenshots attached.

### What changed:
- I removed the v1 web reader tests. They take a while to run and I don't think they are that useful anymore.
- Added a custom command `checkoutAllAccessBook()` that checks out books before web reader tests run. Loans were expiring causing tests to fail.
- Updated Cypress to the latest version.
- Added better checks to the FirstBook login test so that it wouldn't fail while the app was actually logging in (sometimes the login step was taking longer than expected while Cypress had already moved on to the next check, which would then fail).
- Added additional checks and intercepts to the `openLoanedEpub` command to ensure that axis now resources had loaded before moving on. A lot of this was informed by the work done on the web-reader tests, where I realized we could be doing better to wait for requests to finish.
- Commented out failing Clever login test

The `open-book.ts` and `read-book.ts` tests have been failing with a generic Server Error, which Kristo has done work recently to fix, i.e. make these errors more specific and helpful. I don't think anything I changed within this PR will eliminate those errors, as I'm unsure as to what is causing them. I've confirmed that the books are still checked out under the test user account.

### Previously failing FirstBook login test:

FirstBook login -- Can login with FirstBook (failed)

### Failing `open-book.ts` and `read-book.ts` tests:

Users can open loaned books -- Can navigate back to open ebooks (failed)

Users can readnavigate loaned books -- Can navigate Almost Midnight book with TOC (failed)


kristojorg:
@samanthaandrews thanks for making that ticket. Otherwise this looks good. I do think the test is properly testing the login flow, as it is supposed to work if you navigate to any page with the access code in the url, so I think we should fix that. I could be wrong though

src/dataflow/opds1/fetch.ts
```diff
@@ -62,6 +62,7 @@ export async function fetchEntry(
   url: string,
   token?: string
 ): Promise<OPDSEntry> {
+  console.log("URL!!!", url);
```
kristojorg, line 65: I think this could use one more exclamation point

kristojorg (commented):
Just the one thing then i approve!

src/dataflow/opds1/fetch.ts
samanthaandrews: Lol 😂  ok I removed it!

kristojorg (approved):

src/dataflow/opds1/fetch.ts
kristojorg: Oh I was just kidding lol

PR #305, opened 2021-11-15
OE-257: Fix Read & Return button sometimes not displaying after clicking Borrow button

samanthaandrews:
This is a tricky bug that you can pretty easily replicate if you click the Borrow button and then the Return button a few times on any book. At a certain point, you'll click the Borrow or Return button and the correct button won't display. Or, you'll get an error that says that the book you're trying to return isn't returnable.

I made a few changes to ensure borrowing and returning books always works as expected:
- Requests made to the `/borrow` and `/revoke` endpoints are changed from "GET" to "PUT" which means they are no longer being cached server-side.
- Minor refactoring to `src/hooks/useBorrow.ts` and `<CancelOrReturn />` and `<BorrowOrReserve />` for clarity.
- Add Cypress test
- Update useSWR to the latest version to easily cancel inflight requests to the loans endpoint before borrowing or returning a book (thanks @kristojorg !)
- Update emotion and a few other packages that were causing type errors after the `swr` package upgrade.


src/components/BookList.tsx
```diff
@@ -170,9 +171,16 @@ export const BookListItem: React.FC<{
               {authors}
             </Text>
           </div>
-
-          <BookStatus book={book} />
-          <BookListCTA book={book} />
+          {isLoading ? (
+            <LoadingIndicator />
```
samanthaandrews, line 175: I noticed that we were allowing users to borrow/return books while loans were still fetching, which was also causing issues. I'll clean this up so that it looks nicer than just the word "Loading..."

src/components/context/UserContext.tsx
```diff
@@ -67,13 +79,42 @@ export const UserProvider: React.FC = ({ children }) => {
...
-  function setBook(book: AnyBook, id?: string) {
-    const existing = data ?? [];
+  function borrowBook(
+    borrowUrl: string,
+    catalogUrl: string,
+    token: string,
+    setLoading: (isLoading: boolean) => void
+  ) {
+    mutate(async existingLoans => {
+      const book = await fetchBook(borrowUrl, catalogUrl, token);
```
samanthaandrews, line 89: I moved this fetchBook() function inside the mutate as I saw this example in the swr docs and felt like this is what we are trying to do: https://swr.vercel.app/docs/mutation#mutate-based-on-current-data

This fixes the issue, but I'm still seeing the problem where old loans are being fetched if the cache is enabled.

src/components/context/UserContext.tsx
```diff
@@ -67,13 +79,42 @@ export const UserProvider: React.FC = ({ children }) => {
...
+  ) {
+    mutate(async existingLoans => {
+      const book = await fetchBook(revokeUrl, catalogUrl, token);
+      // if the id exists, remove that book and set the new one
+      const withoutOldBook = existingLoans?.filter(
+        loanedBook => loanedBook.id !== book.id
+      );
+      setLoading(false);
+      if (withoutOldBook) {
+        return [...withoutOldBook];
```
samanthaandrews, line 115: I removed the book from the array of loans entirely since we are returning it.

kristojorg:
Ok so just taking a couple notes here:

- I also see that when you click, sometimes you will see a loading button, and then it flashes a borrow button (presumably from our mutation), and then it returns to a return/read button, presumably from the server revalidation with invalid data.

src/dataflow/fetch.ts
```diff
@@ -4,16 +4,19 @@
  */
 
 import ApplicationError, { FetchError, ServerError } from "errors";
+import { OPDSFetchMethods } from "types/opds1";
 
 export default async function fetchWithHeaders(url: string, token?: string) {
+  const method: OPDSFetchMethods =
+    url.includes("/borrow/") || url.includes("/revoke") ? "PUT" : "GET";
```
samanthaandrews, line 11: Does this cover all the endpoints that should be using "PUT"? These were the ones that I noticed in Open eBooks but I realize there may be other endpoints for other configurations?

kristojorg:
So a couple thoughts here:

- When you return or borrow a book we revalidate the loans, and all books will show a loading indicator while that is happening. This isn't the end of the world, but seems like not what we want to do if we don't want to. I wonder if we can simply disable the buttons if there is no loans data existing and loans is currently being fetched. It depends what error case we are trying to resolve with the loading indicator. Can you describe it?
- We no longer use the value returned from the server when borrowing a book to optimistically update our loans cache before revalidating it. Is this necessary, or can we keep the prior behavior?

https://user-images.githubusercontent.com/4398775/142912169-a4cf102e-5707-4825-8862-bf2b1a3c2b19.mov

samanthaandrews:
> So a couple thoughts here:
> 
> * When you return or borrow a book we revalidate the loans, and all books will show a loading indicator while that is happening. This isn't the end of the world, but seems like not what we want to do if we don't want to. I wonder if we can simply disable the buttons if there is no loans data existing and loans is currently being fetched. It depends what error case we are trying to resolve with the loading indicator. Can you describe it?
> * We no longer use the value returned from the server when borrowing a book to optimistically update our loans cache before revalidating it. Is this necessary, or can we keep the prior behavior?
> 
> The reason I press on this is that the whole idea behind the SWR strategy is to trust your cache (and update it optimistically), but revalidate it in the background to make sure you're eventually consistent with the server. In this version, we are revalidating in the foreground.
> 
>  Screen.Recording.2021-11-22.at.7.01.25.PM.mov 
> I know you mentioned above this was the only way to make it show the right button, but I guess I'm wondering - if we switch to the PUT request, then won't it show the right button without doing it this way, even if it briefly shows an incorrect button if you click it multiple times quickly?

You're right; I re-implemented the optimistic updating to loans and it seems to be working. In my previous implementation I was trying to update within the `mutate()` call and that wasn't behaving as expected. I also passed the `false` parameter to the `refetchLoans()` function, because it doesn't really seem necessary to re-validate right away... but curious to hear what you think about this? This way, the loading spinner displays upon first load, but not after clicking the Borrow or Return buttons, which seems like the ideal balance.

I abandoned the Cypress test because I could not get the page to display the correct Return or Borrow components after clicking the Borrow or Return buttons. I think this is an issue with the useSWR hook and Cypress, as someone reported a similar issue here about a year ago, but there has been little action on it. There is an [example repo](https://github.com/bahmutov/test-swr) for how to write a component test that uses the swr hook, but we're not really looking for a component test. Happy to pair on this if you want to look into it!

kristojorg (commented):
Some comments I want to get out now

cypress/support/commands.ts
```diff
@@ -65,9 +65,22 @@ Cypress.Commands.add(
...
+      command: "Network.setCacheDisabled",
+      params: { cacheDisabled: true }
+    });
+  }
+  cy.intercept(`${SERVER_URL}/loans`, { middleware: true }, req => {
+    delete req.headers["if-none-match"];
+    req.on("before:response", res => {
+      // force all API responses to not be cached
+      res.headers["cache-control"] = "no-store, no-cache";
+    });
```
samanthaandrews, line 83: Ensuring loans requests aren't being cached by Cypress

cypress/fixtures/open-ebooks/all-access/loans.html
```diff
@@ -1,478 +0,0 @@
-<feed xmlns="http://www.w3.org/2005/Atom" xmlns:app="http://www.w3.org/2007/app" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:opds="http://opds-spec.org/2010/catalog" xmlns:opf="http://www.idpf.org/2007/opf" xmlns:drm="http://librarysimplified.org/terms/drm" xmlns:schema="http://schema.org/" xmlns:simplified="http://librarysimplified.org/terms/" xmlns:bibframe="http://bibframe.org/vocab/" xmlns:bib="http://bib.schema.org/" xmlns:opensearch="http://a9.com/-/spec/opensearch/1.1/" xmlns:lcp="http://readium.org/lcp-specs/ns">
-  <id>https://qa-circulation.openebooks.us/USOEI/loans/</id>
```
samanthaandrews, line 2: Removed the loans fixture to ensure we are always retrieving the latest loans from the server

cypress/tests/open-ebooks/server-e2e/borrow-and-return/borrow-and-return.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    /** i can confirm that this request is indeed returning the updated loans
+     * but the UI is not changing as it should
+     */
+    cy.wait("@loans");
+
+    cy.log("verify borrow button is visible and click to borrow again");
+    cy.findByRole("button", { name: "Borrow this book" })
+      .should("be.visible")
+      .click();
```
samanthaandrews, line 57: This is where the Cypress test is failing. I can confirm that loans are returning the latest from the server, but UI is not displaying the "Borrow this book button"

src/dataflow/fetch.ts
samanthaandrews: I implemented this change 👍

src/components/__tests__/BorrowOrReserve.test.tsx
```diff
@@ -142,17 +155,3 @@ test("catches unrecognized fetch errors", async () => {
...
-    name: "Borrow this book"
-  });
-
-  mockedFetchBook.mockResolvedValueOnce(fixtures.fulfillableBook);
-  expect(fixtures.mockSetBook).toHaveBeenCalledTimes(0);
-
-  userEvent.click(button);
-
-  await waitFor(() => expect(fixtures.mockSetBook).toHaveBeenCalledTimes(1));
-});
```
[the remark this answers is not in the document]
samanthaandrews: Re-enabled this test

src/hooks/useBorrow.ts
```diff
@@ -25,27 +43,52 @@ export default function useBorrow(isBorrow: boolean) {
...
     try {
-      const book = await fetchBook(url, catalogUrl, token);
-      setBook(book);
+      const book = await fetchBook(borrowUrl, catalogUrl, {
+        method: "PUT",
+        token
+      });
+      setLoading(false);
+      if (loans) {
+        refetchLoans([...loans, book]);
```
samanthaandrews, line 52: Changed this to optimistically set the latest data. This is working with the addition of @kristojorg 's PR that cancels inflight loans requests: https://github.com/NYPL-Simplified/circulation-patron-web/pull/312

samanthaandrews:
Ok, I admit I have sort of gone into a rabbit hole updating types and tests after updating the useSWR package and the cascading effects of that. However, it relates to the work I'm doing to update packages, so not all is lost.

Currently there are three outstanding items that I'd like your help on @kristojorg :
1. Snapshot tests are failing because I updated `emotion` to the latest version. Looking at them, I _think_ it's ok to update the snapshots, as it doesn't seem like anything major was changed except for the way emotion is expressing some of their styles. Can you look at those failing snapshots and let me know what you think?
2. I wrote a Cypress test `cypress/borrow-and-return/borrow-and-return.ts` that is not updating the UI as expected after returning a book. I have confirmed that the loans are being re-fetched and returning the correct data, so it doesn't seem to be a caching issue.
3. As you can see, my Vercel deployments are failing (`circulation-patron-web` and `open-ebooks-web`) but I'm not sure why and I can't inspect into the build.

Maybe we can sync after standup to pair on these remaining items?

kristojorg:
If you run the build locally you'll see the issue is from `next-pwa`... and this is the corresponding ticket: https://github.com/shadowwalker/next-pwa/issues/198 . Not sure we want to upgrade webpack in this PR so maybe we try installing `terser-webpack-plugin` directly?

kristojorg:
Ok I:
- removed the snapshot serializer we had in test-utils/index because this duplicated them and caused errors.
- updated snapshots (it looks like mainly just ordering of classnames changed when updating emotion, no actual styles)
- removed `next-pwa` and created https://jira.nypl.org/browse/OE-291 to track re-enabling it. We don't really need it for now and it's not going to work until we upgrade to next 11/12 and webpack 5.

kristojorg:
@samanthaandrews it appears that in the process of this update, the loading spinner has gotten much larger. Something probably changed with its API that we need to update

kristojorg:
@samanthaandrews nvm it wasn't a theme-ui update, it was in our code. I updating `<LoadingIndicator>` to pass props properly. Should be correctly styled now.

I also made an update to fix the bug where you return a book and weren't seeing the borrow button. This happens because we effectively have _two_ places the book is cached now: in the `/loans` cache and in the cache for the `bookUrl`. When a book is borrowed, the cache looks like this:
```tsx
{
  '.../loans': [{url: 'book1-url.com`, ... }],
  'book1-url.com': {...book},
}
```
When we returned a book, we removed it from the `/loans` cache, but did not mutate the `book1-url.com` cache. Thus the book with fulfillment links was still in there, and we saw the loaned version still. I updated the code to _also_ mutate the `book1-url.com` cache with the value returned from our revoke request. 

In case you're wondering how this suddenly started happening, we didn't previously send a `token` with requests for collections or books, only for `/loans` requests. This means we always got back the _public_ book in a book request, and thus the only loaned version would be found in our `/loans` cache. Once we added the `token` to the book request, we got back loaned books at that url, and now we need to mutate them when we alter the book, along with our loans.

samanthaandrews:
> I also made an update to fix the bug where you return a book and weren't seeing the borrow button. This happens because we effectively have _two_ places the book is cached now: in the `/loans` cache and in the cache for the `bookUrl`. When a book is borrowed, the cache looks like this:
> 
> ```tsx
> {
>   '.../loans': [{url: 'book1-url.com`, ... }],
>   'book1-url.com': {...book},
> }
> ```
> 
> When we returned a book, we removed it from the `/loans` cache, but did not mutate the `book1-url.com` cache. Thus the book with fulfillment links was still in there, and we saw the loaned version still. I updated the code to _also_ mutate the `book1-url.com` cache with the value returned from our revoke request.
> 
> In case you're wondering how this suddenly started happening, we didn't previously send a `token` with requests for collections or books, only for `/loans` requests. This means we always got back the _public_ book in a book request, and thus the only loaned version would be found in our `/loans` cache. Once we added the `token` to the book request, we got back loaned books at that url, and now we need to mutate them when we alter the book, along with our loans.
> 
> I also fixed some jests tests that broke due to swr update

This makes sense. I'm wondering if you saw this bug happen in a specific UI? I don't think I had seen it affecting users before but I was testing things mostly at the book detail view.

I am going to try one last hail mary with the Cypress test. Will update you here.

samanthaandrews:
Linking @kristojorg 's Slack thread related to this issue: https://nypl.slack.com/archives/C0153AFA1RS/p1638896983228900

We identified that we need to modify the server-side response headers in order to fix this bug and the associated Cypress test. For now, I made the `borrow-and-return.ts` Cypress test reload the page after clicking the "Return" button in order to get the test to work as expected, but included a comment that this line should be removed once the response headers have been updated.

@kristojorg do you think this is ok or should I comment out that test entirely? I think we should merge this since it includes other good updates and cleanup, and assign the OE-257 to a backend engineer once we receive feedback from that team in your thread.

kristojorg:
@samanthaandrews yea I think this is the right move. I'll give this a full review today and then we can merge it.

kristojorg (approved):
Looks good!

src/components/BookList.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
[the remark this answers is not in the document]
samanthaandrews: Not even sure how that got there. Good catch! Removed.

cypress/support/commands.ts
samanthaandrews: Yeah I think you're right. I removed this.

cypress/support/commands.ts
```diff
@@ -106,20 +119,23 @@ Cypress.Commands.add("checkoutAllAccessBook", async (requestUrl: string) => {
...
 Cypress.Commands.add("openLoanedEPub", (url: string) => {
-  cy.intercept("GET", `${SERVER_URL}/loans`, {
-    statusCode: 200,
-    fixture: "open-ebooks/all-access/loans.html"
+  cy.intercept(`${SERVER_URL}/loans`, { middleware: true }, req => {
+    delete req.headers["if-none-match"];
+    req.on("before:response", res => {
+      // force all API responses to not be cached
+      res.headers["cache-control"] = "no-store, no-cache";
+    });
```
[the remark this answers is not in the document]
samanthaandrews: yeah i think this is a good idea. i'm going to make a separate PR for this so that it will be easier to review. created ticket https://jira.nypl.org/browse/OE-298 and assigned it to me so it will be tracked.

PR #306, opened 2021-11-16
update epub-to-webpub to 1.0.0

Toxiapo:
This PR updates the` @nypl/epub-to-webpub` package to v1.0.0 which fixes the cover page error due to a missing resource link in the readingOrder source.


kristojorg (approved):
wait(3000)

PR #308, opened 2021-11-17
Cache webpub manifest URL with Vercel

Toxiapo:
This PR removes the valutId query from our API endpoint that was being used to fetch webpub manifests, making this change allow Vercel to cache the API response in their server.

Previously, the server could not cache this endpoint due to the fact the valutId is different per patron. An HTTP request to this endpoint with the same ISBN will now return a cached version of the response instead of making another request to the server, which makes this a lot faster on book rendering!


kristojorg (changes requested):
Looks great! Super happy this works. Still have a couple other things we need to do to get load times down, but this is a great step for us! 

Just a couple small edits requested below

src/pages/api/axisnow/[isbn].ts
```diff
@@ -10,7 +10,8 @@ import track from "analytics/track";
  * eg: http://localhost:3000/api/axisnow/[isbn]/[vaultId]
```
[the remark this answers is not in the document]
Comment (Toxiapo) [reply src/pages/api/axisnow/[isbn].ts]: Updated!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Fix: Update read and open book Cypress tests
 - Fix: Update web-reader which brings numerous changes
 - Fix: Update epub-to-webpub to fail gracefull when books are missing TOC information.
+- Refactor: Change API `axisnow/[isbn]/[vaultId]` to `axisnow/[isbn]` so that Vercel could cache our webpub manfiest responses.
```
kristojorg, line 8: FWIW I think this should be `Perf: ` since it's not just a refactor

src/dataflow/fetch.ts
```diff
@@ -19,6 +19,22 @@ export default async function fetchWithHeaders(url: string, token?: string) {
   }
 }
 
+export async function fetchWithValudId(url: string, valutId: string) {
```
kristojorg, line 22: Spelling in function name and argument name, and can we also add a comment to describe what this function is for

kristojorg (changes requested):
just the one thing, but otherwise awesome!

We _could_ add a test to make sure that we properly call this API endpoint on the read page.

kristojorg (approved):
Looks good, LGTM!

PR #309, opened 2021-11-18
Disable publication service worker

kristojorg:
Our web-reader service worker is flawed and causes a perf problem (and flaky cypress tests). I created a ticket to address this and write it some other way, but for now we should disable it.


Toxiapo (approved):
LGTM!

cypress/tests/open-ebooks/server-e2e/web-reader-v2/read-books.ts
```diff
@@ -39,7 +39,6 @@ describe("Users can read/navigate loaned books", () => {
 
     cy.findByText("Paginated").click();
     cy.findByRole("button", { name: "Settings" }).click();
-    cy.findByRole("button", { name: "Next Page" }).click();
```
Toxiapo, line 42: Removing this because it creates a race condition with line [L45](https://github.com/NYPL-Simplified/circulation-patron-web/blob/d7c3770d1cc4091b6a3d58ab1cbaef24a2cce744/cypress/tests/open-ebooks/server-e2e/web-reader-v2/read-books.ts#L45) where it also triggers a re-render of the iframe.

samanthaandrews (approved):
🚀

PR #312, opened 2021-11-23
Cancel inflight loans requests

kristojorg:
This PR updates SWR, and makes use of new [middleware](https://swr.vercel.app/docs/middleware#keep-previous-result) to add a `cancelInflight` method to the SWR hook, which we now use to cancel inflight loans requests when we make a fulfillment request (borrow, reserve, cancel, revoke). This avoids the edge case where the pending revalidation returns with incorrect data when users click buttons too quickly. You can try it and see the requests being cancelled in the network tab.

I also now only show the loading indicators in place of fulfillment buttons when loading the _first_ user data, not after that. I do it when showing the first because I guess you could get into a weird state where we load the page and the first loans request is pending, and we you borrow button, which you click. This cancels the first pending loans request and might give you an error that you already have the book checked out. I'm not sure if this is the right UX though, as seeing lots of spinners sucks and it's going to happen on every first load, while this edge case is pretty unlikely (you have to click a book you already loaned before loans is returned on first load). Another way to mitigate it is to _not_ cancel the loans request if there is not yet any loans data (as in, it's the first request). This could get you back into the first situation though... Finally, we could also try just disabling the buttons until we have the first loans data. 

Let me know what you think


samanthaandrews (approved):
Did some testing and this works great! I rebased my branch and fixed some tests and type errors, but had to log off before finishing. I'm ok with merging into my branch 👍

PR #314, opened 2021-12-03
Fix: Use URL.createObjectURL for webpubmanifesturl in axisnow book

kristojorg:
This PR updates to the beta AxisNow decryptor, which uses the new AxisNow API domain. I will make this PR ready for review once we have this https://github.com/NYPL-Simplified-Packages/axisnow-access-control-web/pull/9 merged and published. 

I also surreptitiously discovered a bug in the web reader. Thanks to @Toxiapo 's work, we now have a cacheable WebpubManifest request. That's awesome, but it means that we had to send a custom header when fetching the manifest. However, the `WebReader` also fetches the manifest, and doesn't know about the custom header, so it would get an error. Normally this was avoided because the manifest would already be cached by the time the WebReader fetches it, but not always. To get around this, I use `URL.createObjectURL` to make a custom URL for our already fetched manifest, and then I pass that to the web reader.


kristojorg:
Closing this due to Leonard's comment that the old axisnow domain is back to normal.

kristojorg:
Wait we need to merge the fix I also discovered here.

samanthaandrews (approved):

Toxiapo (approved):
Nice, thanks for fixing this! TIL `URL.createObjectURL`

PR #318, opened 2021-12-09
Parallelize cypress

Toxiapo:
There's seems to be a problem with caching on the deployment_status tag, which is what we use for the cypress test, unfortunately, it's currently not part of the supported event for [actions/cache@v2](https://github.com/actions/cache/blob/bd492913657217c4685db3a9cac721bca7d62caa/src/constants.ts#L17).

This PR contains the following changes:

- Set up 6 containers to run cypress test
- Fix the unzip path error and comment out the cache option.
- Remove unnecessary artifact uploads and shorten the retention period  of each upload (Could be set in the settings as well)
- Remove mobile tests


.github/workflows/oe-integration-cypress.yml
```diff
@@ -21,49 +73,23 @@ jobs:
...
-          CYPRESS_OPENEBOOKS_CLEVER_SCHOOL_NAME: ${{ secrets.CYPRESS_OPENEBOOKS_CLEVER_SCHOOL_NAME }}
-          CYPRESS_OPENEBOOKS_CLEVER_HIGH_SCHOOL_USER_USERNAME: ${{ secrets.CYPRESS_OPENEBOOKS_CLEVER_HIGH_SCHOOL_USER_USERNAME }}
-          CYPRESS_OPENEBOOKS_CLEVER_HIGH_SCHOOL_USER_PW: ${{ secrets.CYPRESS_OPENEBOOKS_CLEVER_HIGH_SCHOOL_USER_PW }}
-          CYPRESS_OPENEBOOKS_CLEVER_HIGH_SCHOOL_ACCESS_TOKEN: ${{ secrets.CYPRESS_OPENEBOOKS_CLEVER_HIGH_SCHOOL_ACCESS_TOKEN }}
-          # Recommended: pass the GitHub token lets this action correctly
-          # determine the unique run id necessary to re-run the checks
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          path: ~/.cache/Cypress
+          key: cypress-cache-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
+          restore-keys: cypress-cache-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
```
Toxiapo, line 88: This seems to be the why we are getting the 'cypress binary not found' error, because we never restore the cached

.github/workflows/oe-integration-cypress.yml
```diff
@@ -3,12 +3,64 @@ name: Open Ebooks Integration with Cypress
...
-    container: cypress/browsers:node14.17.0-chrome88-ff89
+    needs: install
+    container:
+      image: cypress/browsers:node14.17.0-chrome88-ff89
+      options: --user 1001 # https://github.com/cypress-io/github-action#firefox
+    strategy:
+      fail-fast: false
+      matrix:
+        # run copies of the current job in parallel
+        containers: [1, 2, 3, 4, 5, 6]
```
Toxiapo, line 63: I downgraded the number of containers to 6 because I think that's the max you can run before you have to pay for it.

.github/workflows/oe-integration-cypress.yml
```diff
@@ -3,12 +3,64 @@ name: Open Ebooks Integration with Cypress
...
+        with:
+          path: ~/.cache/Cypress
+          key: cypress-cache-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
+          restore-keys: cypress-cache-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
+
+      - name: Install Cypress 📥
+        run: npm i cypress
+
+      - name: Zip node_modules
+        run: zip node_modules.zip ./node_modules/* -r
```
Toxiapo, line 40: @kristojorg Maybe we can remove compressing the node_modules folder here, I think we can just restore the cached version from line 26.

Toxiapo:
@kristojorg uploading the cypress binary doesn't seem to work, for some reason, cypress couldn't find the path of the binary location even when I explicitly label the `CYPRESS_CACHE_FOLDER` 

For now, I am only uploading the node_module and installing cypress on each container, which still gives a good boost in test run time.  

We could open up a ticket to further investigate this and also follow the caching package when they support deployment events.

Screen Shot 2021-12-15 at 6 12 58 PM

.github/workflows/oe-integration-cypress.yml
kristojorg: Is this still relevant?

.github/workflows/oe-integration-cypress.yml
kristojorg: Ok nice thanks

kristojorg (approved):
Looks pretty good to me, although cypress dashboard looks like a test is failing?

Toxiapo:
> Looks great to me, although cypress dashboard looks like a test is failing?

Seems like the test is causing a login issue, I think this s a new test fail? "You need to be signed in to view this page"
I will try to solve this now.

samanthaandrews:
I'm wondering if we even need the mobile test suite. They run the exact same tests as the desktop tests and don't seem to be adding much value.

Toxiapo:
> I'm wondering if we even need the mobile test suite. They run the exact same tests as the desktop tests and don't seem to be adding much value.

That's a good point, I agree it does seem redundant, and it's giving more works to the server than it needs to be.

Toxiapo:
@kristojorg The Cypress tests that run on each PRs have been failing due to 502 errors and slow server response, @samanthaandrews and I was wondering if this is necessary due to the fact that we already have a daily cron job that runs through the same tests.

The idea is to have more integration test with fixtures and studs and only write a few e2e tests that test the most important app functionality

For now, I am disabling the e2e tests and only run the integration tests in this PR. A new ticket is created to follow this issue. [OE-313](https://jira.nypl.org/browse/OE-313)

samanthaandrews:
This is great work 👍

PR #320, opened 2021-12-14
OE-299: Remove unused Github Actions workflow files

samanthaandrews:
We aren't using these workflows since we have been deploying with Vercel so these can be removed.


kristojorg (approved):
Does this also include the other firstbook login changes? Either way, looks ok to me

PR #321, opened 2021-12-14
OE-298: Remove caching stubs from Cypress tests

samanthaandrews:
- After some investigation, I don't think we actually need to add any custom cache-control behaviors to our Cypress tests. The issue I was trying to mitigate was the [issue](https://nypl.slack.com/archives/C0153AFA1RS/p1638896983228900) Kristo identified in the way we are fetching books and loans.
- I also noticed some flake in our Firstbook Login test and made a few small changes that seem to fix the issue.


kristojorg (approved):
Nice, looks good!

PR #325, opened 2021-12-15
upgrading webreader to 0.3.3

Toxiapo:
Updates the web-reader to 0.3.3 which includes a fix where internal links were not able to navigate users. [OE-227](https://jira.nypl.org/browse/OE-227)


kristojorg (approved):
Looks good to me!

PR #326, opened 2021-12-16
OE_280 Auth Middleware

kristojorg:
This PR uses Next 12 middleware to check auth status for various pages before allowing the user the view them. This allows us to pre-render all collection and book detail routes at build time with their "public" versions that don't contain user-specific info. This means the browser first loads plain HTML with the public content and then requests the patron-specific info. I had to make some changes in order to make this happen. It's a pretty big optimization though! Everything should feel much faster.

## List of Changes

- Statically generate all collection and book details pages by making requests for the content without a user token to get the shared/public version. We then ensure authentication in middleware before users get to these pages.
- Make sure auth cookie is sent to the server.
- On `/book` pages, check if the user is logged in via middleware by checking the `/patrons/me` endpoint with their token. If not, redirect them to login. If we get a 200, continue the request for the book page.
- On `/collection` pages, make a HEAD request for the collection with the users auth token. If they are not logged in, redirect to login. If they are logged in, but can't access that collection, redirect to the home page. If we get a 200, continue the request for the collection page.
- On `/loans` and `/read` don't run any middleware because that is client-side rendered anyways since there is no shared version of that page.
- On `/app` we previously just fetched the root lane and let the server redirect us to the patron's root lane, but leave them on `/app`. Now, instead, we don't show anything at `/app`, we make a server-side fetch for the catalog root url, get the proper redirected patron root lane, and send them directly there so they land on `/app/collection/[collectionUrl]`. 
- The `/login` page now has middleware that redirects the user if they're already logged in. If there is a login error, it adds the `?loginError` query parameter so it is displayed in the UI. If the user is not logged in and is going to end up on the login page, it removes their auth cookie.
- Make all statically generated pages [blocking](https://nextjs.org/docs/basic-features/data-fetching#fallback-blocking) because our "fallback" page was just a full page loading spinner and looked bad. 
- Add a new top of the page loading indicator.
- I disabled the worker decryptor and am using the workerless decryptor for now. This might not even result in much of a performance hit, it's worth measuring. I had to do this because of vercel/next.js#32590
- I made login errors display better in the login page.
- I removed `AuthProtectedRoute` as this is a client-side auth protection that we no longer use. We simply redirect the user now via the middleware when they try to visit pages that they're not authenticated for.
- We no longer fetch collections on the client-side using `useSWR`. This is because no collection data is per-patron specific. It is all shared information, so we can just fetch it when rendering the page, and then don't need to fetch it again. The same is true of the "search data" used to configure the search bar for that collection.
- We no longer use [shallow routing](https://nextjs.org/docs/routing/shallow-routing). We used this previously because the static props fetched on each page were just the app props, no page-specific information. That's no longer the case. Each page has its own data, so we need to go back to the next.js server and get that data whenever we change pages. For example when changing from one collection to another, it used to be that the static data stayed the same (it was just app props), and the client would fetch the collection. Now, though, the collection is fetched server-side and cached there, so we need to get that fresh data when we change the collection url.

### 502 Errors

We are continuing to get Clever 502 errors. One side effect of the changes here is that 502 errors can occur in middleware, which will cause a redirect to `/login`, which will ultimately remove the auth cookie. So the user will see the login page (with the 502 error visible) and have to log in again. They _should_ just have to click "login with clever" again without putting their username and password in again. 

Anyway, the way to make this better is to make a way for users to "retry" or something. Not sure how that would work yet, as these errors can occur in numerous places.

### Requesting `/app` is slow

Requesting the catalog home page currently takes the longest out of any page (except the web reader). It's marginally slower than it was previously. This is because it has to perform multiple fetches in order to determine where to redirect the user to:

- Fetch the authentication document to get the library's catalog root url
- Fetch the catalog root url to get the patron's root lane url. 
- Redirect the patron to their root lane collection under `/app/collection/[collectionUrl]`.

Some ways we could make this faster would be:

- Remove the multi-library setup, and just fetch the authentication document at build time so we already have it, like we currently do for the config file.
- Include the patron's root lane in the response to `/patrons/me` (or some other lightweight endpoint) so we can get that in a single, fast request.


src/components/context/UserContext.tsx
```diff
@@ -78,6 +80,7 @@ export const UserProvider: React.FC = ({ children }) => {
   function signOut() {
     clearCredentials();
     mutate();
+    router.reload();
```
kristojorg, line 83: This is just good practice, but it also gives the middleware a chance to properly redirect us

src/dataflow/opds1/parse.ts
```diff
@@ -28,7 +28,15 @@ import {
...
-import DOMPurify from "dompurify";
+import createDomPurify from "dompurify";
+
+let DOMPurify: any;
+if (typeof window !== "undefined") {
+  DOMPurify = createDomPurify();
+} else {
+  const { JSDOM } = require("jsdom");
+  DOMPurify = createDomPurify(new JSDOM().window);
+}
```
kristojorg, line 39: DOMPurify needs a window object, which we don't have on the server. So we use JSDOM on server.

src/dataflow/withAppProps.ts
```diff
@@ -15,8 +15,18 @@ export type AppProps = {
...
-  pageGetStaticProps?: GetStaticProps,
+type PageGetStaticProps<
+  PageProps extends { [key: string]: any } = { [key: string]: any }
+> = (
+  ctx: GetStaticPropsContext<ParsedUrlQuery>,
+  librarySlug: string,
+  library: LibraryData
+) => Promise<PageProps>;
+
+export default function withAppProps<
```
kristojorg, line 26: This function gets the site-wide data, but it also can support getting page-specific data. We never used it before, but I wrote it with this in mind. I updated it a little though to make it work better.

src/hooks/useCollection.ts
```diff
@@ -21,7 +21,9 @@ export default function useCollection() {
   const { data: collection, error, isValidating } = useSWR<
     CollectionData,
     Error | ApplicationError
-  >(collectionUrl ? [collectionUrl, token] : null, fetchCollection);
+  >(collectionUrl ? [collectionUrl, token] : null, fetchCollection, {
+    fallbackData: collectionSeed
```
kristojorg, line 25: We pass the statically-fetched data to `useSWR` and it uses that for the first render while revalidating in the background

src/pages/[library]/index.tsx
```diff
@@ -3,20 +3,77 @@ import Collection from "components/Collection";
...
+     * @todo - update parser to not used undefined values
+     */
+    const collection = JSON.parse(
+      JSON.stringify(
+        await fetchCollection(
+          collectionUrl,
+          generateCredentials(username, password)
+        )
+      )
+    );
```
Comment (kristojorg) [inline src/pages/[library]/index.tsx:62]: I should handle errors here, especially 502 errors. Probably by just not sending the page data along and letting it fallback to the way it was before.

kristojorg:
In order for this actually to work:
- collections _do_ change per-patron. So we would need middleware to redirect a patron to the right root collection also. 
- we would need to check if a patron has access to a collection in middleware, ideally without fetching the entire collection.
- we would need a way to fetch collections publicly at build time, like with a client-only universal token

kristojorg:
Update from convo with leonard:
- We can make a HEAD request to a collection to determine if a user can access it.
- Only the root lane changes per patron. We can fetch the root lane for the user and get the redirect that the server sends and redirect the user there.
- There would be no such thing as `/app` then. You would always get redirected to your specific root lane.

kristojorg:
Todo:
- Don't check patron/me if you're on collection page. It's not necessary since auth is checked via collection url
- Resolve lane issue with useLayoutEffect
- Dig in to infinite redirect issue

kristojorg:
I am seeing an issue (https://github.com/vercel/next.js/issues/32590) when trying to build for production, but otherwise this is looking promising. It works in dev!

samanthaandrews:
This is a big improvement! 🎉

The biggest issue I noticed while doing manual testing (this is also showing up in Cypress tests) is that sometimes when I click on a book or navigate to a collection, I am redirected to the login screen where it shows the loading spinner and then I am redirected back to page I was previously on.

https://user-images.githubusercontent.com/34348332/147005783-b074f073-11af-4d38-808f-9376a244e70a.mov

Perhaps related, I am sometimes getting an indefinite loading spinner when navigating to the `/loans` page. When I do a hard refresh, the issue seems to be resolved and I can see my loans.

https://user-images.githubusercontent.com/34348332/147005861-097cff8e-b9b7-4a4e-86a9-276ba017bfae.mov

cypress/tests/open-ebooks/server-e2e/login/clever-login.ts
```diff
@@ -64,12 +65,8 @@ describe("Clever login", () => {
...
-    cy.contains("h1", "Open eBooks (QA Server) Home", { timeout: 30000 });
+    cy.contains("h1", "High School", { timeout: 30000 });
     cy.findByRole("heading", { name: "Patron Support" }).should("exist");
-    cy.location("pathname").should("eq", APP_PATH);
-    cy.findByRole("listitem", { name: "Current location: High School" }).should(
-      "exist"
-    );
   });
 
   it("Incorrect Clever token displays error", () => {
```
samanthaandrews, line 72: This test has been failing and I think it's catching an actual issue. It's checking that if this query param `?loginError=` exists in the url, we display the error to the user in the UI.

Clever login -- Incorrect Clever token displays error (failed) (attempt 2)

src/hooks/useMounted.ts
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+export default function useMounted() {
```
Toxiapo, line 3: I am wondering how is this being used. Does it work as a force re-render?

samanthaandrews:
An update on Cypress tests: We are getting 502 BAD GATEWAY errors on almost every Cypress test that run in CI. These 502 errors come from the /loans endpoint, `https://qa-circulation.openebooks.us/USOEI/loans`

I also occasionally see these errors when doing manual testing, but the CI runs into them much more frequently.

```
detail: "Error in authentication service"
 status: 502
 title: "Third-party service failed."
 type: "http://librarysimplified.org/terms/problem/remote-integration-failed"
```

I've inspected the request and response headers on a successful 200 /loans request and an unsuccessful 502 /loans request. We seem to be sending the same request headers from the app, but the response headers are different between a 200 and 502 status. In a 200 status code response header, we see the presence of two headers that don't exist on a 502 status response header,  `cache-control: private, no-cache` and `last-modified: Wed, 22 Dec 2021 21:05:02 GMT`

### 502 status code response headers:
1. **Connection:** keep-alive
2. **Content-Encoding:** gzip
3. **content-type:** application/api-problem+json
4. **date:** Wed, 22 Dec 2021 21:15:26 GMT
5. **Keep-Alive:** timeout=5
6. **server:** nginx/1.14.0 (Ubuntu)
7. **Set-Cookie:** AWSALB=E6PN/fC7gj+OJb4Z7I3mjq3dSeTTFwSYfZiP/wwuAL9Sn4VlkkRWbssdrrHKEDy4WUgVlswV5Zbetm2r8gpyDdiLkogUVPt3CptM4HJQgrTccQhuGxaIHGs67jpg; Expires=Wed, 29 Dec 2021 21:15:26 GMT; Path=/
8. **Set-Cookie:** AWSALBCORS=E6PN/fC7gj+OJb4Z7I3mjq3dSeTTFwSYfZiP/wwuAL9Sn4VlkkRWbssdrrHKEDy4WUgVlswV5Zbetm2r8gpyDdiLkogUVPt3CptM4HJQgrTccQhuGxaIHGs67jpg; Expires=Wed, 29 Dec 2021 21:15:26 GMT; Path=/; SameSite=None; Secure
9. **Transfer-Encoding:** chunked
10. **Vary:** Accept-Encoding

### 200 status code response headers
1. **cache-control:** private, no-cache
2. **Connection:** keep-alive
3. **content-encoding:** gzip
4. **content-type:** application/atom+xml;profile=opds-catalog;kind=acquisition
5. **date:** Wed, 22 Dec 2021 21:15:53 GMT
6. **Keep-Alive:** timeout=5
7. **last-modified:** Wed, 22 Dec 2021 21:05:02 GMT
8. **server:** nginx/1.14.0 (Ubuntu)
9. **Set-Cookie:** AWSALB=kvVQkRzWkuyF6RsSZzZqR0QVEbEaEdKZuiUG2gptjlwqkwS4p7vcD6/YYwBF/XK++Hfp6aaSzkNcUBR4wFeT7urG8lTCdL86M/iwN3paQNGCAdUJ03jOv4U72eh/; Expires=Wed, 29 Dec 2021 21:15:53 GMT; Path=/
10. **Set-Cookie:** AWSALBCORS=kvVQkRzWkuyF6RsSZzZqR0QVEbEaEdKZuiUG2gptjlwqkwS4p7vcD6/YYwBF/XK++Hfp6aaSzkNcUBR4wFeT7urG8lTCdL86M/iwN3paQNGCAdUJ03jOv4U72eh/; Expires=Wed, 29 Dec 2021 21:15:53 GMT; Path=/; SameSite=None; Secure
11. **Transfer-Encoding:** chunked
12. **Vary:** Authorization, Accept-Encoding

src/hooks/useMounted.ts
kristojorg: Good question. This is being used because we need the server-side render and the _first_ client-side render to result in the same HTML in order for react's hydration function to work properly. What was happening was that the render would result in something ever so slightly different depending if you were on the server or the client. For example, in `useLogin`, we had a line here: https://github.com/NYPL-Simplified/circulation-patron-web/blob/99671fcfa8b183daa01586c7d6c129afcff22e49/src/auth/useLogin.ts#L34 that would change the `nextUrl` query parameter of the login button depending if we were server side or client side. This was due to a detail with how those the `router`'s `asPath` worked. Anyways, it meant that the login button on the server had no `nextUrl` and then it did in the first render on the client, which caused a hydration mismatch and forced react to throw out the whole dom and replace it. Instead, we now use `useMounted` to know if it is the first render or not, and only set the `nextUrl` on the second render and beyond. 

Thinking about it, another way we could handle this is to add some logic that makes the `asPath` the same on both client and server. Then we wouldn't have to worry about this.

kristojorg:
Ok so some notes for myself:
- [ ] when credentials are invalid and that is caught in the middleware, we need to clear them from the client. We can do this either by sending a set-cookie from the middleware during login redirect, or in the client upon reaching login page. Probably best done in the server if possible. Would also be good to send the error message from the server to the client in a query param so it can be displayed. This will stop the infinite login page redirection (hopefully it's the only cause).
- [ ] Need to make `loginError` display properly in the login page. This is not happening because login is statically generated and doesn't account for the query string currently. We need to somehow account for it in our static generation.
- [ ] Figure out infinite loans loading
- [ ] Figure out random login redirect when clicking book...

kristojorg:
@samanthaandrews let me know if you're seeing any more of those errors. Also @samanthaandrews @Toxiapo can you check for either of these two things I've seen in the past but not recently:

- Infinite loading screen on the `/loans` page (or anywhere for that matter)
- Randomly being redirected to login (you will correctly see this however if we get a 502 error in the middleware)
- The root lane being flashed briefly when navigating (or any page flashing one thing for a second then transitioning to the correct content briefly after)

kristojorg:
@samanthaandrews @Toxiapo I'm moving this back to a draft as I've found some issues... : (

kristojorg:
@samanthaandrews @Toxiapo this is back to ready for review. I think I've resolved the errors I was seeing and the ones sam found. It feels pretty stable to me now in my testing (and pretty fast : ) ). LMK if you find any issues here!

samanthaandrews:
Ok two things, one of which I think you mentioned seeing before in a previous comment.

It seems I can sometimes get stuck when clicking on a book link. I see the loading bar at the top, but it never reaches "100%" and I have to reload the page in order to get unstuck. I saw this happen a couple times in my testing.

https://user-images.githubusercontent.com/34348332/147587737-b40406bf-500a-4b28-8152-700d6f877226.mov

Even just after successfully logging in, I will randomly get redirected to the login page with the "Error in authentication service" error message. This seems to consistently happen when clicking a book link as shown in the video below. Perhaps this issue and the other one I just mentioned are related?

https://user-images.githubusercontent.com/34348332/147587865-38e17e57-ae97-46fd-b488-7a3fc0cab460.mov

Toxiapo:
> Ok two things, one of which I think you mentioned seeing before in a previous comment.
> 
> It seems I can sometimes get stuck when clicking on a book link. I see the loading bar at the top, but it never reaches "100%" and I have to reload the page in order to get unstuck. I saw this happen a couple times in my testing.
> 
>  bug-on-book-link.mov 
> Even just after successfully logging in, I will randomly get redirected to the login page with the "Error in authentication service" error message. This seems to consistently happen when clicking a book link as shown in the video below. Perhaps this issue and the other one I just mentioned are related?
> 
>  Screen.Recording.2021-12-28.at.9.30.13.AM.mov
@samanthaandrews That's really weird, I can't reproduce these errors. Does this only happen with Clever login or both?

samanthaandrews:
I've only tested with FirstBook login and using this preview url: https://open-ebooks-web-git-oe280-auth-middleware-nypl.vercel.app/

Toxiapo:
> I've only tested with FirstBook login and using this preview url: https://open-ebooks-web-git-oe280-auth-middleware-nypl.vercel.app/

Oh yeah, I can reproduce both the bugs with this link but not with the latest [preview](https://open-ebooks-web-95rl301md-nypl.vercel.app/).

samanthaandrews:
I was under the impression that the initial preview url updated with each commit! Am I wrong?

I did some initial testing with the url that @Toxiapo posted, and have still run into the issue with hitting the third party authentication error when clicking a book. It doesn't happen with every book, but once I've clicked around enough I eventually hit it.

kristojorg:
> Even just after successfully logging in, I will randomly get redirected to the login page with the "Error in authentication service" error message. This seems to consistently happen when clicking a book link as shown in the video below. Perhaps this issue and the other one I just mentioned are related?

This is the auth error we are waiting for the Circ Manager oAuth work to resolve. We should try to find a better way to handle it, but I think that's for a separate ticket. This isn't a new error originating from this code, we are just checking auth status a bit more frequently now, and thus running in to this error more frequently.

> I was under the impression that the initial preview url updated with each commit! Am I wrong?

This is true, but you might have to do a hard refresh to get the latest update. Do you mind trying with a hard refresh and letting me know if you still see that bug on book links?

samanthaandrews:
@kristojorg I haven't seen it after hard refresh, just the authentication issue. I definitely think these changes give the app a big performance boost, but dang that auth issue results in a pretty poor user experience! I know the backend team is working on the oauth stuff currently, but we should probably let Miguel know that this issue will arise more frequently with these changes. I think it's ok since the app is still in beta, but definitely crucial to fix before taking it out of beta.

I'll review the code later this evening! Going to be offline for the next few hours.

Another good thing is that since these changes make the app run faster, Cypress tests also seem to be running more quickly 🏃

src/utils/auth.ts
```diff
@@ -40,3 +46,18 @@ export function generateCredentials(username: string, password: string) {
   const btoaStr = btoa(`${username}:${password}`);
   return `Basic ${btoaStr}`;
 }
```
samanthaandrews, line 48: nit: Just realized there is a duplicate function in the code base inside `src/auth` called `generateToken()` that we could probably use here instead of `generateCredentials()`.

samanthaandrews (changes requested):
Just a few very small things! Looks really great. I'm excited about these changes.

I'm thinking we should just merge with the failing Cypress tests for now. I think as an admin you can do that? I'll have a PR soon with the refactored tests.

src/auth/useCredentials.ts
```diff
@@ -19,18 +19,22 @@ export default function useCredentials(slug: string | null) {
   const router = useRouter();
   const [credentialsState, setCredentialsState] = React.useState<
     AuthCredentials | undefined
-  >(getCredentialsCookie(slug));
+  >(undefined);
+  // keep track of whether we have tried to load credentials or not.
+  const [hasLoaded, setHasLoaded] = React.useState(false);
```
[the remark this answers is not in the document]
kristojorg: Done!

src/components/Collection.tsx
```diff
@@ -50,13 +40,11 @@ export const Collection: React.FC<{
       }}
     >
       <Head>
-        <title>{pageTitle}</title>
+        <title>{collection.title}</title>
```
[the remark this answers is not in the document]
kristojorg: Yea they should all have titles, that's part of the OPDS contract. Previously we had this weird page at /app where you were viewing a collection but had a different title. I just removed that.

src/components/context/UserContext.tsx
kristojorg: Yea this is something I noticed too and resolved in my new auth pages PR that follows this one up.

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -166,11 +169,15 @@ test("sign out clears cookies and data", async () => {
   });
 
   // should have removed cookie
-  expect(Cookie.remove).toHaveBeenCalledTimes(1);
-  expect(Cookie.remove).toHaveBeenCalledWith("CPW_AUTH_COOKIE/testlib");
-
-  // data should be revalidated
-  expect(mutateMock).toHaveBeenCalledTimes(1);
+  expect(Cookie.remove).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
kristojorg: Yea I was getting a bug where sometimes a cookie wouldn't be removed, and it seemed to be mitigated by simply calling it twice haha... I can try removing and re-testing though to see if it was some other issue

samanthaandrews (approved):

samanthaandrews:
I will fix these failing remaining Cypress tests now, or comment out the ones that are consistently failing (navigating directly to a book is pretty much always hitting the 502 error, which i will resolve in a subsequent PR).

samanthaandrews:
These tests are all failing due to the 502 third party errors that are preventing navigating around the app as expected. I'm working on mocking the nextjs server so we can mitigate this. In the meantime, I am ok with disabling the Cypress status check in order to merge these changes. Or we can wait for my subsequent PR.

kristojorg:
@samanthaandrews I've moved this to ready for QA and am awaiting a QA approval before I merge, so we can wait on your PR for now.

hokei:
> Update from convo with leonard:
> 
> * We can make a HEAD request to a collection to determine if a user can access it.
> * Only the root lane changes per patron. We can fetch the root lane for the user and get the redirect that the server sends and redirect the user there.
> * There would be no such thing as `/app` then. You would always get redirected to your specific root lane.

Thinking about stuff that might be coming up - if the school system wanted to have a collection for everyone, would that be a thing that CPW isn't concerned with?  AKA it would be reflected in the root lane response that's from the circulation manager?

PR #327, opened 2021-12-20
OE-296: Don't use loans to get fulfillment link

samanthaandrews:
Remove `<AuthProtectedRoute />` component wrapper around the <WebReader /> component to eliminate an expensive check to the `/loans` endpoint. I also edited the error boundary to display the message it received from the server if there is one to display. This makes it easier for users to understand why their book won't display if they try to read it and receive the "No active loan" error.

A couple small cleanup things as well.


samanthaandrews:
As I was working on these Cypress tests, I realized that even with my changes we are still checking the `/loans` endpoint unnecessarily because the <WebReader /> component is wrapped in the UserProvider, which runs the fetchLoans request. This value might be cached? but it's weird to me that this request still shows up in the Network tab whenever we need to access data in the UserProvider context. I'm going to take some time to think about the best way to move forward.

kristojorg:
@samanthaandrews it's expected that this request will always show up whenever we are using the `UserProvider`, because we have to fetch loans in order to get user information provided there. The important thing is that rendering the book should no longer be _blocked_ by the loans request. With that said, it's true we don't actually need user information in this page, so we should just not wrap it in a `UserProvider`. That might require piecing together a different set of wrapping components, but it is probably worth it. 

Also though, another thought. So previously we passed a `fulfillmentUrl` but now we pass just the `bookUrl`, which means we have to fetch the `bookUrl` then the `fulfillmentUrl` then the `webpubManifestUrl` and then the actual resource in order to render the first page. That might be faster than using `loans`, but it's still quite a waterfall of requests. I'm wondering, if we keep the `fulfillmentUrl` as the query parameter, do we properly get a 401 if the user is not authorized for the resource? If so, can we just keep that and avoid the `bookUrl` request? 

Another option would be to make a backend change so we could get fulfillment info in the request to the `bookUrl` (the OPDS Entry). This eliminate one request, but would probably be more work. Let's try just using the `fulfillmentUrl` like we were before first.

samanthaandrews:
Thank you for that explanation, @kristojorg ! I do think keeping the `fulfullmentUrl` would be best. However, I'm noticing that we use the `fulfillmentUrl` and the user `token` to get the rest of the fulfillment info like `isbn` and `book_vault_uuid` to construct the manifest. Which means we still need access to that token, aka our component needs to be wrapped in the user provider in order to get the manifest. Unless there is another way to get the `isbn` and `book_vault_uuid` information that I'm not seeing?

Another option would be to pass the user token as a query param? But that seems maybe like bad security practice...

What if we fetched the fulfillment info earlier and passed the `isbn` and `book_vault_uuid` as query parameters?

kristojorg:
@samanthaandrews good points. I think ultimately the solution here will be to have `UserProvider` not actually fetch the loans, but have a separate `LoansProvider` do that using the token provided by `UserProvider`. In fact, thinking about it, there are basically three pieces of user information we are interested in:
1. Their credentials, stored in a cookie
1. Their auth status (signed in, signed out, etc)
1. Their loans

Each of those could be separated into separate providers that we use differently in different places. Additionally, once the middleware is merged, we won't really need #2, since a user should always be authenticated by the time they are viewing authenticated information. 

Anyway, that's for a follow up ticket. For now, it's okay that we are still fetching `/loans` on the read page so long as the book render is not blocked on it. So I would keep using `UserProvider` and `fulfillmentUrl` for now, if possible

samanthaandrews:
Ok this is ready for review. Confirmed that just by removing the <AuthProtectedRoute /> users will still be blocked from reading books whose loan has expired. I made a small change to the error boundary to ensure a useful error message is displayed to users in this case.

kristojorg (approved):
Looks great, thanks! LGTM

PR #328, opened 2021-12-23
OE-314: Removing enzyme & refactoring react context unit test

samanthaandrews:
@Toxiapo posted an [article](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl) in Slack about how enzyme will no longer be compatible with future versions of React. We were only using enzyme in one unit test, so this PR refactors that test and removes the enzyme library.

https://jira.nypl.org/browse/OE-314


Toxiapo (approved):
Nice, thank you for doing this!

kristojorg (approved):
Nice!

PR #330, opened 2021-12-29
OE-315: New Auth Routes

kristojorg:
Previously, the auth pages were handled on the client side. This moves that logic server-side and creates new pages dedicated to each auth method: 
- `/[library]/login/method/clever` - simply redirects to clever in `getServerSideProps`
- `/[library]/login/method/basic` - shows basic auth form
- `/[library]/login/method/saml/[id]` - redirects to the given SAML provider. This one is untested but I moved it over because we used to use this with the SimplyE setup.


src/auth/useLogin.ts
```diff
@@ -5,40 +5,39 @@ import { UrlObject } from "url";
 import { IS_SERVER } from "utils/env";
 import useMounted from "hooks/useMounted";
 import useLibraryContext from "components/context/LibraryContext";
+import { authTypeToPath } from "utils/auth";
+import { OPDS1 } from "interfaces";
```
hokei, line 9: Open Ebooks is still on OPDS 1?

kristojorg: Yep, CM is still OPDS 1 unless I'm missing someting

samanthaandrews (approved):
Looks like there are some failing unit tests but otherwise code looks good 👍

PR #332, opened 2021-12-30
502-errors

kristojorg:
This PR makes the 502 erros a bit better. It will be replaced by Jiayong's more comprehensive work on errors in the future.


kristojorg:
Nevermind, closing this as refreshing the page doesn't work. The only thing that works is to make the user wait 30 seconds or so. Bleh.

PR #333, opened 2022-01-03
OE-313: Add mock server to control Cypress data

samanthaandrews:
Cypress tests have been failing due to "502 third party authentication errors" which are coming from First Book. This PR adds mock service worker, `msw` to run during development if the `NEXT_PUBLIC_API_MOCKING` environment variable is set to `true`, which allows us to bypass these errors and ensure all of the test suites run and test what they intend to.

To turn on `msw` in development, you can set `NEXT_PUBLIC_API_MOCKING=true` in your `.env.local` file, which is not added to git.

Tests inside `cypress/tests/open-ebooks/integration` are being run on every PR.

Tests inside `cypress/tests/open-ebooks/server-e2e` are currently only being run on a daily cron job. However, you can still run these tests locally, and they should pass as long as you aren't also running the mock service worker. As stated above, you can turn off the mock service worker by setting setting `NEXT_PUBLIC_API_MOCKING=false`. Some of these tests still use Cypress `intercept()` and `wait()` commands to wait for responses to finish before moving to the next step and reduce flakiness when being run on daily cron jobs, but none of these responses are returning any stubbed data when msw is disabled.

To do:
- [x] Add mock service worker `msw` to control the data returned from the backend within Cypress tests
- [x] Ensure requests from Cypress tests are intercepted and return a 200 status code and appropriate data
- [x] Move all mocked data out of `cypress/fixtures` folder and into `msw/fixtures` folder
- [x] Ensure all tests in `cypress/integration` folder run with msw enabled
- [x] Ensure all tests in `cypress/server-end-2-end` run with msw disabled -- add a boolean check to determine when to use cypress intercept/wait pattern and when to skip it
- [x] Fix GitHub workflow file to build and run the project locally when running integration tests on PRs
- [x] Ensure CI can install the private `axis-now-decryptor` package
- [x] Update README to include latest info about Cypress tests and their interaction with MSW, how to turn off MSW in development, etc. <-- I will do this once I get some 👀  on this PR to determine that this is indeed the path we want to take


samanthaandrews:
I followed the recommended way to stub network calls in Next.js in the getServerSideProps method from this [YouTube video](https://www.youtube.com/watch?v=33Hq41O0bvU) and [blog post](https://glebbahmutov.com/blog/mock-network-from-server/), which uses a library called `nuxt`. I can confirm that the Next server is being run locally within Cypress, however I'm struggling to actually see the intercepted network calls return the mocked data.

I _think_ that once we get this working, we will be able to intercept the redirects and always return the appropriate data after we hit the `patrons/me` endpoint, but honestly I'm unsure. Cypress doesn't seem to know that we are ever hitting that endpoint, as it's never shown in the Cypress interface when the tests run. It only knows that we are sometimes being redirected to a different page, and that's what I'm hoping to control by adding `nuxt` to intercept these calls.

Maybe we can pair on this @kristojorg because right now I'm unsure if this is the right approach and if so, if I'm somehow missing something about the way `nuxt` is intercepting data.

samanthaandrews:
For some reason, after successfully mocking routes all day, I am now unable to build the app with MSW, getting the following error:

`ApplicationError: App Startup Failure: App Startup Failure: FetchError: request to https://qa-circulation.openebooks.us/USOEI/authentication_document failed, reason: Unexpected MODIFIER at 87, expected END`

Going to log off for the day and come back to this tomorrow.

cypress/support/commands.ts
```diff
@@ -71,24 +69,23 @@ Cypress.Commands.add(
...
-    cy.intercept(`${SERVER_URL}/loans`, req => {
-      req.reply({
-        statusCode: 200,
-        fixture: "open-ebooks/all-access/loans.xml"
-      });
-    }).as("loans");
-});
+// tests are actually hitting the server and not being mocked if using
+// qa url or NEXT_PUBLIC_API_MOCKING is set to false in development
+export const isEnd2EndTest =
```
samanthaandrews, line 74: I added this check to only add Cypress intercepts and waits if `msw` is off or if we are running against the qa-web url. This makes it easier to run these end-to-end tests also when `msw` is enabled if we ever decide we want that functionality. However, it is maybe annoying to see it sprinkled around these tests. Lmk if you think it's ok or if I should remove it.

cypress/support/commands.ts
```diff
@@ -71,24 +69,23 @@ Cypress.Commands.add(
...
+  Cypress.config().baseUrl === "https://qa-web.openebooks.us" ||
+  (Cypress.env("NEXT_PUBLIC_API_MOCKING") === false &&
+    Cypress.config().baseUrl === "http://localhost:3000");
```
samanthaandrews, line 77: I added this check to only add Cypress intercepts and waits if msw is off or if we are running against the qa-web url. This makes it easier to run these end-to-end tests also when msw is enabled if we ever decide we want that functionality. However, it is maybe annoying to see it sprinkled around these tests. Lmk if you think it's ok or if I should remove it.

cypress/tests/open-ebooks/integration/web-reader/read-books.ts
```diff
@@ -32,14 +29,14 @@ describe("Users can read/navigate loaned books", () => {
       .should("have.attr", "alt", "image");
   });
 
-  it("Can turn One Amazing Thing pages in paginated mode & scroll in scrolling mode", () => {
-    cy.openLoanedEPub(ONE_AMAZING_THING_LOAN);
+  it("Can turn Almost Midnight pages in paginated mode & scroll in scrolling mode", () => {
```
samanthaandrews, line 32: I changed this test to also use the Almost Midnight book just because I didn't want to have to add more mocked data at this time. But for future tests I think adding different books to mock is worthwhile.

cypress/tests/open-ebooks/server-e2e/borrow-read-return/borrow-read-return.ts
```diff
@@ -1,101 +1,121 @@
...
+  ALMOST_MIDNIGHT_LOAN,
+  ALMOST_MIDNIGHT_REVOKE_URL,
   V2_IFRAME_SELECTOR
 } from "../../../../support/utils";
 
 describe("Borrow a book, open it, and return it", () => {
   beforeEach(() => {
     // ensure book isn't already checked out before test begins
-    cy.checkoutAllAccessBook(D_IS_FOR_DRAMA_REVOKE_URL);
+    cy.checkoutAllAccessBook(ALMOST_MIDNIGHT_REVOKE_URL);
```
samanthaandrews, line 15: I changed this test to also use the Almost Midnight book to also reduce the number of mocks I was adding for books that were opened in the web reader. Once again, I think we can add additional books as we move forward.

cypress/tests/open-ebooks/server-e2e/borrow-read-return/borrow-read-return.ts
```diff
@@ -1,101 +1,121 @@
...
-      .its("response.statusCode")
-      .should("eq", 200);
-    cy.findAllByRole("heading", {
-      name: "Open eBooks (QA Server)",
-      timeout: 10000
-    }).should("exist", 2);
-
-    cy.log("navigate to d is for drama book page");
-    cy.visit(ALL_ACCESS_DETAIL_BOOK_PATH_D_IS_FOR_DRAMA);
+    // FIXME: navigating back to root with the back button in the web reader
```
samanthaandrews, line 86: I noticed this issue where clicking the "Back to open ebooks" button sometimes doesn't navigate you back to the root collection, but it seems to only be happening in development. I think we should create a ticket to look into this.

msw/handlers/index.ts
```diff
@@ -1,19 +1,175 @@
...
-  //     })
-  //   );
-  // })
+  // if we mock this with a 200 status response, then we get stuck
+  // in an infinite redirect loop on login, so we opt to actually
+  // communicate with the server in this case to determine if
+  // a user is actually authenticated
+  rest.head(SERVER_URL, () => {
+    return;
+  }),
```
samanthaandrews, line 35: When I mocked this with a 200 response, it was causing an infinite redirect loop when logging in with `msw` enabled. Not mocking this one request solves the issue, and thus far I haven't seen any 502 errors in development.

public/mockServiceWorker.js
```diff
@@ -0,0 +1,338 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker (0.36.3).
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
```
samanthaandrews, line 8: It seems we are not supposed to serve this file on production. Should I remove it from git and instead add a step in the README and in the CI with a command that instantiates it?

samanthaandrews:
Currently two suites are failing due to Cypress not being able to install the decryptor: `> Decryption Error: Attempting to instantiate a WebpubDecryptor without access to @nypl-simplified-packages/axisnow-access-control-web.`

cypress/tests/open-ebooks/server-e2e/borrow-read-return/borrow-read-return.ts
kristojorg: Nice, yea let's make a ticket for that

public/mockServiceWorker.js
kristojorg: Hmmm good question. And it needs to be in `/public`? I thought there was a way without putting it there

kristojorg:
> Currently two suites are failing due to Cypress not being able to install the decryptor

Ah yes, this was part of the reason I initially liked having the tests run against vercel. But we can make it work. We need to put a `.npmrc` file in github secrets and use that when installing in CI

samanthaandrews:
I have tried so many configs and none of them seem to installing the decryptor! You'll see in the `.github/workflows/oe-integration-cypress.yml` file I've tried using the `actions/setup-node@v2` way and manually setting the .npmrc file with this command `echo "@nypl-simplified-packages:registry=https://npm.pkg.github.com/nypl-simplified-packages\n//registry.npmjs.org/:_authToken=${{ secrets.AXIS_NOW_ACCESS_TOKEN }} > ~/.npmrc`

Maybe we can look at this together tomorrow morning after our Open Ebooks meetings? @kristojorg

kristojorg:
Yep let's do that @samanthaandrews !

cypress/tests/open-ebooks/server-e2e/borrow-read-return/borrow-read-return.ts
samanthaandrews: https://jira.nypl.org/browse/OE-320

package.json
```diff
@@ -17,6 +17,8 @@
     "build-and-start": "npm run build && npm run start",
     "test": "jest --watch",
     "test:ci": "jest --ci",
+    "build-for-cypress-integration-tests": "APP_ENV=test npm run build",
```
samanthaandrews, line 20: I tried setting NODE_ENV=test here (which is a value allowed by Next), but it wasn't always getting to set to `production` when running npm run build. This is because Next enforces this value to be `production` when running the build command. Here is the [thread](https://github.com/vercel/next.js/issues/17032) about it, which recommends using APP_ENV instead.

We could instead change this to be a boolean labeled by IS_CYPRESS_TEST_ENV or something similar. Lmk if you think that's better.

public/mockServiceWorker.js
samanthaandrews: It does need to be in `/public`. If it's not there, the app can't find it and won't build.

.github/workflows/oe-integration-cypress.yml
```diff
@@ -50,34 +23,37 @@ jobs:
...
-        with:
-          name: node_modules.zip
-
-      - name: Unzip node_modules
-        run: unzip ./node_modules.zip
+      - run: rm .npmrc
 
-      - name: Display current folder structure
-        run: ls -R
+      - name: Authenticate with the Github Package Registry
```
samanthaandrews, line 29: I ended up putting everything into one command instead of running the build and install steps separately. This is because whenever I started the app in the separate step, the decryptor was not available, despite it being available in the previous install step. I really don't know why this is the case, but I think this is also cleaner and the tests only take about 4 to 5 minutes to run, which seems reasonable. Also, this config is now a lot simpler.

kristojorg: Do you mean it's not parallel now?

cypress/support/commands.ts
kristojorg: You could maybe add a command or overwrite the intercept and wait commands to be enabled only when running not as e2e? That way we don't need to if/else in the test body itself.

I'd probably add a new command so it's clear to a new dev what it's doing (not always intercepting)

.github/workflows/oe-integration-cypress.yml
samanthaandrews: nvm i figured this out. build/install step is now still separate and occurs before the tests run.

cypress/support/commands.ts
samanthaandrews: I have a fix for this I'm just waiting to test it once the fixed auth_document issue is deployed to qa.

cypress/tests/open-ebooks/integration/login/login-navigation.ts
```diff
@@ -55,12 +60,10 @@ describe("Open eBooks login navigation", () => {
     cy.findByRole("link", { name: "Login with Clever" }).should("exist");
   });
 
-  it("redirects logged in user to nextUrl present", () => {
+  it.only("redirects logged in user to nextUrl present", () => {
```
kristojorg, line 63: Stray .only

cypress/tests/open-ebooks/integration/web-reader/read-books.ts
kristojorg: Yea makes sense. We can do that explicitly in tests for multiple books

cypress/tests/open-ebooks/server-e2e/borrow-read-return/borrow-read-return.ts
kristojorg: 👍

cypress/tests/open-ebooks/server-e2e/login/firstbook-login.ts
```diff
@@ -1,17 +1,18 @@
...
-  ALL_ACCESS_ROOT_COLLECTION_PATH
+  ALL_ACCESS_ROOT_COLLECTION_PATH,
+  SERVER_URL
 } from "../../../../support/utils";
 
 describe("FirstBook login", () => {
   it("Can login with FirstBook", () => {
-    cy.interceptLoans();
-    cy.intercept("GET", `${SERVER_URL}/groups/**`).as("groups");
+    cy.intercept("GET", `${SERVER_URL}/loans`).as("loans");
```
kristojorg, line 10: Do we want to be intercepting any requests in e2e tests?

kristojorg: Oh nvm it's so confusing in the api language that this is just aliasing it.

msw/fixtures/open-ebooks/fulfillmentInfoAlmostMidnight.json
```diff
@@ -0,0 +1,4 @@
+{
+  "book_vault_uuid": "C2612086-48AD-4354-B849-3634E68215A1",
+  "isbn": "9781466892859"
+}
```
kristojorg, line 4: These should be fake info, although that admittedly makes it impossible to read the actual book. We would need to provide the book as a fixture also. Annoying. But we can't have this data in the repo unfortunately.

package.json
kristojorg: I think a boolean is more clear, but also should be `NEXT_PUBLIC_` otherwise it is only available server side. I would also rename this to `build:cypress:ci` or something in line with our other names

src/middleware/check-auth.ts
```diff
@@ -59,24 +59,24 @@ export async function checkAuth(req: NextRequest, _ev: NextFetchEvent) {
     if (!response.ok) {
       if (response.status === 401) {
         console.log(
-          "Patron no authorized for book, redirecting to login.",
+          "Patron not authorized for book, redirecting to login.",
           req.nextUrl.pathname
         );
       } else if (response.status === 502) {
         console.log(
-          "502 Error while checking patrhon auth. Redirecting to login."
```
kristojorg, line 67: Lol thanks for fixing these

kristojorg (changes requested):
Just a few things

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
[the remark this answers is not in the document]
samanthaandrews: Ah yeah, let's just keep it there

cypress/tests/open-ebooks/integration/browse/browse.ts
```diff
@@ -1,20 +1,19 @@
...
   beforeEach(() => {
     cy.checkoutAllAccessBook(ALMOST_MIDNIGHT_FETCH_URL);
     cy.loginByApi("ALL_ACCESS_USER");
-    cy.loadApp();
   });
 
-  it.only("Can browse collections", () => {
-    cy.visit(APP_PATH);
+  it("Can browse collections", () => {
+    cy.visit(ALL_ACCESS_ROOT_COLLECTION_PATH);
```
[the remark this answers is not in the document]
samanthaandrews: Yeah I think you're right. I'll change it back!

README.md
```diff
@@ -253,6 +254,17 @@ OPENEBOOKS_CLEVER_HIGH_SCHOOL_ACCESS_TOKEN=
...
+
+MSW is also enabled by default in development mode since NEXT_PUBLIC_API_MOCKING is set to true in the `.env` file. You will know it is enabled because you will see warnings in the console letting you know that your responses are being stubbed by MSW. We need this value to be set to true in GitHub, because Cypress integration tests depend on it in CI.
+
+To turn off MSW in development, you can set NEXT_PUBLIC_API_MOCKING=false in your `.env.local` file, which is not added to git. However, if you're running any of the tests in this folder `cypress/tests/open-ebooks/integration` locally during development, you'll want to set the flag back to true in your `.env.local` file in order for them to behave as they will in CI.
```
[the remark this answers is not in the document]
samanthaandrews: I'm not sure it should _always_ be enabled in Cypress, though. It should always be enabled for test files in the `integration` directory, but never in the `server-e2e` directory. If I want to debug something in the server-e2e folder locally, then I should be able to easily turn msw off. I could make an event that runs on a before hook in each file in the `integration` directory? We'd just have to remember to use this command in all future test suites in that directory.

msw/fixtures/open-ebooks/fulfillmentInfoAlmostMidnight.json
samanthaandrews: I could add these values as secrets in GH Actions. Would that be an appropriate fix? I tried not mocking this call at all and the tests fail if we don't include this fixture.

kristojorg (approved):
Let's do this thing yeehaw

PR #336, opened 2022-01-10
OE-303: Update Clever and First Book login buttons

samanthaandrews:
These buttons have previously been served from the circulation manager, but after speaking with Nick we decided it makes more sense to package them within CPW. This PR updates the buttons with the latest designs from Jennifer.

My plan is to immediately make another PR to prod to release this, since Miguel wants to update First Book partner with these updates this week. Hence the changelog includes an updated release tag with just this one fix.

Screen Shot 2022-01-10 at 12 25 03 PM


Toxiapo (approved):
Nice!

PR #337, opened 2022-01-10
Update prod version 4.7.3

samanthaandrews:
Includes the update to use the latest First Book and Clever login buttons.

_Note: I think that the reason the older commits are displaying on this PR is because when I merged to prod last week, I selected the "squash and merge" option, which means these commits aren't registered as happening before on this branch. Should I not use squash and merge on prod?_


kristojorg:
> Note: I think that the reason the older commits are displaying on this PR is because when I merged to prod last week, I selected the "squash and merge" option, which means these commits aren't registered as happening before on this branch. Should I not use squash and merge on prod?

I think we should codify this on the readme: I _always_ use squash and merge when merging from a feature branch into a shared branch like `qa`. From then on, I _don't_ use squash and merge. The idea is that each feature branch/ticket gets its own commit, and only one clean commit. Then when we merge `qa` to `production` we just create a merge commit, preserving all of those feature commits. That way we have a commit history showing every feature/PR, not just every version

kristojorg (approved):

samanthaandrews:
> I think we should codify this on the readme: I _always_ use squash and merge when merging from a feature branch into a shared branch like `qa`. From then on, I _don't_ use squash and merge. The idea is that each feature branch/ticket gets its own commit, and only one clean commit. Then when we merge `qa` to `production` we just create a merge commit, preserving all of those feature commits. That way we have a commit history showing every feature/PR, not just every version

I will update the readme with this info. Thanks @kristojorg I think this is a good pattern.

PR #338, opened 2022-01-12
OE-301: Fix one more FirstBook ➡️ First Book naming instance

samanthaandrews:
Miguel noticed that I missed one of these changes. This instance of First Book doesn't display on our current configuration of the open ebooks app in dev or qa, but it does in prod.

https://jira.nypl.org/browse/OE-301

Here's what Miguel found and I changed to First Book instead:
image


src/auth/BasicAuthHandler.tsx
```diff
@@ -30,8 +30,14 @@ const BasicAuthHandler: React.FC<{ method: ClientBasicMethod }> = ({
...
-  const usernameInputName = method.labels.login;
-  const passwordInputName = method.labels.password;
+  const usernameInputName = method.labels.login.replace(
+    "FirstBook",
+    "First Book"
+  );
+  const passwordInputName = method.labels.password.replace(
+    "FirstBook",
+    "First Book"
+  );
```
Toxiapo, line 40: I am wondering if this is a good place to update this text, I believe the auth data comes from the backend, should we make an update there instead?

kristojorg: Yea we should update the backend here, not the frontend. Good catch.

kristojorg (changes requested):
We need to update this in the circulation manager

src/auth/BasicAuthHandler.tsx
samanthaandrews: Ok! I wasn't able to find this string anywhere in CM. I think it is coming from the authentication document that is configured for the prod deployment. Anyone know where I would find that code to edit or should I ask to make the change?

Toxiapo: Yeah, I can't find it either, It might be coming from the database, I think we need @leonardr 's help on this.

leonardr: Those strings are in fact served through the authentication document and can be customized through the circulation manager's admin interface (specifically, [here](https://circulation.openebooks.us/admin/web/config/patronAuth/edit/19), under "optional fields"). I've changed the strings and added a comment to [OE-263](https://jira.nypl.org/browse/OE-263) so that the underlying code will be cleaned up when we migrate to an upcoming version of the First Book patron authentication API.

samanthaandrews: Thanks, @leonardr . Looks good to me! Closing this PR.

Toxiapo: Thanks, @leonardr. That's good to know!

PR #339, opened 2022-01-19
handle network-related errors

Toxiapo:
This PR implements the strategy to handle all network-related errors that could happen when making HTTP requests to our backend server.

- [x] Create mock tests for each error state 
- [x] Make sure errors are caught by the ErrorBoundary, with the appropriate UI rendered
- [x] Inline errors that do not render a 404 page (Borrowing / Returning books, login page)
- [x] Configure global and inline SW config 
- [x] Update Error document with written descriptions on what was changed
- [x] Update unite tests(Fix current tests and add new withAppProps/getStaticProps, getServerSideProps, and middleware test


cypress/tests/open-ebooks/integration/errors/book.ts
```diff
@@ -0,0 +1,126 @@
...
+  ALL_ACCESS_DETAIL_BOOK_PATH_ALMOST_MIDNIGHT,
+  SERVER_URL
+} from "../../../../support/utils";
+
+let worker: SetupWorkerApi;
+
+before(() => {
+  // Let's manually start the worker
+  worker = setupWorker(...handlers);
+  worker.start();
```
Toxiapo, line 14: @kristojorg Unfortunately, I don't think this is going to work. I have to start the MSW server during build time and not during the run time. (That's what we have currently) because of the race condition when starting the msw server. (@samanthaandrews Have you experienced a similar issue when setting up msw?)

The reason I wanted to pull the worker out of the build time was that I need to mock different responses for the same endpoint(return 401, 402 from '/works'), in order to do that, I have to override the response on each test suite, but since I can't get access to the msw worker, I can not modify them.

Another thing I can try is maybe using the cy.intercept function and just mock the return response that way. However, this requires a request to the server initially, I will have to wait till we fix the CORS issue.

kristojorg: Hmmm I believe there should be a way to access the worker in tests and control how it responds

src/components/context/ContextProvider.tsx
```diff
@@ -18,16 +20,21 @@ type ProviderProps = {
...
-          <LibraryProvider library={library}>
-            <UserProvider>
-              <CatchFetchErrors>{children}</CatchFetchErrors>
-            </UserProvider>
-          </LibraryProvider>
+          <CredentialsProvider credentials={creds}>
+            <LibraryProvider library={library}>
+              <UserProvider>
+                {/* TODO: Why are we wrapping another config here? I think this is stopping me from doing a correct 401 redirect b/c CatchFetchErrors has it's own way of handling it. */}
+                {/* <CatchFetchErrors>{children}</CatchFetchErrors> */}
```
Toxiapo, line 33: I don't think we need this anymore? The parent SWRConfig wrapper should handle that.

src/pages/[library]/login/method/basic.tsx
```diff
@@ -45,7 +44,7 @@ export const BasicAuthHandler: React.FC = () => {
     });
   }
 
-  const { signIn, error, isLoading } = useUser();
```
Comment (Toxiapo) [inline src/pages/[library]/login/method/basic.tsx:48]: removing the "error" here should be fine too because we throw these errors in the global swr config.

cypress/support/utils.ts
```diff
@@ -109,3 +109,30 @@ export const ALMOST_MIDNIGHT_FETCH_URL = `${SERVER_URL}/works/Axis%20360%20ID/00
...
 
 export const V2_IFRAME_SELECTOR = "#iframe-wrapper iframe";
+
+// BOOKS USED FOR ERRORS, we use these to test the error handling for MSW intercepts eventhough they are totally valid books
+export const JFK_FETCH_URL = `${SERVER_URL}/works/Axis%20360%20ID/0016692239/borrow/15`;
+
+export const JFK_BOOK_ENDPOINT = `${SERVER_URL}/works/Axis%20360%20ID/0016692239`;
+export const GN_BOOK_ENDPOINT = `${SERVER_URL}/works/Axis%20360%20ID/0020037525`;
+export const D_BOOK_ENDPOINT = `${SERVER_URL}/works/Axis%20360%20ID/0016686310`;
+export const A_BOY_BOOK_ENDPOINT = `${SERVER_URL}/works/Axis%20360%20ID/0018923188`;
```
Toxiapo, line 119: I am reserving these books as the "problematic" books, and only use them for errors, I don't think they are used anywhere.

src/components/Error.tsx
```diff
@@ -3,16 +3,23 @@ import { H1 } from "./Text";
...
-  const { title = "Something went wrong", status, detail } = info ?? {};
+const ErrorComponent: React.FC<{
+  info?: OPDS1.ProblemDocument;
+  error?: any;
+}> = ({ error, info }) => {
+  const {
+    title = "Something went wrong",
+    status,
+    detail
+  } = error?.info ?? info ?? {}; // Q: Sometimes I get the error info from error.info but sometimes i get it from info.
```
Toxiapo, line 18: One from the error boundary and the other from our custom re-throw error?

src/components/bookDetails/index.tsx
```diff
@@ -52,21 +51,18 @@ export const BookDetails: React.FC<BookDetailsProps> = ({
       fetchBook(bookUrl, catalogUrl, { method: "GET", token }),
     { fallbackData: bookSeed }
   );
+
+  if (error instanceof ServerError && error.info.status !== 401) {
+    throw error;
+  }
```
Toxiapo, line 57: I have to extract the error from useswr and re-throw it here for the ErrorBoundary to catch them unfortunately : \  same for other components with similar setup

src/utils/swrConfig.ts
```diff
@@ -1,34 +1,42 @@
...
-      if (err.info.status === 401 || err.info.status === 404) {
-        severity = "info";
+      if (err.info.status === 401) {
+        // severity = "info";
+        console.warn("Caught 401 error in swr config");
+        const redirectUrl = `/${library.slug}/login?${LOGIN_ERROR_QUERY_PARAM}=${err.info.title}`;
+        clearCredentials(redirectUrl);
+      } else {
+        console.warn("Caught all error in swr config");
+        // we can catch this and log errors but we can't throw to the ErrorBoundary
```
Toxiapo, line 25: This is fine, we can catch errors here, and do error logging or redirect, but we can't re-throw the error, the error won't get to the error boundary, hence the setup above, to handle them inline.

Toxiapo:
@kristojorg - I don’t see a way to throw and catch errors within the ‘onError’ method. I can handle the redirect error, but throwing it within the onError function doesn’t trigger the error boundary (I can log it, but the page stays the same). Instead, I just extracted the “error” from the useSwr response, and if there’s a ServerError, I re-throw that, which triggers the ErrorBoundary neatly.

Another thing I tried to do was to intercept multiple endpoints. Still, it seems like if you try to intercept an endpoint that’s already registered and redeclare it later, the second one doesn’t get triggers. Instead, I mocked only a few endpoints with different book URLs. This should give us the confidence to verify that our useswr errors are getting caught correctly by the ErrorBoundary. It sucks because now I have to go into each component that uses the useSWR hook and manually add those re-throws. But this setup allows us to persist the MSW config that Sam had set up in the Github action to run the mocker simultaneously.

__mocks__/next/server.ts
```diff
@@ -0,0 +1,6 @@
+// Temporary solution for the middleware export error
+// https://github.com/vercel/next.js/issues/32848#issuecomment-1028325424
+import { NextRequest } from "next/dist/server/web/spec-extension/request";
+import { NextResponse } from "next/dist/server/web/spec-extension/response";
+
+export { NextRequest, NextResponse };
```
Toxiapo, line 6: I was getting the "SyntaxError: Unexpected token 'export" error when running unit tests for the middleware tests, so I had to add this mock file in our app suggested by the commented link.

src/__tests__/middleware.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      createMockReq({ collectionUrl: "/" }),
+      {} as NextFetchEvent
+    );
+
+    const locationKey = getSymbolKey(promise.headers, "map");
+    const responseInternalsKey = getSymbolKey(promise, "Response internals");
+
+    //   A redirect shows a 302 status
+    expect(promise[responseInternalsKey].status).toBe(302);
+    expect(promise.headers[locationKey].Location[0]).toBe("/");
```
Toxiapo, line 85: I couldn't access the response object from NextResponse directly via promise['map'], 'map' is a Symbol object that returns a unique instance each time Symbol is called I think? So this is the hack I use for now...

src/middleware/utils.ts
```diff
@@ -10,16 +11,39 @@ export function stringIsValidUrl(str: string): boolean {
...
+  });
+}
+
+export function rewriteError(origin: string, error: any, statusCode = 500) {
+  return new NextResponse(null, {
+    headers: {
+      "x-middleware-rewrite": String(new URL("error", origin))
+    },
+    status: statusCode,
+    statusText: JSON.stringify(error)
```
Toxiapo, line 47: We parse the error detail in the src/pages/error.tsx component

src/pages/[library]/index.tsx
```diff
@@ -20,55 +21,78 @@ const LibraryHome: NextPage = () => {
...
+      }
+    };
+  } catch (e: any) {
+    if (e instanceof ServerError) {
+      track.error(e, { severity: "error" });
+      if (e.info.status === 404) {
+        return {
+          props: {
+            notFound: true
+          }
```
Comment (Toxiapo) [inline src/pages/[library]/index.tsx:80]: returning { notFound: true; } renders the 404 page here!

Comment (kristojorg) [reply src/pages/[library]/index.tsx:80]: Nice!

src/auth/__tests__/BasicAuthForm.test.tsx
```diff
@@ -36,11 +34,9 @@ test("sumbits", async () => {
...
       )
   );
-  const utils = render(
-    <UserProvider>
-      <BasicAuthHandler />
-    </UserProvider>
-  );
+
+  const wrapper = makeContextWrapper();
+  const utils = render(<BasicAuthHandler />, { wrapper });
```
kristojorg, line 39: Could you just describe why this is necessary now? We previously would wrap everything in context inside of the `/test-utils/index` file

src/components/Error.tsx
kristojorg: Hmm I'm not sure why that would be. What do you mean by custom re-throw error? Maybe it could be that the error boundary passes it as `error` while the `_error` page passes just the `info`? Is that right?

src/components/ErrorBoundary.tsx
```diff
@@ -1,24 +1,16 @@
...
-      info={{
-        title: error.name,
-        detail:
-          error.message ??
-          "Something went wrong. Please refresh and try again. If the issue persists, contact your library support staff."
-      }}
-    />
-  );
+export const DefaultFallback: React.FC<FallbackProps> = ({ error, info }) => {
+  return <ErrorComponent error={error} info={info} />;
```
kristojorg, line 13: I think we should only need to pass _either_ `error` or `info` here.

kristojorg: You'll have to decide which one based on if catching an error thrown by `getStaticProps` in the `_error` page can pass a full `Error` class to the page and then down to `ErrorComponent`. I suspect it can only pass JSON, in which case we'll only want to pass some info like title, description, status code maybe since that's really all the `ErrorComponent` needs to render properly.

In fact, it might be simplest to just decide that the `ErrorComponent` receives a `title` and `description` and that's it. Then you just pass that through wherever you need it.

src/components/WebReader.tsx
```diff
@@ -91,7 +96,12 @@ const ReaderComponent: React.FC = () => {
       : null,
     fetchWithBookValutId
   );
-  if (manifestError) throw manifestError;
+  if (
+    manifestError instanceof ServerError &&
+    manifestError.info.status !== 401
```
kristojorg, line 101: Does this also handle 401 errors properly though by redirecting to login?

src/components/bookDetails/index.tsx
kristojorg: Yea that's frustrating. I guess [suspense](https://swr.vercel.app/docs/suspense) would make this easier by allowing errors to be caught by a boundary. But this seems okay for now

kristojorg: But just wondering, why are we only rethrowing if it is not a 401?

src/components/context/ContextProvider.tsx
kristojorg: Okay sure. I don't remember the exact reasoning for adding `CatchFetchErrors`, but I think it was meant to handle errors thrown by fetch itself (like CORS errors). So long as we still show a proper message on those errors, I'm fine to remove this.

src/components/context/UserContext.tsx
```diff
@@ -65,16 +63,24 @@ export const UserProvider: React.FC = ({ children }) => {
...
-          setError(err);
-          clearCredentials();
-        }
-      }
+      revalidateOnReconnect: false
     }
   );
+
+  if (error instanceof ServerError) {
+    // For a 404 error, we want to catch that in the errorboundary and not 404 page
```
kristojorg, line 71: I would slightly rewrite this comment: 404 errors on the loans endpoint should never happen. Therefore they're treated as ApplicationErrors.

I would also maybe switch this to be an `ApplicationError` since throwing a 500 `ServerError` will maybe be confusing in bugsnag... the server didn't actually give us a 500, we are just trying to say "something is broken here". What do you think?

src/components/context/__tests__/UserContext.test.tsx
```diff
@@ -27,8 +31,19 @@ mockSWR.mockReturnValue(defaultMock);
...
+function renderUserContext(creds?: AuthCredentials | undefined) {
+  return render(
+    <CredentialsProvider
+      credentials={{
+        clearCredentials: jest.fn(),
+        credentials: creds ?? undefined,
+        isLoading: false,
+        setCredentials: jest.fn()
+      }}
+    >
```
kristojorg, line 43: Oh we maybe need to add this to the global test context wrapper in `test-utils/index`

kristojorg: Along with the ability to pass credentials in, you can follow a similar pattern to what is in there already for things like library data

src/dataflow/withAppProps.ts
```diff
@@ -71,6 +71,20 @@ export default function withAppProps<PP extends PageProps = PageProps>(
...
+              notFound: true
+            }
+          };
+        } else {
+          throw e;
+        }
+      }
+
       // if it is not already an application error, wrap it in one
       const error =
```
kristojorg, line 89: Do we still need all this code below here? It seems like your code above handles it right?

src/middleware/check-auth.ts
```diff
@@ -81,7 +86,13 @@ export async function checkAuth(req: NextRequest, _ev: NextFetchEvent) {
           );
         }
       }
-      return redirectToLogin(library, requestedPathname);
+
+      // Redirect to login with an error message
+      const loginError = "Unkonwn Error Occurred. Please Relog.";
+      const loginUrl = new URL(`/${library}/login`, req.url);
+      loginUrl.searchParams.set(LOGIN_ERROR_QUERY_PARAM, loginError);
```
kristojorg, line 93: If we reuse this logic elsewhere it would be good to encapsulate it into something similar to `redirectToLogin`

src/middleware/check-auth.ts
```diff
@@ -92,25 +103,12 @@ export async function checkAuth(req: NextRequest, _ev: NextFetchEvent) {
...
-      });
-    }
-    /**
-     * @todo - send them to the error page
-     */
+    const errorDetail = {
+      status: 404,
+      title: "Error authenticating credentials",
+      detail: "An unexpected error occurred."
+    };
```
kristojorg, line 110: I think this should be a 500 error probably, and maybe try to use the `e.message` if it is an `Error` object

kristojorg: Also we should make a ticket to properly track errors from middleware

src/middleware/check-auth.ts
```diff
@@ -92,25 +103,12 @@ export async function checkAuth(req: NextRequest, _ev: NextFetchEvent) {
...
+    const origin = new URL(req.url).origin;
+    return rewriteError(origin, errorDetail, 404);
```
kristojorg, line 112: Do you need to pass 404 if the errorDetail already has a status defined?

src/middleware/utils.ts
```diff
@@ -10,16 +11,39 @@ export function stringIsValidUrl(str: string): boolean {
...
   return NextResponse.redirect(redirectUrl);
 }
 
-export function rewrite404() {
-  return NextResponse.rewrite(`/404`);
+// Custom rewrite with 404 status code
+export function rewrite404(origin: string) {
+  return new NextResponse(null, {
+    headers: {
+      "x-middleware-rewrite": String(new URL("404", origin))
```
kristojorg, line 35: I would probably make this accept a URL instead of origin, and then use `new URL("404", baseUrl)`. It will automatically just use the origin

src/pages/[library]/collection/_middleware.ts
```diff
@@ -45,11 +52,26 @@ export async function middleware(req: NextRequest, _ev: NextFetchEvent) {
...
-    );
-    return rewrite404();
+    if (response.status === 401) {
+      // Redirect to login with an error message
+      const loginUrl = new URL(`/${library}/login`, req.url);
+      const loginError = "Invalid Credentials";
+      loginUrl.searchParams.set(LOGIN_REDIRECT_QUERY_PARAM, requestedPathname);
+      loginUrl.searchParams.set(LOGIN_ERROR_QUERY_PARAM, loginError);
+
+      return removeAuthCookie(library, NextResponse.redirect(loginUrl));
```
Comment (kristojorg) [inline src/pages/[library]/collection/_middleware.ts:62]: Yea I think we should make a `redirectLogin` utility again that encapsulates this logic

src/pages/[library]/collection/_middleware.ts
```diff
@@ -45,11 +52,26 @@ export async function middleware(req: NextRequest, _ev: NextFetchEvent) {
...
+      const origin = new URL(req.url).origin;
+      return rewrite404(origin);
+    } else {
+      const errorDetail = {
+        status: response.status,
+        title: "Something went wrong while loading the collection page.",
+        detail: "An unexpected error occurred."
+      };
+      const origin = new URL(req.url).origin;
+      return rewriteError(origin, errorDetail, response.status);
```
Comment (kristojorg) [inline src/pages/[library]/collection/_middleware.ts:73]: We should make a new ticket to track middleware errors to bugsnag

src/pages/[library]/index.tsx
```diff
@@ -20,55 +21,78 @@ const LibraryHome: NextPage = () => {
...
+      }
     }
-  });
 
-  /**
-   * if the response doesn't work, make them login again.
-   * @todo - refine this a bit for 500 and other errors.
-   */
-  if (!response.ok) {
+    track.error(e, { severity: "error" });
```
Comment (kristojorg) [inline src/pages/[library]/index.tsx:87]: I think this track needs to be moved above the return or throw statements

src/pages/[library]/index.tsx
```diff
@@ -20,55 +21,78 @@ const LibraryHome: NextPage = () => {
...
+    // For other errors, we just take them to the login page.
     return {
       redirect: {
         destination: `/${librarySlug}/login`,
         permanent: false
       }
     };
   }
```
Comment (kristojorg) [inline src/pages/[library]/index.tsx:95]: We don't need any of this (or below) anymore right? You return or throw above so we should never hit this I think

src/pages/[library]/login/method/basic.tsx
Comment (kristojorg) [reply src/pages/[library]/login/method/basic.tsx:48]: I believe this is necessary to show the user login errors when trying to login via firstbook. For example if they put in the wrong credentials, they should stay on this page and see an "invalid credentials" message

src/pages/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const errorInfo = {
+    status: error?.status ?? 500,
+    title: error?.title ?? "Something went wrong rendering the page.",
+    detail: error?.detail ?? "An unexpected error occurred."
+  };
+
+  track.error(err);
+
+  return { errorInfo };
+};
```
kristojorg, line 34: Does this work? If this works, then why don't we just rewrite to the existing `_error.tsx`?

src/test-utils/index.tsx
```diff
@@ -64,11 +66,13 @@ const customRender = (ui: any, options?: CustomRenderOptions) => {
...
-                <BreadcrumbProvider>{children}</BreadcrumbProvider>
-              </UserContext.Provider>
-            </LibraryProvider>
+            <CredentialsProvider credentials={credentialsState}>
+              <LibraryProvider library={library}>
+                <UserContext.Provider value={user}>
+                  <BreadcrumbProvider>{children}</BreadcrumbProvider>
+                </UserContext.Provider>
+              </LibraryProvider>
+            </CredentialsProvider>
```
kristojorg, line 75: Oh shoot I see you did add the CredentialsProvider here. I'm wondering why it was necessary to use a custom wrapper in those tests then?

kristojorg (commented):
Apologies for the number of comments here! I know this is a tough ticket. Good work so far, just a little more to push it over the hump and get it merged

.github/workflows/oe-integration-cypress.yml
```diff
@@ -57,14 +57,13 @@ jobs:
     needs: [install]
     container:
       image: cypress/included:9.3.1
-      # options: --user 1001 # https://github.com/cypress-io/github-action#firefox
+      options: --user 1001 # https://github.com/cypress-io/github-action#firefox
```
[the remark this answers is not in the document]
Toxiapo: oh haha this was an attempt to fix the cypress build, still didn't work. : \ I will remove it.

msw/handlers/index.ts
```diff
@@ -124,34 +120,6 @@ export const handlers = [
...
-    }
-  ),
-
-  rest.get(
-    `${SERVER_URL}/works/Axis%20360%20ID/0016692239`,
-    (req, res, ctx) => {
-      return res(ctx.xml(bookJohnFKennedy), ctx.status(200));
-    }
-  ),
-
```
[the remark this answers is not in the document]
Toxiapo: I am currently reserving these books for the useSWR tests, this file takes precedence for network interceptions so I had to take them out.

I don't think we test these endpoints in our tests except for the ALMOST_MIDNIGHT one, is this correct @samanthaandrews ?

src/pages/[library]/login/method/basic.tsx
Comment (Toxiapo) [reply src/pages/[library]/login/method/basic.tsx:48]: yeah, that makes sense!

src/pages/error.tsx
Toxiapo: Yes, I can't figure out why but the rewrite will render this page with "x-middleware-rewrite: error", but "x-middleware-rewrite: _error" does not render the `_error.tsx` page.

src/middleware/utils.ts
Toxiapo: hmm yeah, you are right, let me see if I can store this piece of information somewhere else.

I don't see a way to add a query param in the URL that let us parse it in our error component,

src/pages/[library]/index.tsx
Comment (Toxiapo) [reply src/pages/[library]/index.tsx:95]: It would still hit this, we only return or throw ServerError.

cypress/tests/open-ebooks/integration/errors/useSwr.ts
```diff
@@ -0,0 +1,94 @@
...
+      );
+    })
+  );
+
+  // Allow ErrorBoundary to catch errors
+  cy.on("uncaught:exception", () => false);
+  cy.loginByApi("ALL_ACCESS_USER");
+});
+
+describe("Mock endpoint that catches 401, 404, and 502 errors in the ErrorBoundary.", () => {
```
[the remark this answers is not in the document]
Toxiapo: Yep, we are mocking the read book path for now because the loan or the collection page uses the same endpoint so we can't mock them, unfortunately. I will mark this in our document and come back to this with a better strategy.

src/middleware/utils.ts
Toxiapo: Yeah, attaching the error to *searchParams* works! I thought it didn't work because I couldn't get the error detail in my error page but seems like if I change the error page to `getServerSideProps`, I was able to get the query!

src/components/Error.tsx
Toxiapo: Sorry, I missed this comment.. yeah I think that's the only explanation for this.

src/__tests__/middleware.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+    });
+    fetchMock.once("Bad Response", {
+      status: 401,
+      url: "/",
+      ok: false
+    } as any);
+
+    await CollectionMiddleware(mockedReq, {} as NextFetchEvent);
+
+    expect(mockedRemoveAuthCookie).toBeCalledTimes(1);
```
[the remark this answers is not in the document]
Toxiapo: Ah! Okay, I remember this and it was something I would like to discuss with you but forgot. 

I was not able to set up the middleware unit test with Jest without getting the `SyntaxError: Unexpected token 'export'` error from `import { NextRequest, NextFetchEvent, NextResponse } from "next/server";`, The fix I found was to create a mock file and import those functions but then I was getting another error with "removeAuthCookie" function not found... so I mocked the `removeAuthCookie` function in the test...

src/components/WebReader.tsx
Toxiapo: Yep! test added for the read page

src/components/bookDetails/index.tsx
kristojorg: But wouldn't the global swr config handle the 401 so it shouldn't ever arrive here? And if it _does_ arrive here, we should probably throw it so it gets handled by the error boundary and we don't continue trying to render

src/components/ErrorBoundary.tsx
kristojorg: Yea so bugsnag reporting happens in the `<ErrorBoundary>`, which should get an `Error` object. The `<ErrorComponent>` however, doesn't do any tracking right? It just renders the UI. So maybe just pass it the strings it needs to render directly?

src/dataflow/withAppProps.ts
kristojorg: Oh good point

src/components/WebReader.tsx
Toxiapo: ahh okay, yea that makes sense.

src/__tests__/middleware.test.tsx
```diff
@@ -198,10 +184,9 @@ describe("Middleware: login", () => {
       ok: true
     } as any);
 
-    const promise: any = await LoginMiddleware(mockedReq);
+    const promise = await LoginMiddleware(mockedReq);
 
-    const responseInternalsKey = getSymbolKey(promise, "Response internals");
-    return expect(promise[responseInternalsKey].status).toBe(302);
+    return expect(promise.status).toBe(307);
```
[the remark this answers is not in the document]
Toxiapo: Yeah. This is something we have to address, we have a ticket open [here](https://jira.nypl.org/browse/OE-350).

I will put a comment here.

cypress/tests/open-ebooks/integration/browse/browse.ts
```diff
@@ -14,7 +13,7 @@ describe("All-access browsing", () => {
   });
 
   it("Can browse collections", () => {
-    cy.visit(APP_PATH);
+    cy.visit(HIGH_SCHOOL_ACTION_COLLECTION_PATH);
```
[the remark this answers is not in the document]
Toxiapo: You are right, this should be APP_PATH! I must've changed it by accident with the merge.

src/__tests__/getServerSideProps.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+const ctx = {
+  req: { headers: { ...request.headers, host: "test-domain.com" } },
+  query: { [LOGIN_REDIRECT_QUERY_PARAM]: "/testlib/test-path" },
+  res: {},
+  resolvedUrl: "/test-path"
+} as unknown as GetServerSidePropsContext;
+
+test("Throw ServerError: 404 error returns props notFound:true", async () => {
+  fetchMock.mockRejectedValueOnce(
+    new ServerError("/", 404, {
```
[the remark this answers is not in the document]
Toxiapo: yea I think this is fine. The example from their doc takes an error object. `new Error('Async error')`

src/components/BookList.tsx
```diff
@@ -73,6 +76,11 @@ export const InfiniteBookList: React.FC<{
     }
   );
 
+  if (error instanceof ServerError) {
+    track.error(error);
+    throw error;
+  }
```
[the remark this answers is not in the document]
Toxiapo: Yeah, we don't need to check this actually, we should throw this whenever there's an error.

src/pages/error.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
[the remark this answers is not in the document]
Toxiapo: yea, it wouldn't let me rewrite to the _error page so I had to create this specifically for the `rewrite` function.

src/__tests__/getServerSideProps.test.tsx
```diff
@@ -0,0 +1,89 @@
+import { GetServerSidePropsContext } from "next";
```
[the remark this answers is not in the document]
Toxiapo: Updated, also added a note to include the test for the /read page once we merge OE-356.

src/__tests__/getStaticProps.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+const ctx = {
+  req: { headers: { ...request.headers, host: "test-domain.com" } },
+  query: { [LOGIN_REDIRECT_QUERY_PARAM]: "/testlib/test-path" },
+  res: {},
+  resolvedUrl: "/test-path"
+} as unknown as GetStaticPathsContext;
+
+// This test is meant to test all the pages using the getStaticProps funciton,
+// I am mocking with the collection path only because all the page are essentially the same
+// withAppProps function
```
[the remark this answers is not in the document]
Toxiapo: Yea I missed the book detail page. Updated!

src/pages/error.tsx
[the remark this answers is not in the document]
kristojorg: Ok strange but sounds good

kristojorg (approved):
OK yeeeehaw this looks good let's get it merged!

PR #342, opened 2022-01-20
OE-326/mobile-webview

kristojorg:
This is a POC that removes header/footer for mobile user agents. We would probably want to refactor a bit to reduce duplication, but that shouldn't be difficult.


src/pages/[library]/_middleware.ts
```diff
@@ -0,0 +1,28 @@
+import { NextRequest, NextFetchEvent, NextResponse } from "next/server";
+
+const IOS_WEBVIEW_UA =
+  "Mozilla/5.0 (iPhone; CPU iPhone OS 15_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148";
```
Comment (ettore) [inline src/pages/[library]/_middleware.ts:4]: this will change based on the device iOS version. For instance the `15_1` will change for every minor release. Similarly for WebKit version and the string after `Mobile/`. I wonder if we should just add a header to every web-view requests...

Comment (kristojorg) [reply src/pages/[library]/_middleware.ts:4]: Yea we should add a header. Hopefully that header will persist between web-view navigations too. Let's try it.

PR #351, opened 2022-02-01
[WIP] OE-285

samanthaandrews:


Toxiapo (commented):
I am not sure If I am testing this correctly. I see the offline message after clicking on a book.

https://user-images.githubusercontent.com/14434233/153272136-7865e26e-69ba-46ef-ae83-8a0464b3f636.mov

samanthaandrews:
> I am not sure If I am testing this correctly. I see the offline message after clicking on a book.
> 
>  Screen.Recording.2022-02-09.at.2.04.51.PM.mov

Weird, this is what I am seeing. 

https://user-images.githubusercontent.com/34348332/153275356-6dc5e394-45cf-4544-bdeb-9992442c1b98.mov

I'm not sure if there is much we can do to prevent the offline message from showing at a certain point without using a PWA. I know we previously set this up but then disabled it. I think we can re-enable it now that we have webpack 5 up and running. I see this as a ticket in the current sprint https://jira.nypl.org/browse/OE-291, but I also see there is a bullet point in the issue for some UX input and I'm not sure if we've received that yet. @Toxiapo feel free to assign that ticket to me if you'd like me to take a look and follow up with UX!

Toxiapo:
> > I am not sure If I am testing this correctly. I see the offline message after clicking on a book.
> > Screen.Recording.2022-02-09.at.2.04.51.PM.mov
> 
> Weird, this is what I am seeing.
> 
>  Screen.Recording.2022-02-09.at.12.20.13.PM.mov 
> I'm not sure if there is much we can do to prevent the offline message from showing at a certain point without using a PWA. I know we previously set this up but then disabled it. I think we can re-enable it now that we have webpack 5 up and running. I see this as a ticket in the current sprint https://jira.nypl.org/browse/OE-291, but I also see there is a bullet point in the issue for some UX input and I'm not sure if we've received that yet. @Toxiapo feel free to assign that ticket to me if you'd like me to take a look and follow up with UX!

sure thing, thank you!

kristojorg (commented):
I have thought about this ticket and I think we should close it actually. Here are my reasons:

1. Offline browsing isn't a goal currently (as discussed in our recent meeting)
2. The only change I see is that we don't throw FetchErrors or other errors originating from fetch logic anymore. While it makes sense to show the page information we _do_ have, I think if we want to be offline capable, we would want to show some kind of UX here for the error we caught. But like I said, we don't have offline browsing as a goal, so I think better to not swallow errors for the time being.

PR #353, opened 2022-02-03
OE-342: Refactor fetching logic

kristojorg:
This is a little refactor that makes our fetch logic much more composable, which will make it easier to extract out of CPW into a shared module in the future.


samanthaandrews (approved):
LGTM

PR #359, opened 2022-02-10
OE-347 listview-loader

kristojorg:
This is a simple PR to provide fallback data to the infinite listview so we don't see an initial loader when viewing those pages. To verify, simply click "view more" on the books for integration testing. You shouldn't see a loader.


samanthaandrews (approved):
nice!

PR #361, opened 2022-02-10
OE-339/static-pages

kristojorg:
This PR:

- Adds static files for the old marketing pages to `/public/_dotnet/`.
- Adds "rewrites" in next.config.js to those files in `_dotnet`. So we rewrite `/faq` to `/_dotnet/faq`. 
- Adds rewrites to allow the JS, image and css requests from those files to properly land in `_dotnet`.
- Adds redirects from `/slug.html` to `/slug`, because some of the links on dotnet go to .html urls.
- Adds a new marketing header component written with **Chakra** instead of theme-ui. This is because we would like to ultimately migrate to Chakra from theme-ui. The marketing pages will be our first opportunity to do so, and this starts to set us up for that.
- Updates the `Footer` component with different links that are not dynamic based on the auth document but hardcoded.

Note that switching between the home page and the other marketing pages you will see a difference in header design. This is going to be the case until we get new designs for all the marketing pages.


src/marketing/Layout.tsx
```diff
@@ -0,0 +1,31 @@
+import { ErrorBoundary } from "components/ErrorBoundary";
```
kristojorg, line 1: I just extracted this out from the OpenEbooksLandingComponent so we can use it in future new marketing pages

src/components/Button/styles.ts
```diff
@@ -67,6 +67,7 @@ export const styleProps = (
           cursor: "default"
         }
       };
+    }
```
samanthaandrews, line 70: not really related to these changes but I'm thinking we should add `break;` statements at the end of each case

public/_dotnet/about.html
```diff
@@ -0,0 +1,372 @@
...
+            ages 4-14, comprised of 3,000 titles.
+          </li>
+        </ul>
+        <h4>
+          Are you a publisher who wants to make your content available to kids
+          in need?
+        </h4>
+        <ul class="actions">
+          <li>
+            <a href="contribute.html" class="button big">Learn More</a>
```
samanthaandrews, line 210: I see this reference to the `contribute.html` page but don't see this page in the ones added to this PR. The `app_acknowledgements.html` page is another one that is being referenced from our mobile apps (I have edited and deployed it in recent months from the openebooksdotnet repo, which is why I'm thinking about it), and I'm wondering if we need to add it to this PR as well.

I guess the larger question here is: how do we make sure we're moving everything over that we need?

public/_dotnet/app_privacy.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Open eBooks Privacy Policy</title>
+  </head>
+  <body>
+    <font face="arial">
```
Toxiapo, line 8: A ticket to remove HTML4 elements? haha

src/pages/index.tsx
```diff
@@ -1,14 +1,315 @@
 import React from "react";
-import { IS_OPEN_EBOOKS } from "utils/env";
-import OpenEbooksLandingPage, {
-  landingPageStaticProps
-} from "components/OpenEbooksLanding";
-import MultiLibraryHome from "components/MultiLibraryHome";
+import MarketingPageLayout, {
```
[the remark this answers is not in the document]
Toxiapo: I think we can remove the "MultiLibraryHome" component.

public/_dotnet/app_privacy.html
samanthaandrews: I too noticed this when fixing the `app_acknowledgements.html` page. Could be a good ticket to add to the backlog.

public/_dotnet/about.html
samanthaandrews: I did another pass to create a more complete list of pages & content we might be missing:
- `getstarted.html`
- `contribute.html`
- `app_acknowledgements.html`
- `getinvolved.html` (maybe the same as getinvolved.html?)
- `support.html`
- the `educators` page doesn't seem to be linking to the PDFs as we'd expect them to
- `app.html`, `app_acknowledgments.html`, `app_privacy.html`, and `app_user_agreement.html` are all likely being linked from the Open eBooks mobile app. I think it would be helpful for Miguel to determine if these all need to be moved. How is the app privacy different from the privacy policy we already have in the `privacy.html` page?

package.json
```diff
@@ -37,7 +37,7 @@
     "@bugsnag/plugin-react": "^7.14.0",
     "@csstools/normalize.css": "^12.0.0",
     "@emotion/react": "^11.7.1",
-    "@nypl/design-system-react-components": "^0.7.2",
+    "@nypl/design-system-react-components": "^0.25.9",
```
hokei, line 40: How are we using this?

hokei:
can you add a readme type of thing to the `/dotnet` folder just explaining what that folder is, like how future engineers should see this folder if we haven't managed to move everything out by the time we're gone

src/marketing/Header.tsx
```diff
@@ -0,0 +1,194 @@
...
+} from "@chakra-ui/react";
+import { Icon, IconNames } from "@nypl/design-system-react-components";
+import useUser from "components/context/UserContext";
+import Link from "next/link";
+import React from "react";
+
+const existingOEGray = "#383b43";
+const menuButtonColor = "gray.300";
+
+export default function Header() {
```
hokei, line 31: can you put some documentation in here differentiating this header from https://github.com/NYPL-Simplified/circulation-patron-web/blob/4d5393fa4ce6b98674043a07c682fabd59cb5209/src/components/Header.tsx ?

public/_dotnet/about.html
samanthaandrews: Ok confirmed with Miguel that we do need to move these pages, although he admits he did not think to include them in the original ticket. Here's the thread: https://nypl.slack.com/archives/C0153AFA1RS/p1644849000269259

src/components/Button/styles.ts
kristojorg: I think the `return` statements handle that. You should only need `break` if you don't return something, but hopefully these all do

public/_dotnet/app_privacy.html
kristojorg: What does this mean haha

src/components/Button/styles.ts
samanthaandrews: Ah you're right!

public/_dotnet/about.html
```diff
@@ -0,0 +1,372 @@
...
+    <script src="js/skel-layers.min.js"></script>
+    <script src="js/init.js"></script>
+    <noscript>
+      <link rel="stylesheet" href="css/skel.css" />
+      <link rel="stylesheet" href="css/style.css" />
+      <link rel="stylesheet" href="css/style-xlarge.css" />
+    </noscript>
+    <script>
+      (function (i, s, o, g, r, a, m) {
+        i["GoogleAnalyticsObject"] = r;
```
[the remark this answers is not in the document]
kristojorg: I'm not sure we would lose those analytics. The script is still on the page so I'm guessing it'll still fire. We should ask about whether we need them though

public/_dotnet/app_privacy.html
kristojorg: Yikes I had no idea haha

samanthaandrews (approved):

src/marketing/Header.tsx
hokei: Based on how my IDE is set up, I'm concerned I'd be importing the wrong `<Header>` all the time.  My preference would be to use `<MarketingHeader>`

hokei: But also I mostly meant putting in a jsdoc style thing explaining this component.

Toxiapo (approved):

PR #362, opened 2022-02-10
OE-320: Fix web reader back button not working in development

samanthaandrews:
When you clicked the "Return to Open eBooks (QA Server)" button in the web reader in development mode, the app wouldn't navigate back to the root collection as expected. It would just re-render the web reader page.

As far as I can tell this was only happening in development, but this meant we couldn't test the web reader's "back" button functionality in Cypress.

It seemed the issue was coming from not appropriately passing the `sx` styles prop down to the `<BaseLink />` component. I also updated the `borrow-read-return` Cypress test in the `server-e2e` folder and it is now passing.


Toxiapo (approved):
Nice, I wonder how you figure that out, that's super weird.

The button is working as expected and has more test coverage, so this is LGTM! 🎉

samanthaandrews:
> Nice, I wonder how you figure that out, that's super weird.

Debugging steps went something like this:
1. Try replacing the `<Link />` component with a regular `<a href="/" />` tag to determine if it was indeed something within the Link component that was causing the problem. The regular `a` tag seemed to work fine, so I knew there was something going wrong within that component.
2. Make sure the <Link /> component was rendering an `a` tag in the DOM using the element inspector dev tools. It was indeed rendering everything as expected.
3. Do some quick googling to figure out if other people were having trouble with `next/link` following hrefs, and I did find a thread where some people were talking about CSS not playing well with `next/link`. So then I tried removing the `sx` prop from the `<Link />` component and that also seemed to fix the issue. From there I tried wrapping the children of the <Link /> in a separate `div` with the `sx` prop, which also seemed to work. Ultimately I determined that we just weren't passing the sx through all the way to the `a` tag.

PR #364, opened 2022-02-14
OE-335: Update README with Vercel deploy to prod instructions

samanthaandrews:
- Clean up README deployment instructions to describe our release process with Vercel https://jira.nypl.org/browse/OE-335


README.md
```diff
@@ -267,7 +269,16 @@ You can still run these tests locally, and they should pass as long as you aren'
 
 When creating links using `<Link>`, you don't need to worry about whether it is for a single or multi-library route config. Write the `as` and `href` like you would if the package only supported one-library setups, and the `<Link>` will prepend `/[libraryId]` to your routes if needed.
 
-# Deploying
+# Deploying with Vercel (currently using this deployment strategy—the Docker instructions are outdated)
```
samanthaandrews, line 272: I added this which describes the deployment process with Vercel and the changes in this PR. I didn't remove the Docker instructions since I know we have a ticket for the next sprint to re-implement our Docker deployments. I figured in that ticket we should update the README again to reflect the new Docker process.

.github/workflows/release-prod.yml
```diff
@@ -16,26 +16,48 @@ jobs:
...
-
-      - name: Release
-        id: create_release
-        if: steps.autotagger.outputs.tagname != ''
-        uses: actions/create-release@v1.0.0
+          tag: ${{ steps.bump-semver.outputs.new_version }}
+          message: '${{ steps.bump-semver.outputs.new_version }}: PR #${{ github.event.pull_request.number }} ${{ github.event.pull_request.title }}'
+
+      # generate and commit changelog
+      - name: Run Changelog CI
```
samanthaandrews, line 46: We can do fancier stuff with the changelog CI action by adding a config file to the repo and using more specific labels on PRs. I don't think this is really necessary but nice to know it exists if we decide to use it: https://github.com/marketplace/actions/changelog-ci

.github/workflows/oe-e2e-cypress.yml
```diff
@@ -1,4 +1,4 @@
-name: Open Ebooks End-to-End Tests with Cypress
```
kristojorg, line 1: Maybe rename this file `chron-e2e-cypress`?

.github/workflows/prod-e2e-cypress.yml
```diff
@@ -0,0 +1,66 @@
...
+env:
+  GH_PACKAGE_AUTH_TOKEN: ${{ secrets.GH_PACKAGE_AUTH_TOKEN }}
+
+jobs:
+  cypress-tests:
+    name: Cypress Tests
+    runs-on: ubuntu-latest
+    ## only run this check on deployments that build with the openebooks-qa.vercel.app url (aka pull requests to the production branch)
+    ## because we cannot check for the branch reference on a deployment_status event trigger
+    if: github.event.deployment_status.state == 'success' && github.event.deployment_status.environment == 'Preview – open-ebooks-web' && github.event.deployment_status.target_url == 'https://openebooks-qa.vercel.app'
```
kristojorg, line 14: Since we are now using staging.openebooks.org, can we change to that url actually? And maybe change the daily chron jobs as well?

.github/workflows/release-prod.yml
```diff
@@ -16,26 +16,48 @@ jobs:
...
 
-        # the tag will match the package.json version (eg. v1.0.0)
-      - name: Tag
-        id: autotagger
-        uses: butlerlogic/action-autotag@stable
+      # output a semver update level major, minor, patch from a pull request release label
+      # labels are release/major, release/minor, and release/patch
+      - uses: actions-ecosystem/action-release-label@v1
+        id: release-label
+        if: ${{ github.event.pull_request.merged == true }}
```
kristojorg, line 23: Will the workflow fail if there is no label assigned?

.github/workflows/release-prod.yml
```diff
@@ -16,26 +16,48 @@ jobs:
...
+        if: ${{ steps.release-label.outputs.level == 'major' || steps.release-label.outputs.level == 'minor' || steps.release-label.outputs.level == 'patch' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ steps.autotagger.outputs.tagname }}
-          release_name: ${{ steps.autotagger.outputs.tagname }}
+          tag_name: ${{ steps.bump-semver.outputs.new_version }}
+          release_name: ${{ steps.bump-semver.outputs.new_version }}
           # use the body of the PR commit as the release body
           body: ${{ github.event.pull_request.body }}
```
kristojorg, line 62: Is it possible to get the changelog output into the body of the github release?

kristojorg (commented):
Looks good, just a couple small things

.github/workflows/prod-e2e-cypress.yml
```diff
@@ -0,0 +1,66 @@
...
+env:
+  GH_PACKAGE_AUTH_TOKEN: ${{ secrets.GH_PACKAGE_AUTH_TOKEN }}
+
+jobs:
+  cypress-tests:
+    name: Cypress Tests
+    runs-on: ubuntu-latest
+    ## only run this check on deployments that build with the openebooks-qa.vercel.app url (aka pull requests to the production branch)
+    ## because we cannot check for the branch reference on a deployment_status event trigger
+    if: github.event.deployment_status.state == 'success' && github.event.deployment_status.environment == 'Preview – open-ebooks-web' && github.event.deployment_status.target_url == 'https://openebooks-staging.vercel.app'
```
kristojorg, line 14: I don't think this is the right url. I think you want staging.openebooks.org

samanthaandrews:
I am second-guessing a lot of this... I did a test run and the bump-semver created a tag with version 3.1.4 or something that seems very off. Going to move this back to draft until I come back to it.

.github/workflows/prod-e2e-cypress.yml
samanthaandrews: The last time we made a PR to `production` branch, vercel built a target_url at `openebooks-qa.vercel.app`, so I just want to confirm that by changing it to staging.openebooks.org, this check will still pass.

I'm realizing that by running these tests on staging.openebooks.org we won't catch issues related to the environment variables being off in Vercel, which is what happened in one of our recent releases. Not sure if there is any way around that, though.

.github/workflows/release-prod.yml
samanthaandrews: I added a `release.yml` file to the `.github` directory, which is described in the docs as a way to generate automatic release notes based on the PRs that have been merged since the last release: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

samanthaandrews:
Ok I ended up changing this PR to only include the Cypress tests to run on staging.openebooks.org and update the README with our Vercel deploy instructions. since we're moving to a docker deploy process, i'm going to hold off on doing the automatic changelog and release tags until that piece is completed. i learned a lot about these github actions while working through this so now I have a better idea of what's possible, but i'm still noodling on the best way to release to prod without making two PRs. i don't want staging and production branches to become out of sync if we can avoid it. @kristojorg let's chat about exactly what we'd like to see so I can write up a ticket describing it.

.github/release.yml
```diff
@@ -0,0 +1,4 @@
+changelog:
+    - title: Changelog
+      labels:
+        - "*"
```
kristojorg, line 4: Do we need this still?

.github/workflows/prod-e2e-cypress.yml
```diff
@@ -0,0 +1,66 @@
+name: Production End-to-End Cypress Tests
```
kristojorg, line 1: I think this should be renamed, it's not production is it? It's testing staging on new deploys

.env
```diff
@@ -1,5 +1,5 @@
 # must set a config file
-CONFIG_FILE=/app_config/open-ebooks-config.yml
```
kristojorg, line 2: We currently have a `/config` folder added to the gitignore. Should we just be using that folder, or are we intending to commit these config files to the repo? I think we could commit a dev only config file (without bugnsag), but not the prod one

kristojorg (changes requested):
This seems good, but had a couple questions. I think we should change the PR name as well to remove the other ticket name just in case. In the future, we might switch the Lerna which has auto changelog updating built in, so I think it makes sense to wait on that aspect.

.github/release.yml
samanthaandrews: I think we should keep it to see if it generates release notes in the github release. The [docs](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes) say that it does, but I'm unsure if it will work with the `actions/create-release@v1.0.0` that we're using.

.env
samanthaandrews: We need to commit the open ebooks config in order to build the app and run the Cypress tests in CI. I just noticed we had two files for open ebooks, so I removed one. The remaining `opene.yml` doesn't include bugsnag. It looks like this:

```
instance_name: qa-web.openebooks.us
companion_app: openebooks
show_medium: false
libraries:
  app: "https://qa-circulation.openebooks.us/USOEI/authentication_document"
media_support:
  application/vnd.librarysimplified.axisnow+json: show
openebooks:
  default_library: app
```

.github/release.yml
samanthaandrews: If it doesn't do what's expected in our next release to Prod, I'll remove it.

.github/workflows/prod-e2e-cypress.yml
kristojorg: So I _think_ we want to test staging.openebooks.org because when making a PR to prod, this should represent the exact code that will go to production. The exceptions are:

1. This url will deploy against the QA circulation manager, not production
2. There _can_ be environment variable differences between these staging.openebooks.org and openebooks.org. However, this risk is very difficult to mitigate. 

What happened in the last problem we experienced was that the env vars for _both_ staging.openebooks.org and openebooks.org were incorrect in vercel. So this should, at least, protect us against that.

.env
kristojorg: Ok perfect, so we will have the non-production version committed. We should probably make a ticket to move all of this configuration into ENV vars in the `.env` file as part of removing the multi-library setup. Seems like a pretty easy first part of that ticket.

kristojorg (approved):
I guess you can make a PR to prod just to see if the `target_url` is indeed staging.openebooks.org. I can't imagine it's openebooks-qa.vercel.app, but maybe it's openebooks-staging.vercel.app now, since that might be the "base" domain vercel uses for the branch, which is then aliased to staging.openebooks.org, and vercel might just sent the base domain in the github event.

Otherwise, looks great to me!

.github/workflows/prod-e2e-cypress.yml
samanthaandrews: Ok I made a draft PR to production to test what the `github.event.deployment_status.target_url` variable is, and it looks like it is currently building at whatever I named the branch .vercel.app. So in the draft PR, it's `https://open-ebooks-web-git-prod-test-vercel-deploy-nypl.vercel.app/`.

Is there a way within Vercel to alias preview URLs to the `production` branch to `staging.openebooks.org`? Since this workflow runs on the `deployment_status` event, instead of on `push` or `pull_request` events, we don't have access to the git branch name in the workflow, and I need to check for something else in this `if` statement to ensure this workflow runs only on pull requests to the `production` branch.

kristojorg: Huh really strange. How did this work properly before then I wonder. I'll do some digging

.env
samanthaandrews: Yeah I agree. This will also be more of a standardized setup based on how other projects are being deployed here at the library. I know DRB uses env vars too.

PR #366, opened 2022-02-18
Fix: Check for "staging" branch instead of "qa"

samanthaandrews:
Bugsnag is being configured to categorize release stages based on the branch name. Since we recently changed the "qa" branch to "staging" we needed to update these checks in the code.


kristojorg (approved):
Looks good to me

src/components/__tests__/Footer.test.tsx
```diff
@@ -101,7 +101,7 @@ describe("toggling SimplyE Branding", () => {
     expect(iosbadge).toBeInTheDocument();
     expect(iosbadge).toHaveAttribute(
       "href",
-      "https://apps.apple.com/us/app/simplye/id1046583900"
+      "https://apps.apple.com/us/app/open-ebooks/id1033669250"
```
samanthaandrews, line 104: @kristojorg I updated these tests because they were failing but want to make sure that's what we are expecting. Are these values expected to be configured only for open-ebooks now or is something off?

kristojorg: No this is what we are expecting now. I thought I had updated this in my previous PR, but guess not?

PR #367, opened 2022-02-22
Improve loaned book load time

Toxiapo:
This PR adds a new API to the app for fetching fulfillment info and the manifest.
 
In conjunction with the built-in `prefetch` property from [next/link.](https://nextjs.org/docs/api-reference/next/link), loaned books will get a head start with the pre-fetching process in the background which _should_ improve the book load time when users click on the Read button.

- [x] new API /api/fulfillment/[fulfillmentUrl].tsx
- [x] Prefetch the new API in the book details and loans page with `<Head>`, the prefetching process should start as soon as the user logged in and got the loaned books from the server.
- [x] update unit and cypress test
- [x] create a new ticket to track book reads


src/components/bookDetails/__tests__/BookDetails.test.tsx
```diff
@@ -103,40 +103,6 @@ describe("book details page", () => {
     ).not.toBeInTheDocument();
   });
 
-  test("shows simplyE callout when NEXT_PUBLIC_COMPANION_APP is 'simplye'", async () => {
```
Toxiapo, line 106: @kristojorg  Do we still need this test and the test above with the companion setup? I think we are just keeping openebboks in the footer now.

src/components/FulfillmentButton.tsx
```diff
@@ -119,14 +120,33 @@ const ReadOnlineInternal: React.FC<{
...
+            window.location.origin
+          }/api/fulfillment/${details.url.replace("/read/", "")}`}
+        />
+      </Head>
+      <NavButton
+        // next/link prepends the librayId in the link already so just pass in the raw url
+        href={details.url}
+        {...getButtonStyles(isPrimaryAction)}
+        prefetch={true}
+        onClick={open}
```
kristojorg, line 145: I'm not sure we should be doing this anymore. It should just be a normal link without an `onClick` I think, and we should probably track the `openBook` event when the `/read` page is actually requested or navigated to. Maybe just remove this for now and make a new ticket to track book reads properly

src/components/bookDetails/__tests__/BookDetails.test.tsx
kristojorg: Yea we can remove this, and in a separate PR remove the code to show that callout

src/pages/api/fulfillment/[fulfillmentUrl].ts
```diff
@@ -0,0 +1,95 @@
...
+import { IS_DEVELOPMENT } from "utils/env";
+import { handleCMError, handleFetchError, json } from "dataflow/fetch";
+import track from "analytics/track";
+
+/**
+ * This is a handler for Open eBooks fulfillment and manifest.
+ * eg: http://localhost:3000/api/fulfillment/[fulfillmentUrl]
+ *
+ */
+const fulfillmentUrl: NextApiHandler = async (req, res) => {
```
Comment (kristojorg) [inline src/pages/api/fulfillment/[fulfillmentUrl].ts:14]: FYI I think this should be named `fulfillment`

src/pages/api/fulfillment/[fulfillmentUrl].ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+  }
+
+  // get the token
+  const credentials = getCredentials(req.cookies, "app"); //FIXME: get libraySlug
+  const token = credentials?.token;
+
+  if (!token) {
+    return res.status(401).json({
+      title: "Invalid Request",
```
Comment (kristojorg) [inline src/pages/api/fulfillment/[fulfillmentUrl].ts:31]: Title here should be `Not Authorized` and detail `You must be signed in to access this resource` or something

src/pages/api/fulfillment/[fulfillmentUrl].ts
```diff
@@ -0,0 +1,95 @@
...
+      detail: "Unknown Error",
+      status: 500
+    });
+    return;
+  }
+};
+
+export default fulfillmentUrl;
+
+const fetchWithBookValutId = (url: string, bookId: string) =>
```
Comment (kristojorg) [inline src/pages/api/fulfillment/[fulfillmentUrl].ts:80]: spelling

src/components/FulfillmentButton.tsx
Toxiapo: Ticket: OE-378

Toxiapo:
@kristojorg I was trying to do manifest caching with [this](https://flaviocopes.com/nextjs-cache-data-globally/) article, but I think we are actually don't need to? Vercel already helps us cache the API endpoint since we are keeping the /axisnow/[sibn] API. The manifest loads under a second after the first initial load.

src/components/bookDetails/__tests__/BookDetails.test.tsx
```diff
@@ -92,7 +92,7 @@ describe("book details page", () => {
     expect(utils.getByText("Publisher:")).toBeInTheDocument();
   });
 
-  test("does not show simplyE callout when NEXT_PUBLIC_COMPANION_APP is 'openebooks'", () => {
+  test("shows openebooks callout when NEXT_PUBLIC_COMPANION_APP is 'openebooks'", () => {
```
kristojorg, line 95: This can be removed as well

src/pages/api/fulfillment/[fulfillmentUrl].ts
```diff
@@ -0,0 +1,102 @@
...
+      title: "Invalid Request",
+      detail: "Missing fulfillmentUrl",
+      status: 400
+    });
+  }
+
+  if (!library) {
+    return res.status(400).json({
+      title: "Invalid Request",
+      detail: "Missing library"
```
Comment (kristojorg) [inline src/pages/api/fulfillment/[fulfillmentUrl].ts:29]: Missing status

src/components/FulfillmentButton.tsx
```diff
@@ -112,21 +111,29 @@ const ReadOnlineExternal: React.FC<{
...
-}> = ({ details, isPrimaryAction, trackOpenBookUrl }) => {
-  const router = useRouter();
-  const { buildMultiLibraryLink } = useLinkUtils();
+}> = ({ details, isPrimaryAction }) => {
+  const { slug } = useLibraryContext();
+  const fulfillmentUrl = new URL(
+    `/api/fulfillment/${details.url.replace("/read/", "")}`,
+    window.location.origin
+  );
+  fulfillmentUrl.searchParams.set(LIBRARY_QUERY_PARAM, slug);
```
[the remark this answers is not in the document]
Toxiapo: This button(Read) starts the prefetching process, so we need the `library` slug for the new api/fulfillment endpoint for it to properly return a response.

src/pages/api/fulfillment/[fulfillmentUrl].ts
```diff
@@ -0,0 +1,102 @@
...
+
+  if (!fulfillmentUrl) {
+    return res.status(400).json({
+      title: "Invalid Request",
+      detail: "Missing fulfillmentUrl",
+      status: 400
+    });
+  }
+
+  if (!library) {
```
Comment (kristojorg) [inline src/pages/api/fulfillment/[fulfillmentUrl].ts:26]: Is this used anywhere else? If not we can remove the requirement for it to have a library. I actually don't see where this query param would come from in the /api/fulfillment/fulfillmentUrl route

Comment (Toxiapo) [reply src/pages/api/fulfillment/[fulfillmentUrl].ts:26]: I was thinking we can remove the library variable after we finish removing the multi-library setup. This is currently needed because I am using the `getCredentials` function. I had to manually set the library with `fulfillmentUrl.searchParams.set()` whenever we call this route.

kristojorg:
> @kristojorg I was trying to do manifest caching with [this](https://flaviocopes.com/nextjs-cache-data-globally/) article, but I think we are actually don't need to? Vercel already helps us cache the API endpoint since we are keeping the /axisnow/[sibn] API. The manifest loads under a second after the first initial load.

Yea so I suppose I had thought we would remove the request to the `/api/manifest` route, and just copy the code to generate the manifest directly in to this new handler. But this works fine too! 

This is pretty HUGE savings. When you load a book indirectly (not by directly going to the url), the load time goes from ~7s to ~3s. On second load it is down to ~1s. That is a 2.3x speed increase for first load and 7x speed increase for second load! 🎉 

It will still help to get the CM response faster, and possibly make the webpub manifest generate faster as well, but this is an awesome step forward in UX. 

The one case we don't yet optimize for is direct loads (when you directly follow a link to a read page). There are a couple things we actually could do that I just realized:
- Put the fulfillment url in the browser response of the /read page, using response headers to prefetch. Works much like <link> elements in <head>, but simply provided in the `headers` of a response. This would get the browser to start fetching that content before react has even rendered. 
- We could make a non-blocking fetch request to the fulfillment api from the server when we receive a request to /read. Basically, fire one off in `getServerSideProps` or `middleware`, but don't wait for any response. That would just kick start the generation of the manifest. 

Not sure how much time either of those would save, but they're certainly worth a shot. Maybe in a new ticket.

Finally, last thing I am thinking about... I like your placement of the prefetch link adjacent to the fulfillment button, however, that prefetch can still take 6-9 seconds to resolve, at which point a user may have already clicked to the book itself before the prefetch is done. So I'm wondering if you think there is a place we could put those not just when we render a <FulfillmentButton> but any time we have a user's loans? Then we would start the prefetching even on the home page.

src/pages/api/fulfillment/[fulfillmentUrl].ts
Comment (kristojorg) [reply src/pages/api/fulfillment/[fulfillmentUrl].ts:25]: Another way to do this would have been making the route `/api/fulfillment/[library]/[fulfillmentUrl].ts`. This is fine too though

Toxiapo:
> > @kristojorg I was trying to do manifest caching with [this](https://flaviocopes.com/nextjs-cache-data-globally/) article, but I think we are actually don't need to? Vercel already helps us cache the API endpoint since we are keeping the /axisnow/[sibn] API. The manifest loads under a second after the first initial load.
> 
> Yea so I suppose I had thought we would remove the request to the `/api/manifest` route, and just copy the code to generate the manifest directly in to this new handler. But this works fine too!
> 
> This is pretty HUGE savings. When you load a book indirectly (not by directly going to the url), the load time goes from ~7s to ~3s. On second load it is down to ~1s. That is a 2.3x speed increase for first load and 7x speed increase for second load! 🎉
> 
> It will still help to get the CM response faster, and possibly make the webpub manifest generate faster as well, but this is an awesome step forward in UX.
> 
> The one case we don't yet optimize for is direct loads (when you directly follow a link to a read page). There are a couple things we actually could do that I just realized:
> 
> * Put the fulfillment url in the browser response of the /read page, using response headers to prefetch. Works much like  elements in , but simply provided in the `headers` of a response. This would get the browser to start fetching that content before react has even rendered.
> * We could make a non-blocking fetch request to the fulfillment api from the server when we receive a request to /read. Basically, fire one off in `getServerSideProps` or `middleware`, but don't wait for any response. That would just kick start the generation of the manifest.
> 
> Not sure how much time either of those would save, but they're certainly worth a shot. Maybe in a new ticket.
> 
> Finally, last thing I am thinking about... I like your placement of the prefetch link adjacent to the fulfillment button, however, that prefetch can still take 6-9 seconds to resolve, at which point a user may have already clicked to the book itself before the prefetch is done. So I'm wondering if you think there is a place we could put those not just when we render a but any time we have a user's loans? Then we would start the prefetching even on the home page.

Yea I am super excited about this! I will give the direct load and the new prefetching placement a try!

src/components/context/UserContext.tsx
```diff
@@ -128,7 +131,57 @@ export const UserProvider: React.FC = ({ children }) => {
     clearCredentials
   };
 
-  return <UserContext.Provider value={user}>{children}</UserContext.Provider>;
+  return (
+    <UserContext.Provider value={user}>
+      {children}
+      {/* Prefetch all the loaned books inside the Header for a faster load time when users click on a book */}
+      {isAuthenticated && data && (
+        <PrefetchFulfillmentBooks books={data} slug={slug} />
```
Toxiapo, line 139: I am starting the prefetching process as soon as the user login in, but I am not sure this works now. I know the prefetch starts in the background, but I now see some slowness on other page transitions, i.e. home > my books. (Maybe it's my computer, does it happen to you?)

Toxiapo:
> Put the fulfillment url in the browser response of the /read page, using response headers to prefetch. Works much like elements in , but simply provided in the headers of a response. This would get the browser to start fetching that content before react has even rendered.

@kristojorg  I assume you mean something like `res.setHeader("Link", "prefetching_link")`, but I am confused about where do I put this,  via the `getServerSideProps` function?

src/components/bookDetails/__tests__/BookDetails.test.tsx
Toxiapo: Ticket: OE-331

src/components/context/UserContext.tsx
Toxiapo: NVM it's just my internet

kristojorg:
> @kristojorg I assume you mean something like res.setHeader("Link", "prefetching_link"), but I am confused about where to put this, via the getServerSideProps function?

Yep you would add a `getServerSideProps` function that sets the header using `ctx.res.setHeader` before returning a value.

src/components/context/UserContext.tsx
```diff
@@ -128,7 +131,61 @@ export const UserProvider: React.FC = ({ children }) => {
     clearCredentials
   };
 
-  return <UserContext.Provider value={user}>{children}</UserContext.Provider>;
+  return (
+    <UserContext.Provider value={user}>
+      {children}
+      {/* Prefetch all the loaned books inside the Header for a faster load time when users click on a book */}
+      {isAuthenticated && data && !isOnReadPage(router) && (
```
kristojorg, line 138: Why not if we are on the read page?

kristojorg (approved):
Looks great, just the one little comment!

src/components/context/UserContext.tsx
Toxiapo: I removed it. This check is unnecessary, we should also prefetch the books on the read page actually.

PR #368, opened 2022-02-22
OE-214: Switch to First Book OAuth authentication method

samanthaandrews:
This PR makes the switch from First Book Basic Authentication to First Book OAuth. We have a new login type and endpoint from the authentication_document that we are using to fetch the JWT. This bearer token is stored as a cookie passed as an Authorization header to protected Circ Manager endpoints. These tokens expire after one hour, but we still store the basic token in the cookie and use that, if available, to re-fetch the JWT.

Todo:
- [x] Change login flow from submitting form to /api endpoint to existing signIn method. I think showing the loading state is important for users, especially young readers.
- [x] Add basic token to the cookie
- [x] Remove SAML auth method from the codebase
- [x] Remove Basic auth method from the codebase
- [x] In middleware pages, if a request returns a 401 response, re-request the jwt using the basicToken
- [x] Handle Clever tokens and Firstbook tokens the same way if possible <-- this is not currently possible since we don't have access to the Clever login credentials. Refresh tokens are also not implemented for Clever.
- [x] Update Cypress tests to use new login method
- [x] Ensure error shows for users if they enter invalid credentials & update Cypress test
- [x] Update unit tests
NOTE: In order to run locally you will need to set `API_URL=https://qa-circulation.openebooks.us` in your `.env.local` file.


src/pages/[library]/login/method/oauth.tsx
```diff
@@ -0,0 +1,109 @@
...
+        maxLength={25}
+        name={usernameInputName}
+      />
+      <FormInput
+        label={passwordInputName}
+        id="password"
+        type="password"
+        placeholder={passwordInputName}
+        name={passwordInputName}
+        required={true}
```
Comment (samanthaandrews) [inline src/pages/[library]/login/method/oauth.tsx:77]: By adding the `required` attribute, users can't submit the form if either the password or username are empty. If they submit the incorrect credentials, I'm setting the `?loginError=` query param which is being handled by the existing middleware that displays the error on the base login screen.

src/pages/api/auth/firstbook-oauth.ts
```diff
@@ -0,0 +1,72 @@
...
+    const cookie = JSON.stringify({
+      token: `Bearer ${accessToken}`,
+      methodType: OPDS1.OAuthType
+    });
+    // TODO: Why is this header getting set even if we don't return data and instead reach the catch block?
+    if (data) {
+      res.setHeader(
+        "Set-Cookie",
+        serialize(AUTH_COOKIE_NAME, cookie, { maxAge: data.expires_in })
+      );
```
samanthaandrews, line 47: this cookie is being set, but then disappears after the redirect. debugging this tomorrow.

src/pages/api/auth/firstbook-oauth.ts
```diff
@@ -0,0 +1,72 @@
...
+type JWTResponse = {
+  access_token: string;
+  token_type: "bearer";
+  expires_in: number;
+};
+
+const fetchJWTToken = async (token: string): Promise<JWTResponse> => {
+  // TODO: make this string into a Type or pass it from the req.body
+  const url =
+    "https://qa-circulation.openebooks.us/USOEI/http_basic_auth_token";
```
kristojorg, line 19: If we are going to hard-code these urls, which is technically ok since we are no longer multi-library, I think we need to:
- Make a file to keep these backend API endpoints in
- Make them relative urls so we can switch CMs easily (like `${CM_ENDPOINT}/http_basic_auth_token`)

However, I also think this is defined in the auth_document, so it should be available on `APP_CONFIG` unless I'm mistaken

src/pages/api/auth/firstbook-oauth.ts
```diff
@@ -0,0 +1,72 @@
...
+      Authorization: token
+    }
+  })
+    .catch(handleFetchError(url))
+    .then(handleCMError(url))
+    .then(json);
+};
+
+const handleLogin = async (req: NextApiRequest, res: NextApiResponse) => {
+  const token = generateToken(req.body["Access Code"], req.body["PIN"]);
```
kristojorg, line 32: For now, let's keep using the field labels in the auth document, so these field names will need to be based on that I think

src/pages/api/auth/firstbook-oauth.ts
kristojorg: I'm wrong, it is available in the auth_document, which populates `LibraryData`, but in order to get that you have to fetch the auth doc. We ultimately want to move to fetching the auth doc only once at app startup, but for the time being we don't have easy access to it here. We can potentially hardcode these values for now and make a ticket to use auth doc data once we fetch it only the one time

Toxiapo:
> `Dynamic Code Evaluation (e. g. 'eval', 'new Function') not allowed in Middleware`

Weirdd, I was able to build if i don't `import { handleCMError, handleFetchError, json } from "dataflow/fetch";` for the `fetchFirstbookJWT` function and just create those callback functions in the same file.

Maybe the `fetch.ts` file contains 3rd party plugins that use 'eval'?

samanthaandrews:
> > `Dynamic Code Evaluation (e. g. 'eval', 'new Function') not allowed in Middleware`
> 
> Weirdd, I was able to build if i don't `import { handleCMError, handleFetchError, json } from "dataflow/fetch";` for the `fetchFirstbookJWT` function and just create those callback functions in the same file.
> 
> Maybe the `fetch.ts` file contains 3rd party plugins that use 'eval'?

You're right, the fetch.ts file included the OPDS parser which was causing issues in middleware. I split out those handlers we need into a separate file and that seems to have fixed it. Thank you for looking into this!! @Toxiapo

samanthaandrews:
Cypress tests continue to fail in CI even when they are working when running locally. They do seem to be failing due to auth issues but it also seems like the same tests fail in the same way each time, which leads me to believe there is still something we can do on our end to remedy this. I'll look into it again on Monday.

next.config.js
```diff
@@ -96,6 +98,7 @@ const config = {
   env: {
     CONFIG_FILE: CONFIG_FILE,
     REACT_AXE: REACT_AXE,
+    API_URL: API_URL,
```
kristojorg, line 101: I think we should name this `NEXT_PUBLIC_CM_URL` actually, so we don't need to manually pass it through in the `next.config.js` like this. It can still be renamed to `CM_URL` in `utils/env.ts` though. But idk, if you like this pattern better then we can keep it instead of the `NEXT_PUBLIC` thing. Your choice

kristojorg, line 101: I think we should name this `NEXT_PUBLIC_CM_URL` actually, so we don't need to manually pass it through in the `next.config.js` like this. It can still be renamed to `CM_URL` in `utils/env.ts` though. But idk, if you like this pattern better then we can keep it instead of the `NEXT_PUBLIC` thing. Your choice

kristojorg (commented):
This looks great to me! You can actually ignore my comments about the `API_URL` env var because I renamed it in my follow up PR that does some login refactor work. Had a couple questions tho

next.config.js
samanthaandrews: yeah i think this is a good idea. i changed it to `NEXT_PUBLIC_CM_URL`.

cypress/tests/open-ebooks/integration/errors/useSwr.ts
```diff
@@ -108,14 +115,15 @@ describe("Mock endpoint that catches 401, 404, and 502 errors in the ErrorBounda
     cy.findByText(new RegExp(incorrectBookBorrow.detail)).should("exist");
   });
 
-  it("shoud redirect the users to the login page and display the error from the server on 401 error.", () => {
-    cy.visit(MOCK_ERROR.bookDetailPage[401].path);
+  // FIXME: this test is failing on staging... why?
```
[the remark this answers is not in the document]
samanthaandrews: https://jira.nypl.org/browse/OE-401

samanthaandrews:
I realized Cypress tests were failing in CI due to my erroneously encoding the cookie 🤦 They are now passing successfully. However, we do still run the risk of hitting those 502 errors from First Book (and we likely will still hit them on the daily cron tests) so it makes sense to continue to identify the root-cause and explore a permanent fix.

kristojorg (approved):
Looks good! Let's put this under review : )

cypress/tests/open-ebooks/integration/errors/useSwr.ts
```diff
@@ -108,14 +115,15 @@ describe("Mock endpoint that catches 401, 404, and 502 errors in the ErrorBounda
...
 
-    // FIXME: This is a workaround, useCredential hook should clear the cookie for the app
-    cy.clearCookies();
+  //   // FIXME: This is a workaround, useCredential hook should clear the cookie for the app
+  //   cy.clearCookies();
 
-    cy.findByText(/Invalid credentials/).should("exist");
-  });
+  //   cy.findByText(/Invalid credentials/).should("exist");
+  // });
```
Toxiapo, line 126: Q: Is this still failing with your latest update? If yes, this is also filed under OE-352 :), maybe we can link the two in Jira.

Toxiapo (approved):
🚀 🚀 🚀

PR #370, opened 2022-02-28
OE-374: Upgrade Axisnow Decryptor for fontdoor.axisnow.com

kristojorg:


samanthaandrews (approved):
I did some light manual testing on the preview branch and I'm able to open books. I can also see that content is now coming from the new `frontdoor` endpoint.

PR #377, opened 2022-03-05
OE-311: Upgrade Web Reader

kristojorg:
Just upgrades to use the new HTML renderer. I tested and it looks ok. Just waiting to merge that and use a full version.


cypress/support/utils.ts
```diff
@@ -111,7 +110,7 @@ export const ALMOST_MIDNIGHT_LOAN = `${READ_PATH}/${encodeURIComponent(
 export const ALMOST_MIDNIGHT_FETCH_URL = `${SERVER_URL}/works/Axis%20360%20ID/0017409891/borrow/15`;
 export const ALMOST_MIDNIGHT_REVOKE_URL = `${SERVER_URL}/loans/121907/revoke`;
 
-export const V2_IFRAME_SELECTOR = "#iframe-wrapper iframe";
+export const V2_IFRAME_SELECTOR = "#html-reader-iframe";
```
Toxiapo, line 113: I think we can remove the V2 prefix now.

PR #378, opened 2022-03-05
Fix: Don't bundle MSW in production

kristojorg:
I ran `ANALYZE=true npm run build` to check out the new web reader size, and it looks like we are bundling MSW with the `_app` chunk, which is ~2MB! It's because webpack is still pretty dumb when it comes to dead code elimination, and we had to separate out our `if (process.env.NODE_ENV === "development")` check into its own statement.


src/pages/_app.tsx
```diff
@@ -25,14 +25,15 @@ Router.events.on("routeChangeError", progress.finish);
...
  */
-if (
-  (process.env.NODE_ENV === "development" &&
-    process.env.NEXT_PUBLIC_API_MOCKING === "true") ||
-  (process.env.NEXT_PUBLIC_CYPRESS_ENV === "true" &&
-    process.env.NEXT_PUBLIC_API_MOCKING === "true")
-) {
-  console.warn("Using MSW to intercept network requests");
-  require("../../msw");
+if (process.env.NODE_ENV !== "production") {
```
samanthaandrews, line 28: Ugh yea, we should definitely not make this bundle in production. However, this NODE_ENV is set to "production" whenever we run `npm run build` which is what we run in CI to build it for Cypress tests. As far as I can tell there isn't a way for us to set this to something like "test" or another value. We need to use a different check. I noticed next has these different "phases" that we might be able to leverage, but because we need to run `next build` in CI we might need to come up with a different solution.

Documentation: https://nextjs.org/docs/api-reference/next.config.js/introduction
Phase constants: https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/constants.ts#L1-L5

kristojorg:
@samanthaandrews, it looks like this is now working except for the fact that we get 502 errors. We should probably make a ticket to mock _all_ integration tests so it never even hits the CM at all

samanthaandrews:
I am now realizing that this change isn't working as expected 😬 We can't actually set the NODE_ENV=test in this way because Next.js enforces `production` during next build to avoid performance de-optimizations. See this thread for context: https://github.com/vercel/next.js/issues/17032

I can make a PR tomorrow to revert this. I'm thinking we should use the recommended addition of `APP_ENV=test` and then ensure MSW is bundled if that is set along with `NEXT_PUBLIC_API_MOCKING=true`.

PR #381, opened 2022-03-08
OE-348: Login Refactors

kristojorg:
This PR includes some refactors to the login code. I tried to simplify as much as possible. 

1. Previously during clever login we would have the CM redirect the user to their final destination (like a book page), and then catch the `#access_token` in `useCredentials` and set it in a cookie. We no longer do that. Instead there is now a `/app/login/method/clever/callback` route which has the sole responsibility of being where the CM redirects to with the `#access_token`. It extracts the token, sets a cookie and redirects to the final destination. 
2. I use a static cookie name. We only support one library now, so I just set a constant for the name of our auth cookie, whereas before we generated it from the library slug.
3. I renamed `nextUrl` query param to `redirectTo`, as `nextUrl` name is used in Next.js middleware (`req.nextUrl`).
4. Add a logout url `/api/auth/logout` which uses a middleware to clear the auth cookie and redirect the user home. This removes that work from being performed client-side.

I tried to keep my commits clean so you can look through the PR that way if you'd like.


src/auth/LoginWrapper.tsx
```diff
@@ -2,32 +2,16 @@ import React from "react";
...
-  const { successPath } = useLoginRedirectUrl();
-  /**
-   * If the user becomes authenticated, we can redirect
-   * to the successUrl
-   */
-  React.useEffect(() => {
-    if (isAuthenticated) {
-      push(successPath);
-    }
-  }, [isAuthenticated, push, successPath]);
```
kristojorg, line 30: We now only do this on `/app/login/method/clever/callback`.

samanthaandrews:
This is looking really good and I like that we are removing some of that client login logic. Just a couple things I noticed while doing some manual testing:

Clever login doesn't seem to be recognizing the cookie or redirecting users to their root catalog page if they've never logged in before, i.e. on a fresh private browser tab. Video attached. Once I complete the Clever login flow a second time, it knows I'm logged in (I can see the Catalog option in the main header).

https://user-images.githubusercontent.com/34348332/157314154-28c54a2d-81e7-467d-a7a7-a069c0f58040.mov

I did get the 502 "Third party auth error" while logging in using First Book once, but the UI didn't display that error and the UI showed the loading spinner forever.

Toxiapo:
This looks great. I love the dedicated logout API and the credential refactor, it really simplifies a lot of the stuff!

samanthaandrews:
Did a little more testing today as I was working on removing the multi library setup and I noticed that reloading a book detail page causes the fulfillment button to display twice for some strange reason!

Screen Shot 2022-03-09 at 2 47 45 PM

kristojorg:
Okay @samanthaandrews thank you for that! I believe I fixed those problems and added more tests for them. I have now added the tests that were covered in my document [here](https://docs.google.com/document/d/1gW7-Q3WHajvtjgCB5flEx3QDVzcVOLKR44Ok9d56vRg/edit#). It includes a bunch of new login integration tests, so that's good I think!

Let me know what y'all think and we'll move it under QA ASAP

kristojorg:
@samanthaandrews I'm getting 502 errors trying to login during tests... I'm thinking maybe we try to make integration tests just never even hit the CM server. Always use fixtures. What do you think?

package.json
```diff
@@ -124,6 +125,7 @@
     "msw": "^0.36.3",
     "prettier": "2.5.1",
     "react-test-renderer": "^17.0.2",
+    "type-fest": "^2.12.0",
```
samanthaandrews, line 128: do we need this?

samanthaandrews, line 128: do we need this?

src/auth/OpenEbooksLoginPicker.tsx
```diff
@@ -1,18 +1,42 @@
...
+};
 
 export default function LoginRegion(): JSX.Element {
   const { authMethods } = useLibraryContext();
   const { query } = useRouter();
 
   const loginError = extractParam(query, LOGIN_ERROR_QUERY_PARAM);
+  const redirectTo = extractParam(query, LOGIN_REDIRECT_QUERY_PARAM);
+
+  const cleverUrl = { pathname: "/app/login/method/clever", query: {} };
```
samanthaandrews, line 34: nit: consider putting these pathnames into the `utils/constants` file

samanthaandrews, line 34: nit: consider putting these pathnames into the `utils/constants` file

src/utils/auth.ts
```diff
@@ -64,3 +88,40 @@ export const fetchFirstbookJWT = async (
...
+  const urlParams = new URLSearchParams(window.location.search);
+  const redirectTo =
+    urlParams.get(LOGIN_REDIRECT_QUERY_PARAM) ?? CATALOG_ROOT_PATH;
+
+  const finalPath = isValidRedirect(redirectTo)
+    ? redirectTo
+    : CATALOG_ROOT_PATH;
+
+  const finalRedirect = new URL(finalPath, window.location.origin);
+  window.location.href = finalRedirect.toString();
```
samanthaandrews, line 114: maybe a dumb question, but what is the difference between calling `router.push(pathname)` and assigning `window.location.href` like we do here?

samanthaandrews, line 114: maybe a dumb question, but what is the difference between calling `router.push(pathname)` and assigning `window.location.href` like we do here?

samanthaandrews (approved):
Just a few small questions, but otherwise this looks good. I've been doing manual testing as well and haven't run into issues 🎉

samanthaandrews:
> @samanthaandrews I'm getting 502 errors trying to login during tests... I'm thinking maybe we try to make integration tests just never even hit the CM server. Always use fixtures. What do you think?

This would be ideal. The only thing I can think to make this happen would be a never-expiring JWT token to use in the cookie for these tests, which we have talked about doing with Leonard. Otherwise we really do need to push for First Book to address the issue once and for all.

src/utils/swrConfig.ts
```diff
@@ -27,8 +31,12 @@ const makeSwrConfig = (
           }
         });
         console.warn("Caught 401 error in swr config");
-        const redirectUrl = `/${library.slug}/login?${LOGIN_ERROR_QUERY_PARAM}=${err.info.title}`;
-        clearCredentials(redirectUrl);
+        const redirectUrl = new URL(LOGIN_PICKER_PATH, window.location.origin);
+        redirectUrl.searchParams.set(LOGIN_ERROR_QUERY_PARAM, err.info.title);
+        // remove the cookie
+        Cookies.remove(AUTH_COOKIE_NAME);
```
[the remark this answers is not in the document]
samanthaandrews: I created https://jira.nypl.org/browse/OE-404 and assigned it to me. I'll fix this first thing next sprint. Weird that I didn't hit it in my manual testing when I was switching over to OAuth.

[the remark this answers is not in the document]
kristojorg: Yea good point... I wonder how normal OAuth implementations handle this? It seems like chaos to have to make sure every fetch request in every part of the app will refresh the token if it's expired.

package.json
kristojorg: Props not. I used it for the `JsonValue` type it has, but I actually think I switched to something else in the end. I can remove

src/utils/auth.ts
kristojorg: Not dumb. TBH I don't know the exact details of what `router.push` will do, but I know it uses some next.js specific logic and will usually do a client-side navigation. Setting window.location.href will completely refresh the page, and when changing auth settings it seemed appropriate to trust that mechanism more.

package.json
kristojorg: I did end up using it actually. It is generally a very useful library of typescript utility types that I think doesn't hurt to have around. Won't get bundled with app code, and isn't introducing a major dependency or anything

PR #382, opened 2022-03-08
[TEST] Explore prefetching on read page

Toxiapo:
Continue with https://github.com/NYPL-Simplified/circulation-patron-web/pull/367#issuecomment-1056018262.

- On the /read page, start prefetching the fulfillment request in the server.


src/pages/[library]/read/[bookUrl].tsx
```diff
@@ -42,11 +51,43 @@ const ReaderPage: NextPage<ReaderPageProps> = ({ library, error }) => {
...
+  const authDocUrl = getAuthDocUrl(librarySlug);
+  const authDocument = await fetchAuthDocument(authDocUrl);
+  const library = buildLibraryData(authDocument, librarySlug);
+
+  const credentials = getCredentials(ctx.req.cookies, librarySlug);
+
+  if (credentials?.token) {
+    ctx.res.setHeader(
+      "Link",
+      `${buildFulfillmentUrl(bookUrl, librarySlug)}; rel=prefetch`
```
Comment (Toxiapo) [inline src/pages/[library]/read/[bookUrl].tsx:77]: @kristojorg I am actually not sure if this is actually helping the load time when directly landing the read page, it's quite difficult to test these books, and fetching time is quite arbitrary (**A Boy, a Ball, and a Dog** book took about 50s to load while during the day it took about 10s). I assume it does because I see the pre-fetch header in the network tab.

Ideally, this should give the page a head start and begin the download in the server, and when the page starts loading, it will continue the download from there.

Toxiapo:
Let's come back to this at a later time. I am getting so many 504 errors with this update.

PR #384, opened 2022-03-09
[WIP] OE-387: Remove `APP_CONFIG` env

samanthaandrews:
To do:
- [x] Remove `APP_CONFIG` and fetch the authentication document at build time
- [ ] Ensure app fails with useful warning message if it cannot fetch the authentication_document <-- write Jest test for this
- [ ] Fix and update existing Jest tests


src/utils/fulfill.ts
```diff
@@ -186,7 +186,9 @@ export function getAppSupportLevel(
   contentType: OPDS1.AnyBookMediaType,
   indirectionType: OPDS1.IndirectAcquisitionType | undefined
 ): MediaSupportLevel | "unsupported" {
-  const mediaSupport = APP_CONFIG.mediaSupport;
+  const mediaSupport = JSON.parse(`{
```
samanthaandrews, line 189: I am thinking of storing a mediaSupport.json object at the root of the project that stores this information. Let me know if that makes sense to you or if you think it should be somewhere else. @kristojorg @Toxiapo (Previously this was part of the CONFIG_FILE yaml, but since we are getting rid of that file we need to put it somewhere else.)

.env
```diff
@@ -1,13 +1,21 @@
-# must set a config file
-CONFIG_FILE=/app_config/opene.yml
-
 # qa circulation manager url
 NEXT_PUBLIC_CM_URL=https://qa-circulation.openebooks.us
 
+# authentication document which will be fetched at build time
+AUTH_DOC_URL=https://qa-circulation.openebooks.us/USOEI/authentication_document
```
samanthaandrews, line 5: Should we also add the INSTANCE_NAME and DEFAULT_LIBRARY as values in this config or somewhere else within the app? I don't think we do, since I removed all references to these values in the code, but maybe they're useful to have for debugging purposes? Like maybe it's useful for us to know the instance name when we communicate with the server team... @kristojorg @Toxiapo 

This is what was in our config file previously:

```
instance_name: qa-web.openebooks.us
companion_app: openebooks
show_medium: false
libraries:
  app: "https://qa-circulation.openebooks.us/USOEI/authentication_document"
media_support:
  application/vnd.librarysimplified.axisnow+json: show
openebooks:
  default_library: app
```

src/utils/fulfill.ts
kristojorg: Yea that checks out to me. We could also just get rid of the code that depends on this and hard code the values in. But maybe that's for later. This is a good first step

.env
kristojorg: I don't think we need them anymore!

PR #385, opened 2022-03-11
OE 387 remove app config

kristojorg:


kristojorg:
Okay @samanthaandrews I _think_ this has all the necessary commits from your branch. Sorry about that! LMK if it's missing something. I just ended up cherry picking the applicable ones over to here

.env
```diff
@@ -1,13 +1,21 @@
-# must set a config file
-CONFIG_FILE=/app_config/opene.yml
-
 # qa circulation manager url
 NEXT_PUBLIC_CM_URL=https://qa-circulation.openebooks.us
 
+# authentication document which will be fetched at build time
+AUTH_DOC_URL=https://qa-circulation.openebooks.us/USOEI/authentication_document
```
kristojorg, line 5: I think this can be replaced by the `NEXT_PUBLIC_CM_URL` plus a `CM_LIBRARY_PATH="/USOEI"` constant and an `AUTH_DOC_PATH` constant

kristojorg: Or we could just use the `AUTH_DOC_URL` as the single input and everything else can be based off that:
```
new URL("/USOEI/feed/...", AUTH_DOC_URL)
```

kristojorg:
PS: Any jest tests that we have Cypress tests for already can simply be removed

mediaSupport.ts
```diff
@@ -0,0 +1,6 @@
+import { MediaSupportConfig } from "interfaces";
```
kristojorg, line 1: I would probably put this file next to our constants. Could even just inline this in to there too if you want.

src/components/BookCard.tsx
```diff
@@ -38,7 +37,7 @@ const BookCard = React.forwardRef<
         aria-label={`View ${book.title}`}
         sx={{ "&:hover": { textDecoration: "none" } }}
       >
-        <BookCover book={book} showMedium={APP_CONFIG.showMedium} />
+        <BookCover book={book} showMedium={false} />
```
kristojorg, line 40: We could remove the medium indicator, or make a separate ticket for that. Whichever you prefer

src/components/BookStatus.tsx
```diff
@@ -14,8 +13,7 @@ const BookStatus: React.FC<{ book: AnyBook }> = ({ book }) => {
     ? shouldRedirectToCompanionApp(book.fulfillmentLinks)
     : false;
 
-  const companionApp =
-    APP_CONFIG.companionApp === "openebooks" ? "Open eBooks" : "SimplyE";
+  const companionApp = "Open eBooks";
```
kristojorg, line 16: We can simply inline this value, it will never change now

src/__tests__/getStaticProps.test.tsx
```diff
@@ -96,18 +101,19 @@ describe("Test getStaticProps in the collection page", () => {
...
-        collectionUrl: "test-collection",
-        library: "lib1"
-      }
-    });
-
-    expect(promise).rejects.toThrowError(ServerError);
-    expect(promise).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"Third-party service failed.: The library could not complete your request because a third-party service has failed."`
-    );
+    try {
```
samanthaandrews, line 104: Not sure why my changes affected the `getStaticProps` test suite, but I had to make some changes in order to get it to even run. I then added these try/catch blocks to ensure the test didn't fail when the mocked fetchCollection threw a 502. I'm not sure this is what we want though. Let me know if you think I should make a small ticket to fix this.

src/dataflow/opds1/__tests__/parse.test.ts
```diff
@@ -7,9 +7,20 @@ import {
...
 import { bookIsBorrowable } from "utils/book";
+import { mockMediaSupport } from "test-utils/fixtures";
 const sanitizeHtml = require("dompurify").sanitize;
+import { mediaSupport } from "../../../utils/mediaSupport";
+
+jest.mock("utils/fulfill", () => ({
+  getAppSupportLevel: jest.fn()
+}));
+
+jest.mock("../../../utils/mediaSupport", () => ({
```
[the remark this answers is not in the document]
kristojorg: Hmmm yea I can take a look tomorrow!

src/__tests__/getStaticProps.test.tsx
kristojorg: Hmm yea I think we should make a ticket to review these. We probably shouldn't be mocking `fetchCollection` at all, but should be mocking `fetch` itself

samanthaandrews: I created https://jira.nypl.org/browse/OE-407 to track this.

src/dataflow/opds1/__tests__/parse.test.ts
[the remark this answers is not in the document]
kristojorg: Ok I basically changed one thing and removed a bunch of tests we don't need anymore. I also made a comment for miguel to make two more tickets to continue stripping cruft out of this area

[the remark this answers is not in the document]
samanthaandrews: Awesome, I think this is a good idea. The `parse` code is likely going to be modularized, correct?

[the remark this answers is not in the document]
kristojorg: Yea although I think it will change significantly if/when we decide to extract it out.

src/middleware/check-auth.ts
```diff
@@ -7,15 +7,15 @@ import {
...
 import { PATRON_ENDPOINT } from "utils/constants";
 
 function isLibraryPage(page: NextRequest["page"]): boolean {
   if (!page.name) return false;
   const lib = page.params?.library;
   if (!lib) return false;
-  const libs = { ...APP_CONFIG.libraries };
-  return libs[lib] !== undefined;
+  const openEBooksLibrary = AUTH_DOC.id;
+  return openEBooksLibrary !== undefined;
```
kristojorg, line 18: I don't think this is necessary. We previously were checking that the app config had a library set for this path. But now it seems like we should just check if we are on a `/app` url

kristojorg (commented):
A few comments

.env
```diff
@@ -1,13 +1,21 @@
-# must set a config file
-CONFIG_FILE=/app_config/opene.yml
-
 # qa circulation manager url
 NEXT_PUBLIC_CM_URL=https://qa-circulation.openebooks.us
+CM_LIBRARY_PATH=/USOEI
```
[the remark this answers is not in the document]
samanthaandrews: I agree. The issue here is we need to reference the `AUTH_DOC_URL` in the `next.config.js` file, which we haven't yet converted into an ECMAScript module, so it errors when we try to import or require the `/utils/constants.ts` file. Until we convert it (I'll make a ticket for this), I think the cleanest thing might be to just include the full AUTH_DOC_URL in the `.env` file. It's only referenced once in the `next.config.js` so this doesn't seem like we're introducing too much code duplication.

next.config.js
```diff
@@ -37,12 +39,8 @@ const BUILD_ID = encodeURIComponent(
   `${APP_VERSION}-${GIT_BRANCH}.${GIT_COMMIT_SHA}`
 );
 
-// fetch the config file synchronously. This will wait until the command exits to continue.
-const APP_CONFIG = JSON.parse(
-  execSync("node --unhandled-rejections=strict src/config/fetch-config.js", {
-    encoding: "utf-8"
-  })
-);
+const AUTH_DOC_URL =
```
kristojorg, line 42: Ah sorry we do need this to be variable per environment. What I would suggest is using `new URL("/USOEI/authentication_document", NEXT_PUBLIC_CM_URL).toString())` so that the base updates between QA/prod circulation managers

src/__tests__/getStaticProps.test.tsx
```diff
@@ -61,7 +66,7 @@ describe("Test getStaticProps in the collection page", () => {
   });
 
   test("Throw ServerError: 404 error returns props notFound:true", async () => {
-    fetchMock.mockRejectedValueOnce(
+    (fetchCollection as any).mockRejectedValueOnce(
```
kristojorg, line 69: I would rather we keep the original `fetchMock` so that this test _also_ covers the `fetchCollection` code. Otherwise if we change something in `fetchCollection` that breaks this we will have no idea

kristojorg: I'm curious why these tests needed to be changed at all actually

src/middleware/check-auth.ts
```diff
@@ -7,15 +7,13 @@ import {
...
-import { APP_CONFIG } from "utils/env";
 import { PATRON_ENDPOINT } from "utils/constants";
 
 function isLibraryPage(page: NextRequest["page"]): boolean {
   if (!page.name) return false;
   const lib = page.params?.library;
   if (!lib) return false;
-  const libs = { ...APP_CONFIG.libraries };
-  return libs[lib] !== undefined;
+  return true;
```
kristojorg, line 16: Just FYI this is going to fail when we change `/[library]/` to `/app`. We will simply have to test somehow to see if we are within `/app` I guess. Or we can look at how this is being used and figure out a different strategy

kristojorg (commented):
a few comments

next.config.js
samanthaandrews: good point. updating now.

src/middleware/check-auth.ts
samanthaandrews: Yeah, I will address this as part of https://jira.nypl.org/browse/OE-391

kristojorg:
Whoops probably should have waited to merge your other branch in to this, but oh well

src/__tests__/getStaticProps.test.tsx
samanthaandrews: They need to be changed because previously this `fetchMock` was hitting the very first promise that it called, which was the one fetching the auth document. Since we removed that promise, these tests are getting further into the chain of calls within `getPageStaticProps`, aka `fetchBook`, `fetchCollection`, and `fetchSearchData`. These calls are not throwing errors in the same way fetching the auth document was.

PR #386, opened 2022-03-11
OE 363: Login Flow Harmonization

kristojorg:
This PR implements the new Login Flow Harmonization design.

- [x] Implement UI update 
   - [x] Update Header logo and link section
   - [x] Add the Sign in and dedicated login path to the breadcrumb
   - [x] Update Main sign-in page
   - [x] Update Clever and First book form page
   - [x] Remove "back home button" on login page. [OE-373](https://jira.nypl.org/browse/OE-373)
- [x] Create static help page
- [x] Rename login -> signin eveyrwhere
- [x] Update TAD
- [x] Fix Cypress and unit test


src/utils/constants.ts
```diff
@@ -22,10 +22,10 @@ export const AUTH_COOKIE_NAME = "OEW_AUTH_COOKIE";
  * App paths
  * Fully qualified paths that include the "library" aka "/app" prefix.
  */
-export const LOGIN_PICKER_PATH = "/app/login";
-export const LOGIN_CLEVER_PATH = "/app/login/method/clever";
-export const LOGIN_CLEVER_CALLBACK_PATH = "/app/login/method/clever/callback";
-export const LOGIN_FIRSTBOOK_PATH = "/app/login/method/firstbook";
+export const LOGIN_PICKER_PATH = "/app/signin";
```
kristojorg, line 25: I think we should rename these variables too to `SIGNIN_PICKER_PATH` etc. In for the penny, in for the pound ya know?

src/auth/OpenEbooksLoginPicker.tsx
```diff
@@ -1,134 +1,15 @@
...
-  ) as ClientFirstbookMethod | undefined;
-
-  if (!cleverMethod || !firstbookAuthMethod)
-    throw new AppSetupError(
-      "Application is missing either Clever or Firstbook methods"
-    );
+import SigninOption from "./SigninOption";
 
+// This component is only used on the homepage, with a separate design from the login page
+export default function OpenEbooksLoginPicker(): JSX.Element {
```
kristojorg, line 4: I think this file is simple enough to not have it, unless this wrapper is used in multiple places?

src/components/Header.tsx
```diff
@@ -2,19 +2,22 @@ import React from "react";
...
+import OEBLogoText from "icons/OEBLogoText";
+import useUser from "./context/UserContext";
 
 const HeaderFC: React.FC<{ className?: string; searchData?: SearchData }> = ({
   className,
   searchData
 }) => {
   const library = useLibraryContext();
+  const { token } = useUser();
+  const userLoggedIn = !!token;
```
kristojorg, line 20: If a user lands on any /app path excluding /app/login paths, we can assume they are logged in and would like to try and not have logic checking their login status because it depends on fetching the loans endpoint which is slow. So I'd rather check the path instead to see if a user is on a /app/login path, hide these links if they are.

src/components/Header.tsx
```diff
@@ -23,56 +26,62 @@ const HeaderFC: React.FC<{ className?: string; searchData?: SearchData }> = ({
...
-          searchData={searchData}
-        />
-      </Stack>
-    </header>
+        <span sx={{ pr: [1, 3], py: 0 }}>
+          <OEBLogoBook />
+        </span>
+        <span sx={{ pl: [1, 3], py: 0 }}>
+          <OEBLogoText />
+        </span>
```
kristojorg, line 77: Do we need to wrap these in spans or can we pass these styles directly to the component?

src/components/Header.tsx
```diff
@@ -93,6 +102,7 @@ const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
           {link.title}
         </AnchorButton>
       ))}
+
       {helpWebsite && (
         <AnchorButton
```
kristojorg, line 107: I don't think we want this help website link anymore. We should remove that

src/icons/OEBLogoBook.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from "react";
+
+const OEBLogoBook = () => (
```
kristojorg, line 3: I would make these take a `className` so you can pass styles directly to the `svg` component

src/marketing/Header.tsx
```diff
@@ -79,7 +79,7 @@ export default function Header() {
         {isAuthenticated ? (
           <HeaderLink href="/app">Catalog</HeaderLink>
         ) : (
-          <HeaderLink href="/app/login">Log In</HeaderLink>
+          <HeaderLink href="/app/signin">Sign In</HeaderLink>
```
kristojorg, line 82: Can we change this href to use the `APP_SIGNIN_PICKER_PATH` constant?

src/middleware/check-auth.ts
```diff
@@ -51,7 +51,7 @@ export async function checkAuth(
         "Missing credentials. Redirecting to login",
         req.nextUrl.pathname
       );
-      const loginUrl = new URL(`/${library}/login`, req.url);
+      const loginUrl = new URL(`/${library}/signin`, req.url);
```
kristojorg, line 54: These can also use the `APP_SIGNIN_PICKER_PATH` constant now

src/pages/[library]/signin/method/clever/index.tsx
```diff
@@ -64,7 +64,7 @@ export const getServerSideProps: GetServerSideProps = async ctx => {
   const protocol = hostname.includes("localhost") ? "http" : "https";
 
   const fullSuccessUrl = new URL(
-    `${protocol}://${hostname}/app/login/method/clever/callback`
+    `${protocol}://${hostname}/app/signin/method/clever/callback`
```
Comment (kristojorg) [inline src/pages/[library]/signin/method/clever/index.tsx:67]: Let's switch this to use the constants as well

src/theme/theme.ts
```diff
@@ -30,7 +29,9 @@ const zIndices = {
 
 const shadows = {
   modal: "0px 2px 7px 3px rgba(0,0,0,0.3)",
-  focus: "0 0 0 0.2em rgba(0, 109, 255, 0.4)"
+  focus: "0 0 0 0.2em rgba(0, 109, 255, 0.4)",
+  authButton:
```
kristojorg, line 33: I think we can name this just `button` shadow. Hopefully it is reused in other places ultimately. I imagine we won't have multiple button shadows

src/components/Button/styles.ts
```diff
@@ -103,6 +103,33 @@ export const styleProps = (
         }
       };
 
+    case "auth":
```
kristojorg, line 106: I think rename this to "shadow" maybe? I'm hopeful this won't just be used as the auth button, but more generally as the "shadow" variant with the 3d styling

src/components/BreadcrumbBar.tsx
```diff
@@ -4,6 +4,19 @@ import List, { ListItem } from "./List";
...
+  // double check this
+  const isServerUrl = url.includes("/USOEI/");
+
+  if (isServerUrl) {
+    return <Link collectionUrl={url}>{text}</Link>;
+  }
+
+  return <Link href={url}>{text}</Link>;
+};
+
```
kristojorg, line 19: I'm not sure I totally understand this. Is it that most of the time we pass a `collectionUrl` in to the `BreadcrumbBar`, but now we would like to be able to pass non-collection urls like `/app`? If that's the case, then we should make the breadcrumb take _either_ a `collectionUrl` _or_ a `href` prop, and distinguish the two that way instead of trying to detect it based on something in the link string

src/auth/SigninWrapper.tsx
```diff
@@ -0,0 +1,43 @@
...
+}) => {
+  return (
+    <div
+      sx={{
+        flex: 1,
+        mb: 7
+      }}
+    >
+      <Head>
+        <title>{title ?? "Sign In"}</title>
```
kristojorg, line 27: We should probably just move this out to each individual sign in page and make every page declare its own `<Head>` component with individual titles

src/auth/SigninOption.tsx
```diff
@@ -0,0 +1,114 @@
...
+  if (redirectTo) {
+    firstbookUrl.query = { [LOGIN_REDIRECT_QUERY_PARAM]: redirectTo };
+    cleverUrl.query = { [LOGIN_REDIRECT_QUERY_PARAM]: redirectTo };
+  }
+
+  return (
+    <>
+      {loginError && (
+        <Flex
+          aria-label="Login Error"
```
kristojorg, line 30: This got mixed up in a merge somewhere, but the aria-label for this should be the `loginError` itself

src/pages/[library]/signin/index.tsx
```diff
@@ -0,0 +1,66 @@
+import { NextPage, GetStaticPaths, GetStaticProps } from "next";
+import LayoutPage from "components/LayoutPage";
+import withAppProps, { AppProps } from "dataflow/withAppProps";
+import LoginHeader from "auth/SigninHeader";
```
Comment (kristojorg) [inline src/pages/[library]/signin/index.tsx:4]: nit: rename SigninHeader

src/auth/SigninHeader.tsx
```diff
@@ -0,0 +1,19 @@
+import useLibraryContext from "components/context/LibraryContext";
+import { H1, H2, H3 } from "components/Text";
+import React from "react";
+
+// This header only shows up on the login page
```
kristojorg, line 5: Is this only used on `/app/signin` or also in `/app/signin/firstbook`? It looks like only the first to me, in which case we should just inline it there, not extract it to a separate file

src/__tests__/middleware.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { NextFetchEvent } from "next/server";
 import { libraryData } from "test-utils/fixtures";
 import { middleware as CollectionMiddleware } from "../../src/pages/[library]/collection/_middleware";
-import { middleware as LoginMiddleware } from "pages/[library]/login/_middleware";
+import { middleware as LoginMiddleware } from "pages/[library]/signin/_middleware";
```
kristojorg, line 4: nit: rename SigninMiddleware

src/components/BreadcrumbBar.tsx
Toxiapo: Originally, the breadcrumb component only renders collection page URLs, but we are adding sign-in page breadcrumb now so I had to modify this a bit. 

Yea, I think the type check is probably a better idea here.

src/auth/SigninOption.tsx
```diff
@@ -3,31 +3,26 @@ import { NavButton } from "components/Button";
...
+  const redirectQueryParam = redirectTo
+    ? `?${SIGNIN_REDIRECT_QUERY_PARAM}=${redirectTo}`
+    : "";
 
   return (
     <>
       {loginError && (
         <Flex
-          aria-label="Login Error"
+          aria-label={loginError}
```
kristojorg: Nit: missing one switch to `signinError`

src/auth/SigninOption.tsx
```diff
@@ -3,31 +3,26 @@ import { NavButton } from "components/Button";
...
-  const redirectTo = extractParam(query, LOGIN_REDIRECT_QUERY_PARAM);
+  const loginError = extractParam(query, SIGNIN_ERROR_QUERY_PARAM);
+  const redirectTo = extractParam(query, SIGNIN_REDIRECT_QUERY_PARAM);
 
-  const cleverUrl = { pathname: LOGIN_CLEVER_PATH, query: {} };
-  const firstbookUrl = { pathname: LOGIN_FIRSTBOOK_PATH, query: {} };
-  if (redirectTo) {
-    firstbookUrl.query = { [LOGIN_REDIRECT_QUERY_PARAM]: redirectTo };
-    cleverUrl.query = { [LOGIN_REDIRECT_QUERY_PARAM]: redirectTo };
-  }
```
kristojorg: Can we switch back to this method of creating the firstbook and clever urls? I think this got missed in a merge conflict but it was an intended change

src/components/Search.tsx
```diff
@@ -48,6 +48,8 @@ const Search: React.FC<SearchProps & SxProp> = ({
       throw new ApplicationError({
         detail: "Cannot perform search. No search template available."
       });
+
+    if (!value) return;
```
kristojorg, line 52: This is interesting, wondering how this popped up?

src/middleware/utils.ts
```diff
@@ -83,7 +83,7 @@ export async function refreshFirstbookJWT(
     console.log("Patron not authorized for resource, redirecting to login.");
     // Redirect to login with an error message
     const loginError = "Unkonwn Error Occurred. Please Relog.";
-    const loginUrl = new URL(LOGIN_PICKER_PATH, requestUrl);
+    const loginUrl = new URL(SIGNIN_PICKER_PATH, requestUrl);
```
kristojorg: Another `login` that snuck through

src/pages/[library]/signin/_middleware.ts
```diff
@@ -1,7 +1,7 @@
 import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
 import { getCredentials, isValidRedirect } from "utils/auth";
-import { LOGIN_ERROR_QUERY_PARAM, PATRON_ENDPOINT } from "utils/constants";
+import { SIGNIN_ERROR_QUERY_PARAM, PATRON_ENDPOINT } from "utils/constants";
 import { getLoginError, removeAuthCookie } from "middleware/utils";
```
Comment (kristojorg) [inline src/pages/[library]/signin/_middleware.ts:5]: One more `login`

src/pages/[library]/signin/method/clever/index.tsx
```diff
@@ -64,7 +67,7 @@ export const getServerSideProps: GetServerSideProps = async ctx => {
   const protocol = hostname.includes("localhost") ? "http" : "https";
 
   const fullSuccessUrl = new URL(
-    `${protocol}://${hostname}/app/signin/method/clever/callback`
+    `${protocol}://${hostname}${SIGNIN_PICKER_PATH}/method/clever/callback`
```
Comment (kristojorg) [inline src/pages/[library]/signin/method/clever/index.tsx:70]: We can replace everything including /method/clever/callback with `CLEVER_SIGNIN_CALLBACK_PATH`

kristojorg (commented):
Just a few nits, but otherwise looks great!

src/components/Search.tsx
Toxiapo: Oh yes! I forgot to ask you this yesterday. While testing the search form, the page gave me an OPDS feed fetching error if I press the search button without a search term.

src/components/Header.tsx
```diff
@@ -52,23 +51,21 @@ const HeaderFC: React.FC<{ className?: string; searchData?: SearchData }> = ({
...
-          mb: [1, 0]
-        }}
-      >
-        <OEBLogoBook sx={{ pr: [1, 3], py: 0 }} />
-        <OEBLogoText sx={{ pl: [1, 3], py: 0 }} />
-      </a>
-    </Link>
+    <a
+      href="/"
+      aria-label="Library catalog, back to homepage"
```
Toxiapo, line 56: using the generic anchor tag here because the `Link` component still uses the multi-library setup, I need this link to go to the `/` page.

kristojorg: You can you a `Link` component and pass a `href={{path: "/"}}` I believe. Or use `next/link`

kristojorg: Also I think this aria label should be changed to indicate where the link is headed: "Homepage" should be fine

kristojorg:
I updated to use /api/auth/signout instead of /api/auth/logout as well

src/auth/SigninOption.tsx
```diff
@@ -0,0 +1,122 @@
...
+          </div>
+          <NavButton
+            aria-label="Sign in with Clever"
+            variant="shadow"
+            sx={{
+              backgroundImage: 'url("/img/CleverBadge.png")',
+              color: "ui.cleverBlue",
+              "&:hover": {
+                color: "ui.cleverBlue"
+              }
```
kristojorg, line 69: This isn't the hover effect I am seeing. I am seeing a gray hover

src/auth/SigninOption.tsx
```diff
@@ -0,0 +1,122 @@
...
+          </div>
+          <NavButton
+            aria-label="Sign in with First Book"
+            variant="shadow"
+            sx={{
+              backgroundImage: 'url("/img/FirstBookBadge.png")',
+              color: "ui.firstBookBlue",
+              "&:hover": {
+                color: "ui.firstBookBlue"
+              }
```
kristojorg, line 92: Same think here with the gray hover

src/components/Header.tsx
Toxiapo: Thanks! That works. For some reason when I used `passHref` on next/link with a nested <a>, the linter started complaining, but it seems fine this time.

-- I will update the unit tests

src/components/__tests__/Header.test.tsx
```diff
@@ -0,0 +1,48 @@
+import { render } from "test-utils";
+import Header from "../Header";
+import { searchData } from "test-utils/fixtures/searchData";
+
+test("Not authenticated users should only see the OEB logo in the header.", () => {
```
kristojorg, line 5: In the future I think we should do these as cypress tests

src/pages/[library]/login/method/clever/_middleware.ts
```diff
@@ -1,17 +0,0 @@
...
-const callbackMiddleware: NextMiddleware = (req, _ev) => {
-  const loginError = req.nextUrl.searchParams.get(LOGIN_ERROR_QUERY_PARAM);
-  if (loginError) {
-    const url = new URL(LOGIN_PICKER_PATH, req.url);
-    url.searchParams.set(LOGIN_ERROR_QUERY_PARAM, loginError);
-    return NextResponse.redirect(url);
-  }
-};
-
-export default callbackMiddleware;
```
Comment (kristojorg) [inline src/pages/[library]/login/method/clever/_middleware.ts:17]: Does clever login error still work if you remove this?

src/pages/[library]/signin/_middleware.ts
```diff
@@ -49,18 +50,18 @@ export async function middleware(req: NextRequest) {
...
-    url.searchParams.set(LOGIN_ERROR_QUERY_PARAM, loginError);
+    url.searchParams.set(SIGNIN_ERROR_QUERY_PARAM, signinError);
     // we use redirect to add the search param, but if there isn't one added,
     // the user ends up in the same place anyway.
-    console.log("new login error, redirecting with query param", loginError);
+    console.log("new signin error, redirecting with query param", signinError);
     return removeAuthCookie(NextResponse.redirect(url));
   }
 
-  return removeAuthCookie(NextResponse.next());
```
Comment (kristojorg) [inline src/pages/[library]/signin/_middleware.ts:65]: We need to put this back. I have gotten trapped in the infinite redirect loop on this branch, and I don't think this was the cause of the previous error

kristojorg:
image

The clever and firstbook logos are a bit blurry at this size. Can we get larger versions, or else use SVGs?

kristojorg (commented):
A few changes

src/auth/SigninOption.tsx
Toxiapo: This is correct. We want the gray background but keep the text color cleverBlue

src/pages/[library]/login/method/clever/_middleware.ts
Comment (Toxiapo) [reply src/pages/[library]/login/method/clever/_middleware.ts:17]: We can ignore this. The login folder has been removed, we are using the "signin" folder from now on. `src/pages/[library]/sigin/*`

src/pages/[library]/signin/_middleware.ts
Comment (Toxiapo) [reply src/pages/[library]/signin/_middleware.ts:65]: ahhh yes i missed this. Updating!

PR #387, opened 2022-03-14
OE-341 dockerize

hokei:
Adds ECR publish on Staging and Production deploys


.github/workflows/release-qa.yml
```diff
@@ -0,0 +1,56 @@
+# On merge to development,
+# build a container and deploy to ECR
+name: Publish container to ECR
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
mwbenowitz, line 6: Just want to confirm that we want to run this on all of these actions since it will result in a decent number of builds. (Nothing wrong with this will just make the CI/CD run slower!)

Dockerfile
```diff
@@ -1,32 +1,48 @@
...
+ARG NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA
+ARG NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF
+
+ENV CONFIG_FILE app_config/opene.yml
+
+RUN npm run build:docker
+
+# RUNNER
+FROM node:16-alpine AS runner
+
```
mwbenowitz, line 24: In the previous version (the one we were working on), there was an `apk add git` step here. Did that turn out to be unnecessary? I was running into issues without it

next.config.js
```diff
@@ -15,15 +12,18 @@ const {
...
 
 const log = (...message) =>
   console.log(chalk.blue("app info") + "  -", ...message);
 
 // Compute some git info by running commands in a child process
 const execSync = require("child_process").execSync;
-const GIT_COMMIT_SHA = execSync("git rev-parse HEAD").toString().trim();
+const GIT_COMMIT_SHA =
+  NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA ??
+  execSync("git rev-parse HEAD").toString().trim();
```
mwbenowitz, line 26: oh, I see, I bet this is why `git` isn't necessary in the docker build process

Dockerfile
```diff
@@ -1,32 +1,48 @@
...
+# Automatically leverage output traces to reduce image size 
+# https://nextjs.org/docs/advanced-features/output-file-tracing
+COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
+COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
+
+USER nextjs
+
+EXPOSE 3000
+
+ENV PORT 3000
```
mwbenowitz, line 46: I'd just move this up to be with the other `ENV` variables just for consistency/clarity.

mwbenowitz (approved):
This looks good to me! I have a few questions/comments but this should get us a QA instance in ECS. 👍

.github/workflows/release-qa.yml
hokei: No, this is just for me to test right now.  I'm going to have to ask you again about how to get Tugboat up, because we will want that for all actions.

.github/workflows/release-qa.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Build, tag, and push image to Amazon ECR
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: openebooks-nextjs
+          IMAGE_TAG: ${{ github.sha }}
+          IMAGE_TAG_SHORT: ${{ steps.vars.outputs.sha_short }}
+          GH_PACKAGE_AUTH_TOKEN: ${{ secrets.GH_PACKAGE_AUTH_TOKEN }}
+
+        run: |
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG --build-arg NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA=$IMAGE_TAG_SHORT --build-arg NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF=$IMAGE_TAG_SHORT --build-arg GH_PACKAGE_AUTH_TOKEN=$GITHUB_PACKAGE_AUTH_TOKEN .
```
kristojorg, line 49: How do these environment variables get in here?

.tugboat/config.yml
```diff
@@ -0,0 +1,14 @@
...
+    expose: 3000
+    commands:
+      build:
+        - GH_PACKAGE_AUTH_TOKEN=$GH_PACKAGE_AUTH_TOKEN npm ci --cache .npm
+        - NODE_OPTIONS='--max-old-space-size=2048' npm run build
+      start:
+        - export CONFIG_FILE=app_config/opene.yml
+        - export PORT=3000
+        - export NEXT_TELEMETRY_DISABLED=1
+        - npm run start &
```
kristojorg, line 14: Do we need to set all env vars here? Or are they read from `.env` file? Also, we should set these env vars when running `npm run build` as well, as it depends on them.

Dockerfile
```diff
@@ -1,32 +1,48 @@
-# build environment
-FROM node:12.2.0-alpine as builder	
-
-RUN apk update
+# Install dependencies only when needed
+FROM node:16-alpine AS builder
+RUN apk add --no-cache libc6-compat
 RUN apk add git
```
kristojorg, line 4: I don't think we need this anymore. We add the git commit sha and ref when running the build command unless I'm mistaken.

.tugboat/config.yml
mwbenowitz: looks like the `CONFIG_FILE` var is set in the .env file, so that can probably get skipped. The other two might need to change depending on environment, so I think it's better to have them here. They can definitely get set before the build phase though!

.tugboat/config.yml
```diff
@@ -6,9 +6,8 @@ services:
     commands:
       build:
         - GH_PACKAGE_AUTH_TOKEN=$GH_PACKAGE_AUTH_TOKEN npm ci --cache .npm
-        - NODE_OPTIONS='--max-old-space-size=2048' npm run build
+        - PORT=3000 NEXT_TELEMETRY_DISABLED=1 NODE_OPTIONS='--max-old-space-size=2048' npm run build
       start:
-        - export CONFIG_FILE=app_config/opene.yml
         - export PORT=3000
         - export NEXT_TELEMETRY_DISABLED=1
         - npm run start &
```
kristojorg, line 15: Is this supposed to have an ampersand at the end of it?

.github/workflows/release-qa.yml
```diff
@@ -0,0 +1,56 @@
+# On merge to development,
+# build a container and deploy to ECR
+name: Publish container to ECR
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  publish_qa:
```
kristojorg, line 9: Is there a corresponding way to release production?

hokei: yeah I'm gonna copy this over to a `release-production.yml` once we have that set up

.tugboat/config.yml
mwbenowitz: it is! It's a quirk in Tugboat, it needs it to keep that process up and running, otherwise it exits the container for some reason

kristojorg:
@hokei is this ready to merge or still in progress? Just for the sake of my notifications, can we move it to draft status until it's ready to review (unless it is now)?

Dockerfile
```diff
@@ -1,32 +1,48 @@
...
+
+# only install production 
+RUN GH_PACKAGE_AUTH_TOKEN=$GH_PACKAGE_AUTH_TOKEN npm ci --cache .npm
+
+# expose the git SHA and branch as env vars
+# we use the same names that vercel uses for cross-compatibility in deployment
+ARG NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA
+ARG NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF
+
+ENV CONFIG_FILE app_config/opene.yml
```
samanthaandrews, line 19: Since we have been removing multi-library setup in recent weeks, we no longer need this CONFIG_FILE env var. The app is now fetching the open ebooks authentication_document at build time.

Because of this, there are a few other additions to the env vars, which are in the current `.env` file. I'm not sure if this docker file picks up variables from that file?

We do need to add the following env vars to the production build script:

NEXT_PUBLIC_CM_URL=https://circulation.openebooks.us
NEXT_PUBLIC_GTM_ID=XXXXXX @kristojorg do you know if this google tag manager id is somewhere in our Vercel builds? if not i don't think we need to include it
NEXT_PUBLIC_BUGSNAG_API_KEY=XXXXXX @kristojorg should be able to get you this key

and potentially NEXT_PUBLIC_ENABLE_MSW=false only if this setup isn't picking up variables from the `.env` file directly.

samanthaandrews (changes requested):
We've changed env vars in recent weeks 😣, so that is the only thing I see that needs to be changed!

Dockerfile
hokei: Ah I was going to comment that it was interesting that the staging deploy was working without these env variables but Jiayong noticed that we can't read books, which makes sense!  I'm gonna put this PR back into draft mode until I meet with Greg on Monday to fix the env variables on the CF template.

samanthaandrews:
I'm thinking we should also update the README with the latest deploy to prod instructions. I'm also curious about what the rollback process looks like, and if possible let's include these instructions in the README too.

README.md
```diff
@@ -33,29 +33,33 @@ Finally, every PR in this repository has a unique preview deployment so proposed
 
 - [circulation-patron-web](#circulation-patron-web)
   - [Background](#background)
-  - [Demo](#community-demo)
+  - [Community Demo](#community-demo)
```
hokei, line 36: I'm not quite sure how this ToC is being generated?  Deleting stuff from this only results it stuff being added back on save.

hokei:
@samanthaandrews this is ready for review again.  Can you also verify that the deploy to https://staging-docker.openebooks.org/ contains the latest Staging?

README.md
samanthaandrews: I think this ToC is auto-generated based on whatever headings exist in the README.md? But I could be wrong about that

Dockerfile
```diff
@@ -1,32 +1,48 @@
...
+
+# only install production 
+RUN GH_PACKAGE_AUTH_TOKEN=$GH_PACKAGE_AUTH_TOKEN npm ci --cache .npm
+
+# expose the git SHA and branch as env vars
+# we use the same names that vercel uses for cross-compatibility in deployment
+ARG NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA
+ARG NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF
+ARG NODE_ENV
+RUN if [ "$NODE_ENV" = "production" ]; \
```
samanthaandrews, line 19: Does this `$NODE_ENV` get set to production only in the `release-prod.yml` github action? Just curious because I know that anytime we run `npm run build`, the NODE_ENV is set to `production` by default. That doesn't seem to be what's happening here, since on https://staging-docker.openebooks.org/ I see we are hitting qa-circulation endpoints, but just want to be sure.

samanthaandrews: Ah i see that we're not running the build:docker command until after setting NEXT_PUBLIC_CM_URL, so yeah that NODE_ENV must be getting set in the release-prod.yml action.

samanthaandrews (approved):
Confirmed that the code running on https://staging-docker.openebooks.org/ is the most up to date, since I can see this page https://staging-docker.openebooks.org/contribute.html is correctly getting redirected to `https://staging-docker.openebooks.org/partners` which is a recent change that we merged to staging.

Dockerfile
hokei: NODE_ENV gets set here, I think: https://github.com/NYPL-Simplified/circulation-patron-web/pull/387/files#diff-7c7339ed739b954dcbb42e0ce08bfb4425576f550e3e6a06204ef9bcc96ac102R54

kristojorg: I would rather we not change the `NODE_ENV` variable when running in a "production like" environment. We currently rely on this env var being "production" in quite some places, and I know that in some environments (such as a real testing env) it's difficult if not impossible to actually change the `NODE_ENV`. I think instead of this if/else we can simply pass in the `NEXT_PUBLIC_CM_URL` as an env variable itself

README.md
kristojorg: I also don't remember haha

Dockerfile
```diff
@@ -1,32 +1,48 @@
...
+        then NEXT_PUBLIC_CM_URL=https://circulation.openebooks.us npm run build:docker; \
+        else NEXT_PUBLIC_CM_URL=https://qa-circulation.openebooks.us npm run build:docker; fi
```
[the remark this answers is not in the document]
hokei: I'll make a follow up!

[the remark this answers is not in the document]
hokei: I wasn't able to get this done before meetings started up today. Thanks Kristo for filing https://jira.nypl.org/browse/OE-501

PR #388, opened 2022-03-15
OE-388: Remove fulfillment buttons

samanthaandrews:
- Removes all `<FulfillmentButton />` components except `<ReadOnlineInternal />` and renames to `<ReadOnline />`
- Removes fulfillment types that aren't `fulfillable`, `borrowable`, and `unsupported`. These are the only 3 we support in Open eBooks.


src/utils/fulfill.ts
```diff
@@ -71,29 +50,6 @@ export function getFulfillmentFromLink(link: FulfillmentLink): AnyFullfillment {
...
-    case OPDS1.ExternalReaderMediaType:
-      return {
-        id: link.url,
-        type: "read-online-external",
-        getUrl: constructGetUrl(indirectionType, contentType, link.url),
-        buttonLabel: "Read Online"
-      };
-
     case OPDS1.AxisNowWebpubMediaType:
       if (!AXISNOW_DECRYPT) {
```
Toxiapo, line 54: I think we only support AxisNow going forward? maybe we can just return this early instead of using a switch case.

src/components/BookList.tsx
```diff
@@ -266,13 +244,17 @@ const BookListCTA: React.FC<{ book: AnyBook }> = ({ book }) => {
     const singleFulfillment =
       showableFulfillments.length === 1 ? showableFulfillments[0] : undefined;
 
-    if (singleFulfillment && !shouldRedirectUser) {
+    if (
+      singleFulfillment &&
+      !shouldRedirectUser &&
```
Toxiapo, line 249: we can remove the redirect check here, we no longer support companion apps.

Toxiapo (commented):
I believe we also have some stories for the removed book status, let's get rid of them too!

src/components/ReadOnline.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return isPrimaryAction
+    ? ({
+        variant: "filled",
+        color: "brand.primary"
+      } as const)
+    : ({
+        variant: "ghost",
+        color: "ui.gray.extraDark"
+      } as const);
+}
```
kristojorg, line 18: We can probably get rid of this `isPrimaryAction` prop. Seems like it should be the only option now

kristojorg, line 18: We can probably get rid of this `isPrimaryAction` prop. Seems like it should be the only option now

src/interfaces.ts
```diff
@@ -204,6 +204,8 @@ export type AnyBook =
   | FulfillableBook
   | UnsupportedBook;
 
+export type OpenEBook = BorrowableBook | FulfillableBook | UnsupportedBook;
```
kristojorg, line 207: In the future changes we can make this the type of `AnyBook`. We no longer need the other options

samanthaandrews: Yes; I was thinking about making this change in this PR, but decided to do it as part of one of the upcoming tickets so that the change will be easier to review.

kristojorg (approved):
lgtm!

PR #389, opened 2022-03-18
Fix mock-service-worker not running in Github action

Toxiapo:
When looking at the failing tests on https://github.com/NYPL-Simplified/circulation-patron-web/pull/368, I realized that GH never ran the msw plugin. This PR is to make sure we run msw when NEXT_PUBLIC_API_MOCKING is set and make sure it doesn't bundle the plugin in production.


src/pages/_app.tsx
```diff
@@ -25,11 +25,9 @@ Router.events.on("routeChangeError", progress.finish);
 /**
  * We can mock our backend api with MSW (mock service worker).
  */
-if (process.env.NODE_ENV !== "production") {
-  if (process.env.NEXT_PUBLIC_API_MOCKING === "true") {
-    console.warn("Using MSW to intercept network requests");
-    require("../../msw");
-  }
```
Toxiapo, line 32: [Per this comment](https://github.com/vercel/next.js/issues/17032#issuecomment-691491353). We cannot override the value of NODE_ENV when running `npm run build`, and it's set to be 'production' by default, so we never ran msw when running our cypress test. I think the `NEXT_PUBLIC_API_MOCKING` should be fine here, and tell webpack not to bundle in the next.config.js file

kristojorg:
Thanks for this @Toxiapo ! @samanthaandrews also made a PR with the same fix. I made a couple changes to her PR and posted a comment there. Gonna merge that one and close this one. 

closed by https://github.com/NYPL-Simplified/circulation-patron-web/pull/390

PR #390, opened 2022-03-21
Fix MSW bundling

samanthaandrews:
Only require MSW if NEXT_PUBLIC_API_MOCKING is set to true


src/pages/_app.tsx
```diff
@@ -25,11 +25,9 @@ Router.events.on("routeChangeError", progress.finish);
...
  * We can mock our backend api with MSW (mock service worker).
  */
-if (process.env.NODE_ENV !== "production") {
-  if (process.env.NEXT_PUBLIC_API_MOCKING === "true") {
-    console.warn("Using MSW to intercept network requests");
-    require("../../msw");
-  }
+if (process.env.NEXT_PUBLIC_API_MOCKING === "true") {
+  console.warn("Using MSW to intercept network requests");
+  require("../../msw");
```
samanthaandrews, line 30: maybe i'm wrong about this, but can't we just simplify the logic and only bundle msw if NEXT_PUBLIC_API_MOCKING is set to true? this value should only be true if a dev adds it to their `.env.local` file or within the cypress build commands. confused as to how it was ever getting bundled in prod tbh.

src/analytics/track.ts
```diff
@@ -3,7 +3,6 @@ import ApplicationError, { ServerError } from "errors";
 /* eslint-disable camelcase */
 import { NextWebVitalsMetric } from "next/app";
 import { APP_CONFIG } from "utils/env";
-import { AxisNowDecryptionError } from "@nypl-simplified-packages/axisnow-access-control-web";
```
kristojorg, line 6: While analyzing the bundle I realized that importing this was making our bundle size huuuuge. I need to update how we build that package and web-reader to fix some things. For now, though, we don't get much utility out of importing it here so I removed.

PR #394, opened 2022-03-25
OE-391: Renaming [library] to app

samanthaandrews:
- Rename `/[library]` folder to `/app` (hardcoding that one path)
- Remove `buildMultiLibraryLink` utility
- Remove references to `librarySlug` and hardcode `/app` instead


src/hooks/useLinkUtils.ts
```diff
@@ -1,34 +1,32 @@
 import useLibraryContext from "components/context/LibraryContext";
 
-const trailingSlashRegex = /\/$/;
+export const trailingSlashRegex = /\/$/;
 
 export default function useLinkUtils() {
-  const { slug, catalogUrl } = useLibraryContext();
-
-  function buildMultiLibraryLink(href: string): string {
```
samanthaandrews, line 8: i removed this `buildMultiLibraryLink` utility since we no longer need it, but kept the buildCollectionLink and buildBookLink since they are still being used in a couple different places.

as far as our custom `<Link />` component is concerned, in the ticket it was suggested we remove it, but since it is still using `buildCollectionLink` and `buildBookLink`, I'm thinking we should keep it as is for now. when i remove `libraryContext` I'll refactor the <Link /> component and refactor `buildCollectionLink` and `buildBookLink` so they're no longer a React hook.

src/pages/app/collection/[collectionUrl].tsx
```diff
@@ -43,7 +43,7 @@ export const getStaticProps: GetStaticProps = withAppProps<PageProps>(
     if (collectionUrl === library.catalogUrl) {
       return {
         redirect: {
-          destination: `/${slug}`,
+          destination: "/app",
```
Comment (kristojorg) [inline src/pages/app/collection/[collectionUrl].tsx:46]: We should probably have a constant for this

kristojorg (approved):
Hell ya lgtm!

src/pages/app/collection/[collectionUrl].tsx
Comment (kristojorg) [reply src/pages/app/collection/[collectionUrl].tsx:46]: But doesn't need to happen in this PR either.

Comment (kristojorg) [reply src/pages/app/collection/[collectionUrl].tsx:46]: Ultimately I think some centralized object to get all the different routes or build up a route for a given book/collection would be a useful thing to have.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ### UNRELEASED CHANGES
 
-- Chore: OE-331: Remove APP_CONFIG env and unnecessary FulfillmentButton components
+- Chore: OE-331: Remove APP_CONFIG env, unnecessary FulfillmentButton components, and hardcode /app path
```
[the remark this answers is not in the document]
samanthaandrews: I think I'm just going to change this to "Remove multi-library setup" in my base PR, OE-311 and add the no-changelog tag to these branches. Ultimately I think it's best to merge all of this as one change and one simple line-item in the changelog.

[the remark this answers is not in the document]
kristojorg: I actually disagree, I think we should keep these as separate commits personally with individual line items. I'd rather merge your base PR first and then this one so we can do that. Just because they're very wide-ranging changes and the more we can split it up the better if we have some bug and need to find which commit caused it

PR #395, opened 2022-03-25
OE-404: Fix `jwt` refresh logic in middleware and swr

samanthaandrews:
- Add logic to only attempt to refresh the First Book token if the problem doc returns with type `https://librarysimplified.org/terms/problem/expired-http-basic-bearer-token`
- Add refresh logic to the swr config to ensure users aren't being asked to log in again when their basic credentials are available and can be used to get a new jwt
- Add unit tests


samanthaandrews:
Closing this to refactor the refresh logic into a utility

PR #396, opened 2022-03-27
OE-409: Remove `withAppProps` and `LibraryContext`

samanthaandrews:
We previously used `withAppProps()` to fetch the auth document, build library data, and provide that to every page. Since we now fetch the auth document at build time, we can set a constant `LIBRARY=buildLibraryData(AUTH_DOC)`, and use that one LIBRARY constant everywhere, no need to pass it around in context anymore.

https://jira.nypl.org/browse/OE-409


src/utils/auth.ts
```diff
@@ -26,16 +25,6 @@ import {
   LOGOUT_PATH
 } from "./constants";
 
-/**
- * Extracts an array of auth providers from the authentication document,
- * and adds an ID to them
- */
-export function flattenAuth(authDoc: OPDS1.AuthDocument): AppAuthMethod[] {
```
samanthaandrews, line 33: Moved this to `src/dataflow/getLibraryData.ts` (where it is being called) because it cannot exist in a file that also accesses the env constants before they're initialized. This is because `getLibraryData()` is now being called at build time.

cypress/support/utils.ts
```diff
@@ -140,15 +140,15 @@ const bookDetailPage = {
 
 const readPage = {
   401: {
-    path: `${READ_PATH}/${encodeURIComponent(GN_BOOK_ENDPOINT)}`,
+    path: `${READ_PATH}/${encodeURIComponent(GN_FULFIL_ENDPOINT)}`,
```
kristojorg, line 143: nit: spelling

kristojorg: nvm apparently they are both correct spellings I just looked it up lol

src/components/Error.tsx
```diff
@@ -48,7 +46,7 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
               </Link>{" "}
             </>
           )}
-          <Link href={`/${library ?? ""}`}>
+          <Link href={`/app ?? ""}`}>
```
kristojorg, line 49: I don't think we need the `??` here anymore. Should just link to `/app`

src/pages/app/collection/[collectionUrl].tsx
```diff
@@ -68,8 +64,45 @@ export const getStaticProps: GetStaticProps = withAppProps<PageProps>(
...
+            },
+            e as Error
+          );
+    track.error(error, { severity: "error" });
+    return {
+      props: {
+        error: error.info
+      },
+      revalidate: 1
+    };
```
Comment (kristojorg) [inline src/pages/app/collection/[collectionUrl].tsx:103]: Maybe we should extract this into a utility or something?

kristojorg (changes requested):
Yeeehaw this is nice! Just a couple comments

src/pages/app/collection/[collectionUrl].tsx
Comment (samanthaandrews) [reply src/pages/app/collection/[collectionUrl].tsx]: done!

src/components/WebReader.tsx
```diff
@@ -63,6 +63,15 @@ const ReaderComponent: React.FC = () => {
...
+  // const [searchQuery, setSearchQuery] = useState<ParsedUrlQuery | null>(null);
+  // const bookUrl = extractParam(searchQuery, "bookUrl");
+
+  // useEffect(() => {
+  //   if (router.isReady) {
+  //     setSearchQuery(router.query);
+  //   }
+  // }, [router.query, router.isReady]);
+
   if (!bookUrl) {
```
samanthaandrews, line 75: I noticed Cypress errors in the useSwr test, and after fixing some of those paths and endpoints, I realized the test failure was actually identifying a bug where the `query` from `useRouter()` is undefined upon first render, triggering this bookUrl to be undefined. This doesn't happen if you navigate to the page from the Read button, but it does happen if I reload `read` the page directly, as described in this [question](https://stackoverflow.com/questions/69412453/next-js-router-query-getting-undefined-on-refreshing-page-but-works-if-you-navi) on StackOverflow.

We weren't hitting this error in the past because we were previously running `getStaticProps(`) on the component that calls this `<WebReader />` component. According to the [Next docs](https://nextjs.org/docs/advanced-features/automatic-static-optimization#how-it-works) it looks like this means we might want to be leveraging this router.isReady object, but I wanted to check to make sure this is the approach we want to use? Alternatively we could use getServerSideProps to return context.query.

samanthaandrews: I ended up using getServerSideProps because the code seemed more readable than using `router.isReady` and putting another useEffect hook in the WebReader component. Lmk what you think.

kristojorg: So the downside to this is that it means we have to run server-side code for every time we view a web reader page, whereas previously it was statically generated on the server and then the client runs code to fetch the resources it needs. I do hear you that this is more readable, but it also seems like a silly thing to use server side rendering for.

It's annoying that the router object isn't ready on first render, but we have had this problem in other places as well and resorted to `router.isReady`. So I think we should do the same here. I would like the eventually statically render the web-reader's nicer loading indicator here so that users only have one loading screen they see immediately when they request a reader page.

samanthaandrews: Ok this makes sense. I updated it to use `router.isReady` instead of `getServerSideProps()`.

kristojorg (approved):
Awesome!

PR #402, opened 2022-04-04
OE-408: Remove mediaSupportLevel

samanthaandrews:
https://jira.nypl.org/browse/OE-408


src/dataflow/opds1/parse.ts
```diff
@@ -200,33 +136,19 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
...
-
   const relatedLinks = entry.links.filter(isRelatedLink);
   const relatedLink = relatedLinks.length > 0 ? relatedLinks[0] : null;
 
   const fulfillmentLinks = acquisitionLinks
     .filter(link => {
-      return link.rel === OPDSAcquisitionLink.GENERIC_REL;
+      return (
+        link.rel === OPDSAcquisitionLink.GENERIC_REL &&
+        link.type === AxisNowWebpubMediaType
```
samanthaandrews, line 146: Here is where I am filtering out any book that doesn't match the Axis Now type

src/dataflow/opds1/parse.ts
```diff
@@ -80,77 +77,16 @@ function isTrackOpenBookLink(link: OPDSLink) {
...
-  return getAppSupportLevel(contentType, indirectionType);
-}
-
 function buildFulfillmentLink(feedUrl: string) {
   return (link: OPDSAcquisitionLink): FulfillmentLink => {
-    const { contentType, indirectionType } = parseFormat(link);
-    const supportLevel = getAppSupportLevel(contentType, indirectionType);
     return {
-      supportLevel,
+      supportLevel: "show",
```
samanthaandrews, line 83: Because we've already filtered out everything that isn't the Axis Now media type, this `supportLevel` key/value pair isn't super useful anymore. I'm thinking I'll remove this too (& simplify this `buildFulfillmentLink()` function), unless @kristojorg you see value in it still being here?

src/dataflow/opds1/parse.ts
```diff
@@ -312,10 +189,12 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
     };
   }
 
-  // it is unsupported
+  // it's fulfillable
```
samanthaandrews, line 192: I removed the "unsupported" book type because we filter out anything that isn't Axis Now above. So books with the OpenEBook type can either be `fulfillable` or `borrowable`. Lmk if this makes sense to you or if you think we should keep the `unsupported` book type.

src/interfaces.ts
```diff
@@ -15,17 +15,13 @@ export { OPDS1 };
...
 export type IndirectMediaSupport = Partial<
   Record<OPDS1.IndirectAcquisitionType, DirectMediaSupport>
 >;
 
-export type MediaSupportLevel =
-  | "show"
-  | "redirect"
-  | "redirect-and-show"
-  | "unsupported";
+export type MediaSupportLevel = "show";
```
samanthaandrews, line 24: removing `supportLevel` means I would also remove this type

src/dataflow/opds1/parse.ts
```diff
@@ -200,33 +136,19 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
   const borrowLink = getBorrowLink(acquisitionLinks);
   const { availability, holds, copies } = borrowLink ?? {};
 
-  const openAccessLinks: FulfillmentLink[] = acquisitionLinks
```
samanthaandrews, line 203: I removed `openAccessLinks` because it didn't seem like any of the books in the app had them. Lmk if this doesn't seem right to y'all.

src/dataflow/opds1/parse.ts
```diff
@@ -312,10 +206,10 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
...
   }
 
-  // it is unsupported
+  console.error("Unsupported book type found in this entry");
   return {
-    status: "unsupported",
-    ...book
+    ...book,
+    status: "unsupported"
   };
```
kristojorg, line 213: Maybe we should also track to bugsnag

src/interfaces.ts
kristojorg: Oh yes true I made this comment above too. So yep let's do it

src/interfaces.ts
```diff
@@ -116,13 +111,9 @@ export type BookMedium =
   | "http://schema.org/EBook"
   | "http://schema.org/Book";
 
-export type BookAvailability =
-  | "available"
-  | "unavailable"
-  | "reserved"
-  | "ready";
+export type BookAvailability = "available";
```
kristojorg, line 114: Can get rid of this too

src/utils/book.tsx
```diff
@@ -37,7 +37,7 @@ export function getAuthorsString(book: AnyBook): string {
   return authorsArray.join(", ");
 }
 
-export function availabilityString(book: AnyBook) {
+export function availabilityString(book: OpenEBook) {
```
kristojorg, line 40: We can actually totally remove the availability string as well. You can do it here or in a separate PR if you wish

src/utils/fulfill.ts
```diff
@@ -1,12 +1,6 @@
...
-  FulfillmentLink,
-  MediaSupportLevel,
-  OPDS1
-} from "interfaces";
+import { FulfillableBook, FulfillmentLink, OPDS1 } from "interfaces";
 import { bookIsAudiobook } from "utils/book";
 import { AXISNOW_DECRYPT } from "utils/env";
-import { mediaSupport } from "./mediaSupport";
 
 /**
```
kristojorg, line 5: Comment is now outdated

kristojorg (commented):
Looks good, just a few comments!

src/components/MyBooks.tsx
```diff
@@ -3,43 +3,15 @@ import { BookList } from "./BookList";
...
+import { OpenEBook } from "interfaces";
 import PageTitle from "./PageTitle";
 import { SignOut } from "./SignOut";
 import useUser from "components/context/UserContext";
 import { PageLoader } from "components/LoadingIndicator";
 
-const availableUntil = (book: AnyBook) =>
-  book.availability?.until ? new Date(book.availability.until) : "NaN";
-
-function sortBooksByLoanExpirationDate(books: AnyBook[]) {
```
kristojorg, line 15: I think we should still have this

src/dataflow/opds1/__tests__/parse.test.ts
```diff
@@ -165,16 +159,17 @@ const basicInfo = {
   categories: [factory.category({ label: "label" })],
   summary: "summary"
 };
-test("doesn't include borrow links without supported formats", () => {
+test("uses indirect acquistion links to determine if book is borrowable", () => {
```
kristojorg, line 162: It wouldn't hurt to add a test here that it filters out unsupported books entirely

src/dataflow/opds1/parse.ts
```diff
@@ -435,24 +321,34 @@ export function feedToCollection(
...
         }
       } else {
-        books.push(book);
+        track.error(
+          {
+            name: "Unsupported book",
+            message: "A non-AxisNow book format has been detected in the feed"
+          },
+          { severity: "warning" }
+        );
```
kristojorg, line 351: I probably would put this inside `isSupportedOpenEBookLink` so that it is called anywhere we use that function (for example when just fetching a single book instead of a whole feed)

src/utils/constants.ts
```diff
@@ -1,11 +1,5 @@
 import { CM_URL } from "./env";
 
-/**
- * the default assumption is that a book is available. See:
- * https://github.com/NYPL-Simplified/Simplified/wiki/OPDS-For-Library-Patrons#opdsavailability---describing-resource-availability
- */
-export const DEFAULT_AVAILABILITY = "available";
-
```
kristojorg, line 8: Are we sure we can remove this?

src/components/MyBooks.tsx
samanthaandrews: Should we still have the `sortBooksByLoanExpirationDate()` too? It seemed that AxisNow books didn't have the `since` and `until` properties.

I do think I need to add back the compareTitles() sorting that I removed here. Good catch.

src/dataflow/opds1/__tests__/parse.test.ts
samanthaandrews: Done!

src/dataflow/opds1/parse.ts
```diff
@@ -195,26 +147,17 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
...
+  const acquisitionLinks = entry.links.filter(isSupportedOpenEBookLink);
 
-  const relatedLinks = entry.links.filter(isRelatedLink);
-  const relatedLink = relatedLinks.length > 0 ? relatedLinks[0] : null;
+  const borrowLink = getBorrowLink(acquisitionLinks);
+
+  const { availability, holds, copies } =
+    getAvailability(acquisitionLinks) ?? {};
 
   const fulfillmentLinks = acquisitionLinks
```
samanthaandrews, line 162: I think we should consider making another ticket to refactor fulfillmentLinks from an array into a single object, since I think for our use-case, there's only ever one fulfillment link. Let me know if that makes sense or if in the future we expect there to be multiple fulfillment links that we'll need to support.

src/dataflow/opds1/parse.ts
```diff
@@ -195,26 +147,17 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
...
-        supportLevel
-      };
-    });
+  const acquisitionLinks = entry.links.filter(isSupportedOpenEBookLink);
 
-  const relatedLinks = entry.links.filter(isRelatedLink);
-  const relatedLink = relatedLinks.length > 0 ? relatedLinks[0] : null;
+  const borrowLink = getBorrowLink(acquisitionLinks);
+
+  const { availability, holds, copies } =
```
samanthaandrews, line 159: This fixes the availability info being undefined. Now the info about when a book will be returned shows up for all loaned books 👍

src/dataflow/opds1/parse.ts
```diff
@@ -244,7 +181,7 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
       ...availability,
       // we type cast status because our internal types
       // are stricter than those in OPDSFeedParser.
-      status: availability?.status as BookAvailability
+      status: availability?.status as "available"
```
samanthaandrews, line 184: Ok @kristojorg I know we talked about removing "borrowable" and "fulfillable" statuses in favor of `availability.status`, but for AxisNow books, both fulfillable and borrowable books list "available" as the status for OPDSAcquisitionLinks. I hesitate to assign a different status value here because I don't think that would keep us tracking with the OPDS spec.

It looks like fulfillable books include `since` and `until` properties, while with borrowable books these values are undefined. So we _could_ check for the presence of `since` and `until` to know a book's status, but I think it's much clearer to assign "borrowable" and "fulfillable" as we're already doing, which is why I've kept it as is for now.

samanthaandrews: Alternatively, we could make the entire availability object undefined for fulfillable books? And so if a book hasn't yet been borrowed, it won't have `availability.status` at all.

src/dataflow/opds1/parse.ts
kristojorg: Yep this makes sense for us. We should take a look at our `OpenEBook` type as a whole and consolidate it down into the simplest possible thing, and only parse out of the OPDS the items that we need as well. There are also a couple items in the OPDS which _arent_ properly parsed and we parse out of `raw` later, which we should stop doing. Can do that all in one ticket probably. "Review and update internal models" ticket or something

src/dataflow/opds1/parse.ts
kristojorg: I trust your judgement here. I honestly can't decide if we should try to make our model as close to OPDS as possible or try to simplify and flatten it as much as possible into an object that specifically suits our use case. I initially thought the former, but the latter is seeming more and more appealing

src/dataflow/opds1/parse.ts
samanthaandrews: Ok I created a ticket for this and assigned it to myself: https://jira.nypl.org/browse/OE-479. I can probably pick it up in the sprint after this one.

src/dataflow/opds1/parse.ts
samanthaandrews: I think keeping "borrowable" and "fulfillable" makes the most sense. I agree that it seems flattening our model into an object that specifically suits our use case is the more appealing option right now, since future-us will have a simpler model to work with. I've done this with apps that consume data & feed it back to the Mapbox api, and it has worked well because the Mapbox api is well-documented. We can always refer to back to the OPDS spec when we need to add features that require us to update our internal model.

kristojorg: Ok yep makes sense to me

src/dataflow/opds1/__tests__/parse.test.ts
```diff
@@ -21,6 +21,59 @@ const detailLink = factory.completeEntryLink({
   href: "/complete-entry"
 });
 
+test("feedToCollection doesn't include books that aren't AxisNow type", () => {
```
kristojorg, line 24: This seems to test that it includes books which have a axisnow fulfillment link, but not that it excludes ones that are missing it. Unless I'm missing something?

src/dataflow/opds1/__tests__/parse.test.ts
```diff
@@ -21,6 +21,59 @@ const detailLink = factory.completeEntryLink({
...
+    series: {
+      name: "Fake Series",
+      position: 2
+    },
+    language: "en"
+  });
+
+  const acquisitionFeed = factory.acquisitionFeed({
+    id: "some id",
+    entries: [entry],
```
kristojorg, line 69: I think you would want one entry with an axisnow fulfillment link and one without, and then make sure that the one without doesn't make it to the collection

src/dataflow/opds1/__tests__/parse.test.ts
```diff
@@ -186,9 +234,9 @@ test("doesn't include borrow links without supported formats", () => {
...
 
-  expect(bookIsBorrowable(book)).toBeFalsy();
-  expect((book as any).borrowUrl).toBeUndefined();
-  expect(book.status).toBe("unsupported");
+  expect(bookIsBorrowable(book)).toBeTruthy();
+  expect((book as any).borrowUrl).toBeDefined();
+  expect(book.status).toBe("borrowable");
 });
 
 test("chooses the first supported borrow link", () => {
```
kristojorg, line 242: We can remove this, it should not _only_ choose the axisnow borrow link.

src/dataflow/opds1/parse.ts
kristojorg: I don't fully understand why it changed from `getBorrowLink` to `getAvailability`. Is it because the link has a generic rel not a borrow rel in our case?

kristojorg (commented):
just a couple comments

src/dataflow/opds1/__tests__/parse.test.ts
samanthaandrews: You're absolutely right. Fixing it now.

src/dataflow/opds1/parse.ts
samanthaandrews: Yeah that's exactly it. I just pushed up a change that refactors this a little bit to make it easier to understand. Basically, we get the availability information from the acquisitionLink. Whether it's a fulfillment link or a borrow link. Previously we were only looking for the information in the borrow link, which is why we were returning `undefined` for availability, when it should have been returning information about when a loan expires.

src/dataflow/opds1/parse.ts
```diff
@@ -195,41 +143,25 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
...
 
-  const supportedFulfillmentLinks = fulfillmentLinks.filter(
-    link => link.supportLevel !== "unsupported"
-  );
+  const acquisitionLink = entry.links.find(isSupportedOpenEBookLink);
 
-  const revokeUrl = findRevokeUrl(entry.links);
+  let availability, holds, copies, borrowLink, fulfillmentLink;
+  if (acquisitionLink) {
+    ({ availability, holds, copies } = acquisitionLink);
```
samanthaandrews, line 155: I think now this is a more clear representation of where availability information is coming from @kristojorg

kristojorg: Yea this looks good to me, although it calls `buildFulfillmentLink` whether it is a borrow link or not, is that intended or should there be an `else` statement in there?

kristojorg (approved):
Just the one comment about building the fulfillment link. So long as that's intended, I approve!

src/dataflow/opds1/parse.ts
samanthaandrews: Good catch, I just added an `else`.

PR #403, opened 2022-04-05
Fix theming error on books

Toxiapo:
This PR updates the webreader package which includes the fix for the theming error


samanthaandrews (approved):
Did a manual test and The Mouse and The Moon book no longer has the theming error 👍

PR #404, opened 2022-04-07
OE-437: Clever credential bug

Toxiapo:
This PR fixes the bug where _middleware would cache the signout endpoint making the app not able to sign out the second time. 

Including the following refactors:

1. Signout via fetch request with POST method instead of Link, make sure nothing is cached. 
2. Remove _middleware from the signout API and just use a regular handler, this is because _middleware does not accept the POST method


src/components/SignOut.tsx
```diff
@@ -34,8 +46,9 @@ export const SignOut: React.FC<SignOutProps> = ({
           </Button>
           <NavButton
             color="ui.error"
-            href={{ pathname: SIGNOUT_PATH }}
             aria-label="Confirm Sign Out"
+            href={{ pathname: SIGNOUT_PATH }}
+            onClick={signOutHandler}
```
kristojorg, line 51: How come we decided not to go with the `<form>`? Also, if we do go this route, we should probably use a regular `<Button>` and not pass an `href`

src/pages/api/auth/signout/index.ts
```diff
@@ -0,0 +1,9 @@
+import { NextApiHandler } from "next";
+import { AUTH_COOKIE_NAME } from "utils/constants";
+
+const Signout: NextApiHandler = (_req, res) => {
+  res.setHeader("Set-Cookie", `${AUTH_COOKIE_NAME}=""; path=/; Max-Age=0`);
```
kristojorg, line 5: Is this exactly the header that was being set previously by `res.cookie.clear()` in the middleware? I know different browsers handle these differently unfortunately, and I thought we also needed to set an expires time in the past

kristojorg: Sweet thanks! Can also check the network response on staging if that's easier : )

Toxiapo: Yes, good idea, that's should be easier.

src/components/SignOut.tsx
Toxiapo: Yeah, `form` makes more sense i think. I had some problems using the middleware with the form, but since we are just using the normal API route now, let's go back to using the `form` tag.

Toxiapo: NVM! we need the handler function to clear the localstorage.

src/pages/api/auth/signout/index.ts
```diff
@@ -0,0 +1,12 @@
+import { NextApiHandler } from "next";
+import { AUTH_COOKIE_NAME } from "utils/constants";
+
+const Signout: NextApiHandler = (_req, res) => {
+  res.setHeader(
+    "Set-Cookie",
+    `${AUTH_COOKIE_NAME}=""; Path=/; Expires: ${new Date(1)}; Max-Age=0`
```
Toxiapo, line 7: Weird, I have to use `Max-Age` or else cypress won't delete the cookie for me.

kristojorg (approved):
Ok looks good, nice!

PR #405, opened 2022-04-08
OE-404: Refactor refresh token logic

samanthaandrews:
Refactors refresh token logic into a `verifyToken(credentials)` utility that can be called on the browser or server-side. On the browser side, this function is being called inside `getCredentialsBrowser()` which is being called in a new hook called `useGetcredentials()` so that the token can be accessed within the top-level of React components.

This means that First Book tokens are now potentially refreshed (if within 5 minutes of expiration) on each page load (via middleware) or on every component mount that accesses the `token`.


samanthaandrews:
@kristojorg There is one edge-case that I haven't yet handled, but would like to in this PR, with your input on the best way forward. The `verifyToken()` function is currently being run in middleware anywhere `getCredentials()` is happening. It's also being run anywhere the `token` is accessed via the UserContext. This catches most of the places, and has been working well in all of my manual testing.

But let's say a user leaves the open ebooks tab open for a few hours, and then returns to it and immediately clicks the "borrow" or "return" buttons. At that point, the token from the UserContext has expired, so they receive the error that the token has expired.

Paths forward:
1. Run the verifyToken() function again within the borrow/return hook.
2. Let users hit this edge case, but encourage them to reload the page to try again if they hit this error by adding a "Refresh" button.

Number 1 is nice because it means users probably won't ever see the token expiration error, but I hesitate to scatter verifyToken() in too many places throughout the app. In this case, though, perhaps it makes sense.

Toxiapo:
> 1. Run the verifyToken() function again within the borrow/return hook.
> 2. Let users hit this edge case, but encourage them to reload the page to try again if they hit this error by adding a "Refresh" button.
> 
> Number 1 is nice because it means users probably won't ever see the token expiration error, but I hesitate to scatter verifyToken() in too many places throughout the app. In this case, though, perhaps it makes sense.

FWIW, an expiration error message is reasonable for apps that require authentications, but since we are going to make FB users stay logged in _indefinitely_, Number 1 sounds like the approach here.

kristojorg:
I think we should do option 1. I would say we should actually just make it so that the only way we ever access the token is by calling an async function like `getToken` which internally verifies and possibly refetches it. I would pass that down in context from `UserContext` instead of the `token` itself.

src/utils/auth.ts
```diff
@@ -115,3 +114,27 @@ const INVALID_REDIRECTS = [
...
+    const now = Date.now();
+    const expiration = new Date(existingExpiration).valueOf();
+    const FIVE_MIN = 5 * 60 * 1000; // convert 5 minutes to milliseconds
+    if (expiration - now <= FIVE_MIN) {
+      const { token, expiration } = await fetchFirstbookJWT(
+        credentials.basicToken
+      );
+      return { ...credentials, token, expiration };
+    }
+  }
```
kristojorg, line 139: I would probably make this return the original credentials if they don't need to be refreshed so that consumers of the function don't need to have the added step of checking for undefined

kristojorg (commented):
Just a couple thoughts

src/utils/auth.ts
```diff
@@ -63,28 +61,29 @@ export const fetchFirstbookJWT = async (
...
-    .then(json);
+    })
+      .catch(handleFetchError(FIRSTBOOK_OAUTH_JWT_ENDPOINT))
+      .then(handleCMError(FIRSTBOOK_OAUTH_JWT_ENDPOINT))
+      .then(json);
+  // eslint-disable-next-line camelcase
+  const expirationInMilliseconds = expires_in * 1000;
+  const expiration = new Date(
+    Date.now() + expirationInMilliseconds
+  ).toUTCString();
```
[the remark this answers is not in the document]
samanthaandrews: `new Date()` creates a date object in the user's time zone, and `toUTCString()` is the most straightforward way to convert that value to UTC that is a human readable value and not just a number in milliseconds. since I didn't want to be concerned with a user's time zone at all, this seemed like the best value to pass through. ultimately all we need is a value in UTC milliseconds anyway since that's what we use to compare when we verify the token.

i will update to use `expiresAt` i agree it is more explicit!

src/hooks/useBorrowOrReturn.ts
```diff
@@ -39,7 +40,15 @@ export default function useBorrowOrReturn(fulfillmentType: FulfillmentType) {
...
     clearError();
-    if (!token) {
+    /**
+     * Instead of using the token from the UserContext, we get it directly
+     * from the cookies and verify it below to ensure we are using
+     * the most up-to-date token. Covers the edge-case where a user
+     * could leave their browser tab open and the token could expire without
+     * being refreshed upon clicking the Borrow or Return button.
+     */
+    const currentCredentials = getCredentialsBrowser();
```
kristojorg, line 50: Should we just use `useVerifyToken` here instead of manually verifying it before using it?

src/hooks/useBorrowOrReturn.ts
```diff
@@ -78,10 +96,10 @@ export default function useBorrowOrReturn(fulfillmentType: FulfillmentType) {
 
       // optimistically update the loans cache to remove the book,
       // and update the cached book itself to remove fulfillment links
-
+      const updatedCredentials = await verifyToken(currentCredentials);
```
kristojorg, line 99: This is the duplication I mean. I can see this easily being forgotten. Seems it would be better to have only one blessed way to access the token and it automatically verifies it

src/hooks/useVerifyToken.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * A React hook that wraps the verifyToken() function so consumers can
+ * more easily access the most up-to-date token.
+ * Gets credentials from browser cookies & verifies that the token has not expired
+ * Returns the latest credentials from getCredentialsBrowser
+ */
+export default function useVerifyToken() {
+  const credentials = getCredentialsBrowser();
+  React.useEffect(() => {
+    console.log("credentials have changed inside useEffect hook");
```
kristojorg, line 13: Stray log

src/utils/auth.ts
```diff
@@ -62,29 +60,31 @@ export const authTypeToPath: Record<OPDS1.OpenEBooksAuthType, string> = {
 export const fetchFirstbookJWT = async (
   basicToken: string
 ): Promise<FirstbookBearerJWT> => {
+  console.log("fetchFirstbookJWT");
```
kristojorg, line 63: Stray log

src/utils/auth.ts
```diff
@@ -115,3 +115,40 @@ const INVALID_REDIRECTS = [
...
+    const now = Date.now();
+    const expiration = new Date(existingExpiration).valueOf();
+    const FIVE_MIN = 5 * 60 * 1000; // convert 5 minutes to milliseconds
+    if (expiration - now <= FIVE_MIN) {
+      const { token, expiresAt } = await fetchFirstbookJWT(
+        credentials.basicToken
+      );
+      // if we're in the browser, set the new cookie using js-cookie
+      if (!IS_SERVER) {
+        console.log("!IS_SERVER");
```
kristojorg, line 141: stray logs

src/utils/auth.ts
```diff
@@ -115,3 +115,40 @@ const INVALID_REDIRECTS = [
 export function isValidRedirect(redirectTo: string) {
   return !INVALID_REDIRECTS.includes(redirectTo);
 }
+
+/**
+ * Checks First Book credentials to determine if a user's token
+ * can be refreshed. If credentials are not First Book or
+ * the token is not within 5 minutes of expiration, return the
+ * existing credentials
```
kristojorg, line 123: Should note here that in browser context it will also update the cookie. In server context it will simply return the new credentials

kristojorg (changes requested):

src/hooks/useBorrowOrReturn.ts
samanthaandrews: Yes this is the crux of the issue that I've been thinking about and I agree with you 100%. I created `useVerifyToken()` as a hook to wrap the `verifyToken()` async function so that tokens will be verified whenever we need to access the token within the top-level of a React component the same way we've always done it. `useVerifyToken()` is being called in UserContext, so a token is verified whenever a component mounts. But this doesn't cover the edge-case for when a component has mounted and been sitting for over 55 minutes. In that case, it needs to be verified again when a user clicks the "Borrow" or "Return" buttons.

So now I'm thinking I should more change the way we access tokens within React components. Perhaps it's best to  only verify & pass the token through to a fetch function directly, instead of verifying it on component mount. I think it would help me to can talk it through with you tomorrow or whenever you're feeling better!

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -43,31 +31,13 @@ describe("BorrowableBook", () => {
     ).toBeInTheDocument();
   });
 
-  test("shows loading state when borrowing, borrows, and revalidates loans", async () => {
```
samanthaandrews, line 46: Removed some of these assertions in favor of testing this logic directly in the <Borrow> component tests.

src/components/__tests__/Return.test.tsx
```diff
@@ -0,0 +1,133 @@
+import {
```
samanthaandrews, line 1: Added a similar set of tests for <Return />

Toxiapo (approved):
The code looks good to me, I think we can pass this to QA.

src/hooks/useGetCredentials.ts
```diff
@@ -0,0 +1,24 @@
...
+export default function useGetCredentials() {
+  const [credentials, setCredentials] = useState<
+    OpenEBooksAuthCredentials | undefined
+  >();
+  React.useEffect(() => {
+    (async () => {
+      const updatedCredentials = await getCredentialsBrowser();
+      setCredentials(updatedCredentials);
+    })();
+  }, []);
```
kristojorg, line 21: Ok my one question here is whether we need to do anything to this react state when the credentials in the cookie change? Or are we sure that the only time we do that is before a browser refresh? I think that is true, like in login and logout, but am not 100% positive.

kristojorg (commented):
Just the one question, but otherwise I think this is good!

src/hooks/useGetCredentials.ts
samanthaandrews: I don't think we need to do anything to this state when the credentials in the cookie change. I see what you're saying though, like if there are instances where within a component we are setting or clearing the credentials using js-cookie we would probably need to account for that in this hook. But currently we are only setting & unsetting auth cookies on login and logout.

kristojorg (approved):
Looks good!

PR #410, opened 2022-04-19
Release 4.7.8

samanthaandrews:
There were a bunch of merge conflicts when I tried to directly merge staging into production, so I created a feature branch (`release-4.7.8`) off of `production`, merged in `staging`, and fixed the conflicts. If you change the base branch in this PR to `staging` you'll see that this feature branch and `staging` are the same, except for the updated version number in the `package-lock.js` file.

I think this is due to a couple factors:

1. It looks like the last time we pushed to prod, we "Squashed and merged" when we should have just added all the commits and merged. This is hard to remember to do. I'd love to make it so that we could not squash & merge on prs to production, but I don't think github gives us that level of control. We can only enable/disable buttons for all branches.
2. We renamed a lot of folders in recent changes merged to `staging`.
3. In one of my remove library context branches I force pushed which led to a bunch of weird conflicts. Maybe this is having an effect on my merge to prod? I can't imagine it would, but I'm going to avoid force pushing in the future regardless.

We did change environment variables in this PR. I believe Crystal made all the necessary changes to the Vercel deployment as well as the AWS deployment. Once this is merged we can delete the `CONFIG_FILE` env var.

As part of this PR, the following Jira issues should be tested to confirm they are actually fixed:
- [OE-437](https://jira.nypl.org/browse/OE-437)
- [OE-360](https://jira.nypl.org/browse/OE-360)
- [OE-340](https://jira.nypl.org/browse/OE-340)

QA checklist:
https://docs.google.com/document/d/1Sm_mFNAlbUrjT4bfJt5boQMeIA7LoPwoPa4uXqp8kFg/edit


kristojorg:
> It looks like the last time we pushed to prod, we "Squashed and merged" when we should have just added all the commits and merged. This is hard to remember to do. I'd love to make it so that we could not squash & merge on prs to production, but I don't think github gives us that level of control. We can only enable/disable buttons for all branches.

This has been a source of frustration for me too. We should add this to the merge the production PR template @Toxiapo that we need to "create merge commit"

kristojorg:
> In one of my remove library context branches I force pushed which led to a bunch of weird conflicts. Maybe this is having an effect on my merge to prod? I can't imagine it would, but I'm going to avoid force pushing in the future regardless.

I don't think this should have any impact. Force pushes are sometimes useful, just not when anyone has branched off of the branch you are about to force push to.

Toxiapo:
> > It looks like the last time we pushed to prod, we "Squashed and merged" when we should have just added all the commits and merged. This is hard to remember to do. I'd love to make it so that we could not squash & merge on prs to production, but I don't think github gives us that level of control. We can only enable/disable buttons for all branches.
> 
> This has been a source of frustration for me too. We should add this to the merge the production PR template @Toxiapo that we need to "create merge commit"

Yeah, good idea, I will make a note of that.

kristojorg:
Closing this since we need to merge from staging to production in order for staging to be kept up to date as well

PR #419, opened 2022-05-04
OE-291: Enable offline reading

samanthaandrews:
- Adds prefetch links for book content, readium CSS files, and dyslexia font resources <-- this is what enables offline reading
- Moves Readium CSS files into the `public` directory to ensure the Cache-Control headers can be set to cache for 7 days
- Removes `next-pwa`, `file-loader`, and `extract-loader` since they're no longer needed
- Adds a Cypress integration test for confirming prefetching is working as expected
- Refactors `checkoutAllAccessBook` cypress command (renamed to `checkoutBook`) to chain off of the signin commands so that books are checked out only after first being authenticated.


samanthaandrews:
Sigh. Moving this back to draft to investigate these failing Cypress tests.

next.config.js
```diff
@@ -79,28 +78,6 @@ const withBundleAnalyzer = require("@next/bundle-analyzer")({
   enabled: process.env.ANALYZE === "true"
 });
 
-/**
- * Disabled next-pwa config.
- * @todo - re-enable this once we upgrade packages to webpack 5.
- * https://jira.nypl.org/browse/OE-291
```
Toxiapo, line 85: Was the plan to use next-pwa for the offline mode functionality only? Just making sure if we need this in the future.

package.json
```diff
@@ -21,7 +21,7 @@
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
     "cypress:integration:open": "NEXT_PUBLIC_ENABLE_MSW=true cypress open",
-    "cypress:integration:run": "NEXT_PUBLIC_ENABLE_MSW=true cypress run",
+    "cypress:integration:run": "NEXT_PUBLIC_ENABLE_MSW=true cypress run --spec 'cypress/tests/open-ebooks/integration/**/*'",
```
Toxiapo, line 24: This is great, can we also add the pathing to the `cypress:integration:open` command?

Toxiapo (commented):
This looks great! Just a small question then I think this is good to go.

next.config.js
kristojorg: Yea we are removing this for now as it's not currently needed for the limited offline functionality we currently need (only if you've already opened a book, keeping next chapters around)

kristojorg (approved):
Looks good to me! @ayan1229 ready for your QA : )

package.json
samanthaandrews: Looks like `cypress open` doesn't accept a --spec argument https://docs.cypress.io/guides/guides/command-line#Options-1

src/components/WebReader.tsx
```diff
@@ -175,14 +172,27 @@ const ReaderComponent: React.FC = () => {
     <>
       {/* Prefetch all of the readium CSS, resources, and reading order for offline reading */}
       <Head>
-        {/* {injectables.map(injectable => (
+        {injectables.map(injectable => (
```
kristojorg, line 165: I thought we were moving away from importing the injectable css files this way towards simply copying them into the `/public` directory for now?

samanthaandrews: They are copied in the /public directory. Above in the file you can see where I removed them from being imported and instead added the paths to the injectables array. That array is still named `injectables` though. Should this be changed?

PR #420, opened 2022-05-04
OE-433: Remove "9999 of 9999 copies available" string from borrowable books

samanthaandrews:
https://jira.nypl.org/browse/OE-433

This availability text now only displays for books that have already been borrowed and shows how long until the loan expires.


kristojorg (approved):
LGTM!

PR #421, opened 2022-05-05
OE-411: Refactor integration tests with fixtures

Toxiapo:
This PR updates our cypress integration tests to start our APP in the plugin file and use fixtures for all requests response that hits the CM backend.


cypress/support/commands.ts
```diff
@@ -40,6 +41,7 @@ declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace Cypress {
     interface Chainable {
+      signin(userType: UserType, appType?: AppType): void;
```
Toxiapo, line 44: I created this new command to sign in using the GUI and used it throughout the integration tests because MSW doesn't intercept the cy.requests  used in `signinByApi`.

cypress/support/utils.ts
```diff
@@ -110,50 +111,7 @@ export const V2_IFRAME_SELECTOR = "#html-reader-iframe";
 
 // BOOKS USED FOR ERRORS, we use these to test the error handling for MSW intercepts eventhough they are totally valid books
 export const JFK_FETCH_URL = `${SERVER_URL}/works/Axis%20360%20ID/0016692239/borrow/15`;
-
 export const JFK_BOOK_ENDPOINT = `${SERVER_URL}/works/Axis%20360%20ID/0016692239`;
-export const GN_BOOK_ENDPOINT = `${SERVER_URL}/works/Axis%20360%20ID/0020037525`;
```
Toxiapo, line 115: We can finally remove this blob of code now that intercept works : )

cypress/tests/open-ebooks/integration/browse/breadcrumb-navigation.ts
```diff
@@ -5,13 +5,17 @@ import {
...
+    cy.signin("ALL_ACCESS_USER");
   });
 
   beforeEach(() => {
     Cypress.Cookies.preserveOnce(AUTH_COOKIE_NAME);
   });
 
+  after(() => {
+    cy.clearCookie(AUTH_COOKIE_NAME);
+  });
```
Toxiapo, line 17: This is really annoying because sometimes cypress doesn't clear the cookie even though by default it should. So I am just explicitly removing these cookies for now.

msw/handlers/index.ts
```diff
@@ -153,9 +161,9 @@ export const handlers = [
 
   // we can't mock this response because it includes a unique hash
   // and other date/time-specific licencins data
-  rest.get("https://node.axisnow.com/license/*", () => {
-    return;
-  }),
+  // rest.get("https://node.axisnow.com/license/*", () => {
+  //   return;
+  // }),
```
samanthaandrews, line 166: Can we remove this?

samanthaandrews, line 166: Can we remove this?

src/components/Error.tsx
```diff
@@ -20,12 +19,8 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
   const router = useRouter();
 
   return (
-    //It isn't necessary to show an error page for 401 (Unauthorized) errors since the user will be redirected to the signin page
-    //Instead, we display a PageLoader to avoid the undesirable display of an error screen while the user waits for that redirect to occur
```
samanthaandrews, line 24: Is this no longer the case?

cypress/plugins/index.ts
```diff
@@ -48,7 +53,58 @@ module.exports = (
 
   config.env.NEXT_PUBLIC_ENABLE_MSW = process.env.NEXT_PUBLIC_ENABLE_MSW;
```
samanthaandrews, line 55: We can remove all references to this env var, right? And the unused `cypress:ci:build` and `cypress:ci:start` commands within the `package.json` too. There are several references to `NEXT_PUBLIC_ENABLE_MSW` in the code and I just want to confirm it either needs to be there or it can be removed.

Also we should probably update the README to include info about how to run Cypress tests. When I pulled down your PR I defaulted to building and starting the app, and then realized that I Cypress was now controlling the `start` process so I didn't need to do that.

samanthaandrews (changes requested):
This looks so good! Just a few questions & some small cleanup stuff. Let's sync up tomorrow after standup and talk through the `signin` command bc I just want to make sure I understand it and consider what the `e2e` tests will need to change (if anything).

cypress/plugins/index.ts
samanthaandrews: Maybe we should also remove the reference to `NEXT_PUBLIC_ENABLE_MSW` in the `.env` file since it no longer needs to be set at that level and we aren't running the risk of that module getting bundled in production. I'm also wondering if we can delete the `mockServiceWorker.js` file or if it's still necessary.

samanthaandrews: Wait I am wrong about `mockServiceWorker.js`. We still need that.

samanthaandrews: But we can remove `NEXT_PUBLIC_ENABLE_MSW` in `.env`. I ran `ANALYZE=true npm run build` on your branch and confirmed that msw is being tree-shaken out of production builds, so that's good.

next.config.js
```diff
@@ -13,7 +13,6 @@ const {
   NEXT_PUBLIC_GTM_ID,
   NEXT_PUBLIC_BUGSNAG_API_KEY,
   NEXT_PUBLIC_CM_URL,
-  NEXT_PUBLIC_ENABLE_MSW,
```
samanthaandrews, line 16: Realizing we actually might need to still pass this through in order for Cypress to pick up this env var

Toxiapo: ahhh, maybe that's why the tests are failing! I will give this a try.

cypress/plugins/index.ts
Toxiapo: Done

next.config.js
Toxiapo: nvm, it was because I named the env variable wrong!

cypress/plugins/index.ts
```diff
@@ -48,7 +53,62 @@ module.exports = (
...
+    }
+  });
+}
+
+// Start the msw server from the nodejs environment
+function startMswSever() {
+  const { setupServer } = require("msw/node");
+  const requestInterceptor: SetupServerApi = setupServer(...handlers);
+  requestInterceptor.listen({
+    // onUnhandledRequest: "bypass"
```
samanthaandrews, line 110: Do we need this to bypass or should we remove it?

samanthaandrews (approved):
One small question about the 'bypass' that was commented out. I think this is good to go.

cypress/plugins/index.ts
Toxiapo: Oh yes, we can just turn it on now.

PR #422, opened 2022-05-10
OE-495: Add missing redirects

samanthaandrews:
https://jira.nypl.org/browse/OE-495


Toxiapo (approved):
I've checked the list of redirects from Jira and they all seem to be good!

PR #423, opened 2022-05-11
[OE-482] Firstbook signin failure fix

ayan1229:
https://jira.nypl.org/browse/OE-482 

changes address the first book sign in failure issue.


cypress/tests/open-ebooks/server-e2e/signin/firstbook-signin.ts
```diff
@@ -70,6 +72,7 @@ describe("FirstBook signin", () => {
     cy.findByRole("button", { name: "Confirm Sign Out" })
       .should("exist")
       .click();
+    cy.clearCookie(AUTH_COOKIE_NAME);
```
kristojorg, line 75: The idea is to make sure that the app does this itself, so I don't think we want to do it in the test here

samanthaandrews:
I'm going to close this issue in favor of adding Ayan's playwright tests to the repo in this upcoming sprint.

PR #424, opened 2022-05-11
Update epub-to-webpub and webreader

Toxiapo:
This PR updates the epub-to-webpub and the webreader which fixes the following two bugs

- [OE-422](https://jira.nypl.org/browse/OE-422) - Epub conversion failure
- [OE-236](https://jira.nypl.org/browse/OE-236) - Book rendering bug for fixed layouts
- [OE-496](https://jira.nypl.org/browse/OE-496) - Same as OE-422, this issue is a duplicate of that one as far as I can tell


samanthaandrews (approved):
Code looks good & my manual testing confirms that this fixes both of those bugs. Let's kick it over to QA for final review.

PR #426, opened 2022-05-18
OE-485: Install next.js new relic agent on a dockerized environment

Toxiapo:
This PR will install the newrelic agent on docker deployment for both the production and qa deploys.  This does not include bug and browser reporting features, this is to make sure we can set the agent on a docker and be able to see activities when users visit the URL. 

- [x] Setup newrelic in a container and see activities in the dashboard
- [x] Set up new relic GHA secrets
- [x] Temporary deploy to AWS by changing qa-release workflow, see activities in the dashboard


src/components/Collection.tsx
```diff
@@ -25,6 +25,7 @@ export const Collection: React.FC<{
   const { storedBreadcrumbs, setStoredBreadcrumbs } = useBreadcrumbContext();
 
   React.useEffect(() => {
+    console.log("collection page");
```
samanthaandrews, line 28: i'm guessing this is here to make sure we can see it in New Relic?

Toxiapo: Yeah, I used it to debug, I will remove it!

PR #427, opened 2022-05-20
OE-111: Add Codecov

samanthaandrews:
Adds Codecov reports to pull requests - https://jira.nypl.org/browse/OE-111

Here's the project on Codecov: https://app.codecov.io/gh/NYPL-Simplified/circulation-patron-web (you should be able to authenticate using GitHub). You will notice that the dashboard is blank. This is because no coverage information has been uploaded onto the default branch. We can fix this by merging this pull request. https://docs.codecov.com/docs/step-2-getting-a-codecov-account-and-uploading-coverage

I didn't set up any coverage standards or thresholds yet.


samanthaandrews:
I set this up with only Jest test coverage for now. I spent several hours attempting to also configure Cypress coverage and it seems like we should probably make another ticket for it. I created https://jira.nypl.org/browse/OE-524 and assigned it to me.

Toxiapo (approved):

PR #430, opened 2022-06-02
OE-534: Setup chakraui base theme

Toxiapo:
This PR adds the base theme with chakra to the repo.

More details are documented [here](https://docs.google.com/document/d/1Tv0XctyUhwNkFOcFxAxaSOM-hCUGLOdJOBRla9znt_s/edit#) to explain some of the choices I made, hopefully, it will help us when we start our implementations in the future.


src/marketing/theme/foundations/color.ts
```diff
@@ -0,0 +1,65 @@
...
+      100: "#E0F2F3",
+      200: "#B1F8FC",
+      300: "#34CFD8",
+      400: "#0A99A0",
+      500: "#07676C",
+      graident: {
+        primary: tealPrimaryGradient
+      },
+      opacity: {
+        "60%": "#0A99A0"
```
kristojorg, line 43: I suggested on figma that we use the 300 version of each color instead of the opacities, so I don't think we need this necessarily. I think we can just use `teal.300` in these places

src/marketing/theme/foundations/color.ts
```diff
@@ -0,0 +1,65 @@
...
+  "linear-gradient(180deg, #693CEA 0%, #261C5D 100%)";
+
+const colors: Colors = {
+  transparent: "transparent",
+
+  ui: {
+    black: "#000000",
+    semiBlack: "#222222",
+    darkGray: "#8A8A8A",
+    lightGray: "#D9D9D9",
```
kristojorg, line 16: Should this be added to the `gray` scale?

src/marketing/theme/components/button.ts
```diff
@@ -0,0 +1,77 @@
...
+function primaryVariant() {
+  return {
+    color: "ui.bodyText",
+    bgColor: "ui.gray.xxlight",
+    borderRadius: "headerAuthBtn",
+    fontSize: "md",
+    fontweight: "medium",
+    lineHeight: "5",
+    px: "11px",
+    py: "7px",
```
kristojorg, line 24: I think we should avoid using hard coded specific pixel spacing that doesn't fall onto the spacing scale. Let's just choose a value on the scale that appears to fit as best as possible

src/marketing/theme/foundations/shadow.ts
```diff
@@ -0,0 +1,9 @@
+const shadows = {
+  button: "0px 1px 2px rgba(0, 0, 0, 0.25), 0px 2px 1px rgba(0, 0, 0, 0.3)",
+  icon: "0px 1px 2px rgba(0, 0, 0, 0.3), 0px 1px 3px 1px rgba(0, 0, 0, 0.15)",
+  carouselBook:
+    "0px 3px 2px rgba(0, 0, 0, 0.15), 0px 2px 1px rgba(0, 0, 0, 0.3)",
+  gridBook: "0px 3px 2px rgba(0, 0, 0, 0.3), 0px 1px 3px rgba(0, 0, 0, 0.15)"
```
kristojorg, line 6: If these are only used in those places, I don't think we necessarily need to define them in the theme, they can just be inline styles on that component. That said, it's also not a huge problem to define them here if you prefer

src/marketing/theme/foundations/raddi.ts
```diff
@@ -0,0 +1,19 @@
...
+  lg: "0.5rem",
+  xl: "0.75rem",
+  "2xl": "1rem",
+  "3xl": "1.25rem",
+  "4xl": "1.5rem",
+  full: "9999px"
+};
+
+const namedRadii = {
+  button: "50px",
```
kristojorg, line 15: Is the button not "full" radii?

src/marketing/theme/foundations/global.ts
```diff
@@ -0,0 +1,48 @@
...
+    fontSize: ["xl", "2xl", "9xl"],
+    fontWeight: "black",
+    lineHeight: ["shorter"],
+    color: "ui.bodyText"
+  },
+  h2: {
+    fontSize: ["xl", "4xl", "7xl"],
+    fontweight: "black",
+    lineHeight: ["shorter"],
+    color: "ui.bodyText"
```
kristojorg, line 19: I think the `color` we should apply in a global style to the whole page to all text, if all text is supposed to be that color. Can also be defined here too, but just a note that we should do that

src/marketing/theme/components/link.ts
```diff
@@ -0,0 +1,45 @@
...
+    display: "flex",
+    alignItems: "center",
+    textAlign: "center",
+    p: [4, 1, 1],
+    _hover: {
+      bgColor: ["#EEEEEE", "ui.teal.500", "ui.teal.500"],
+      textDecoration: "none"
+    }
+  };
+}
```
kristojorg, line 33: So it looks like you're doing both the mobile menu and the desktop nav links here, but I think we will render them as separate components so I don't think we need to make the styles adapt like this, they can be separate. 

I am also wondering if this should be a `Button` style, and then to render a link version we use `<Button as={Link}>`. That way we can use these same styles for buttons and buttons that look like links (unless we don't need that, you tell me)

src/marketing/theme/components/button.ts
```diff
@@ -0,0 +1,77 @@
...
+  const colorScheme = props.colorScheme;
+
+  switch (colorScheme) {
+    case "magenta":
+      return {
+        color: "ui.white",
+        bgColor: "ui.magenta.400",
+        borderRadius: "button",
+        boxShadow: "button",
+        fontSize: "32px",
```
kristojorg, line 42: Same with the font size, I think we should avoid hard coding a specific pixel value but use a value from the scale instead. I tried to write out all the font sizes and weights we need in the TAD, and I think this one would have been in there

src/marketing/theme/components/button.ts
```diff
@@ -0,0 +1,77 @@
...
+        lineHeight: "shorter",
+        fontWeight: "bold",
+        w: ["225px", "414px", "426px"],
+        h: ["45px", "85px", "95px"],
+        px: "155px",
+        py: "17px",
```
kristojorg, line 48: I don't think an explicit width or height are necessary, the button should size correctly if we simply put the correct padding and fontSize, and then add some text inside. This way it will grow for longer words, shrink for shorter. The padding should also come from the theme scale. And this goes for everywhere in the designs. We shouldn't just use the specific padding in the designs, but choose whatever padding is closest from our theme scale

src/marketing/theme/components/button.ts
```diff
@@ -0,0 +1,77 @@
...
+        _hover: {
+          bgColor: "ui.magenta.500",
+          boxShadow: "none"
+        }
+      };
+    case "teal":
```
kristojorg, line 54: I don't see where the teal and magenta buttons are used, can you point me to them?

kristojorg (changes requested):
This looks good! The one thing I'm unsure about is if we should put more thought into our text styles. It looks like there is a standard size of headings on the non-home marketing pages (like FAQ etc). Should we try to come up with a set of text styles to fit those? Is the font size scale we have here going to fit the font sizes used in the designs?

A few other small things. But otherwise I think this gets us ready to roll out some new designs!

src/marketing/theme/foundations/raddi.ts
Toxiapo: oh, that's right!

src/marketing/theme/components/button.ts
Toxiapo: Oh yes, I do have this defined, will update it now!

src/marketing/asset/top10-teal.svg
```diff
@@ -0,0 +1,21 @@
+<svg width="179" height="179" viewBox="0 0 179 179" fill="none" xmlns="http://www.w3.org/2000/svg">
```
samanthaandrews, line 1: I noticed the kerning between the 1 and the 0 doesn't match the other two Top 10 badges. Going to reach out to Jenn in the Figma file to get this fixed.

src/marketing/theme/components/button.ts
Toxiapo: I made an "auth" variant for the button and they are being used for the "How it works" section and the contact page form, but now that we don't have the contact page, I am wondering if I even need this variant.

I think I will keep the style for the magenta, but without the colorscheme, and remove the teal for now.

Toxiapo:
> This looks good! The one thing I'm unsure about is if we should put more thought into our text styles. It looks like there is a standard size of headings on the non-home marketing pages (like FAQ etc). Should we try to come up with a set of text styles to fit those? Is the font size scale we have here going to fit the font sizes used in the designs?

Yeah, agree. I thought defining the h1-h5 styles in global would work here, but those will get overridden if we use chakra's Heading component. i.e.`<Heading as="h1">`

I would name the heading styles `h1, h2, h3, etc`, and the subtitles with a more descriptive name like `bookTitle` in the textStyles theme,  what do you think? @samanthaandrews @kristojorg

src/marketing/theme/foundations/color.ts
kristojorg: Nice yea I think let's do that to start. It's also fine to have just `gray.light` and `gray.xxlight` to start if you want

src/marketing/theme/components/button.ts
kristojorg: Okay yea I think we can just name this the primary variant and have these different colors. That seems fine to me

kristojorg:
> I would name the heading styles h1, h2, h3, etc, and the subtitles with a more descriptive name like bookTitle in the textStyles theme, what do you think?

That seems great to me!

src/marketing/theme/components/link.ts
Toxiapo: Per our discussion, we don't necessary need to create this in the base theme, but rather just create the react comment for them. e.g. <HeaderLink>, <MobileMenuLink>

src/marketing/theme/foundations/shadow.ts
Toxiapo: Let's keep them now, but we can delete them if it's just for one place. I think we will need a few instances for the carouselBook.

src/marketing/theme/foundations/color.ts
```diff
@@ -0,0 +1,56 @@
...
+    bodyText: "#424242",
+
+    magenta: {
+      100: "#FFF2F8",
+      200: "#FFDFEE",
+      300: "#FF6AAF",
+      400: "#F41666",
+      500: "#BF0C4D",
+      gradient: {
+        primary: magentaPrimaryGradient
```
kristojorg, line 25: Out of curiosity, why are these ones primary and the purple one is secondary?

src/marketing/theme/types.ts
```diff
@@ -0,0 +1 @@
+export declare type Dict<T = any> = Record<string, T>;
```
kristojorg, line 1: I think we already have a Dict type declared somewhere, but could be wrong

kristojorg (approved):
This looks great!

src/marketing/theme/types.ts
Toxiapo: Not in CPW, but we could import it from the webreader.

src/marketing/theme/foundations/color.ts
Toxiapo: Oh! I kept the name that was defined in the Figma file.

src/marketing/theme/types.ts
kristojorg: Oh then I think I would put this under `interfaces.ts`. The only thing we need to keep separate in the `marketing` folder are the ChakraUI components themselves that can't be mixed with the ThemeUI ones.

src/marketing/components/Header.tsx
```diff
@@ -308,8 +307,17 @@ const AuthButton: React.FC<AuthButtonProps> = ({
     router.push(SIGNIN_PICKER_PATH);
   };
 
+  const signOutHandler = async () => {
+    localStorage.clear();
+
+    const res = await fetch(SIGNOUT_PATH, { method: "POST" });
+    if (res.ok) {
+      window.location.href = new URL("/", window.location.origin).toString();
```
Toxiapo, line 315: Unfortunately, I can't re-use the signout function from `auth.ts` because I need to refresh the page after the user pressed the signout button on the homepage.

kristojorg: Huh, interesting. Why do you need to refresh the page?

Toxiapo: The signout API does a redirect to the homepage after signing out, but we are on the homepage so it's not performing any page updates. I needed a trigger to toggle the auth button to show Sign in/ Sign out.

src/marketing/mdx/MDXComponents.tsx
```diff
@@ -0,0 +1,94 @@
...
+const H2 = ({ children }: MDXComponentProps) => {
+  const { id } = createTOCAnchors(children);
+  return (
+    <Text
+      textStyle="h4"
+      as="h2"
+      marginTop="8"
+      marginBottom="2"
+      id={id}
+      // accommodates the height of the sticky nav when jumping using FAQ table of contents
```
samanthaandrews, line 39: @kristojorg This `_before` object is to accommodate the height of the sticky nav on the FAQ page. I added a comment for clarity.

src/marketing/components/Layout.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { ErrorBoundary } from "components/ErrorBoundary";
+import Footer from "marketing/components/Footer";
+import Header from "marketing/components/Header";
+import theme from "../theme";
+import SiteHead from "components/Head";
+
+const MarketingPageLayout: React.FC = ({ children }) => {
+  return (
+    <ErrorBoundary>
+      <SiteHead />
```
samanthaandrews, line 12: @kristojorg I removed the `<MarketingPage />` component i had created previously and just put the SiteHead here. You're right I don't think it makes sense for it to be a separate component.

PR #434, opened 2022-06-15
Oe 503/new carousel

kristojorg:


src/marketing/theme/foundations/breakpoints.ts
```diff
@@ -1,7 +1,7 @@
-const breakpoints = {
-  sm: "20em", // 320px
-  md: "38em", // 600px
-  lg: "80em" // 1280px
-};
+const breakpoints = ["0em", "20em", "38em", "80em"];
+(breakpoints as any).sm = breakpoints[0];
+(breakpoints as any).md = breakpoints[1];
+(breakpoints as any).lg = breakpoints[2];
+// const breakpoints = { sm: "20em", md: "38em", lg: "80em" };
```
Toxiapo, line 5: is there a reason to switch the breakpoints to these? This will break the viewport logic that uses the old breakpoint values.

Toxiapo:
> On large screens, somehow the Top 10 badge is being cut off. I cannot for the life of me figure out what is cutting it off?

@kristojorg  IDK why but seems like changing the order of these renders fixes the problem, somehow the 148px is covering up what's on the view. (Render the largest size at the top).

Screen Shot 2022-06-17 at 9 48 34 AM

src/marketing/Carousel/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+        right={isPrev ? "auto" : 0}
+        zIndex="overlay"
+        h={["3.5rem", "5rem"]}
+        w={["1.75rem", "2.5rem"]}
+        maxW={["29px", "45px"]}
+        borderLeftRadius="full"
+        className={className}
+        fontSize={["7xl", "2xxl"]}
+        aria-label={isPrev ? "Previous" : "Next"}
+        transform={isPrev ? "rotate(180deg)" : "rotate(0deg)"}
```
samanthaandrews, line 175: Ah ok I see, we're rotating the ChevronRight component. Makes sense to me. Maybe we should add a comment to the ChevronRight component.

samanthaandrews (commented):
This looks so good! Just a few questions and comments.

src/marketing/theme/foundations/breakpoints.ts
kristojorg: I _think_ this should preserve the same breakpoints as we had before (small=20em, medium=38em, large=80em). I was getting an error whenever I tried to use the `sx` prop (which I needed to use for certain css targeting). It was saying it `breakpoints.map is not a function`, which indicated that it expected the breakpoints to be an array. When I read the Chakra responsive styles docs, it looked like the object internally was converted to an array anyway, so I went ahead and provided an array, but also put the `sm` `md` and `lg` properties on the array to preserve the object-like functionality. I also added the `0em` breakpoint at the start because it looked like that's what Chakra did, and it seemed to make it work as I expected.

In the end, the error I was seeing was somehow coming from some interplay between Chakra and ThemeUI. The error originated in theme-ui source. I thought maybe it was because we were using the [JSX Pragma](https://theme-ui.com/guides/jsx-pragma/) of theme-ui, which was running some theme-ui code on all pages (specifically to transform the `sx` prop). However, disabling that in `tsconfig` didn't seem to fix it either.

So basically, this was the only way I could make the code run. I'm hoping that this setup avoids the error while still working the same way as the object based format we had before. Maybe try switching it in your header branch @Toxiapo and let me know if it breaks things or not?

kristojorg:
Thank you @Toxiapo for finding that fix! Works perfect : ) !

src/marketing/icon/ChevronRight.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+import { chakra } from "@chakra-ui/react";
+
+const SvgComponent = (props: React.ComponentProps<typeof chakra.svg>) => (
```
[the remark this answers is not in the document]
kristojorg: No, we I just rotate it 180deg for the left one

src/marketing/Carousel/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+        <GridItem gridColumn="1 / 1" gridRow={["1/1", "1 / 1", "1 / 3"]}>
+          <Top10Badge colorScheme={colorScheme} />
+        </GridItem>
+        <GridItem>
+          <Text textStyle="h1" as="h1" mt={[3, 4, 6]}>
+            {title}
+          </Text>
+        </GridItem>
+        <GridItem gridColumn={["1/3", "1 / 3", "2 / 3"]} gridRow="2 / 3">
+          <Text fontSize={["md", "xl"]} maxW="545px">
```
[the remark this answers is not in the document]
kristojorg: The chakra `spacing` scale is used for padding and margin, but not for width and height. You can use the [sizes](https://chakra-ui.com/docs/styled-system/style-props#layout-width-and-height) theme key for that, but I feel like our designs don't have consistent enough w/h to make that useful (I wouldn't really know what to put in the theme there). 

I could change this to an em value though of 34. I'll do that.

src/marketing/Carousel/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+  );
+});
+
+function BookCover({ book }: { book: SimpleBook; numberOfBooks: number }) {
+  return (
+    <AspectRatio ratio={0.7} width="100vw" maxW={"100%"}>
+      <Box
+        shadow="carouselBook"
+        position="relative"
+        bgColor="rgba(160, 174, 192)"
```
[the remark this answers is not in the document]
kristojorg: Oh actually I should have used a theme gray here. I was trying to see if I could use `rgba` and make it kind of see through on the background but that didn't work, so I just reverted back to opaque gray which I should pull from an existing gray in theme.

src/marketing/Carousel/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+        animation="0.8s linear 0s infinite alternate none running animation-snm94m"
```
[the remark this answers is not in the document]
kristojorg: So I was trying to make the skeleton animation from Chakra apply on this (and that's the name of it), but it actually wasn't working. I just placed the Skeleton component inside instead.

src/marketing/theme/foundations/breakpoints.ts
Toxiapo: The responsiveness reacted differently with these values the last time I tried, maybe it's because the "0em" was defined explicitly? I will give it another test in my PRs.

kristojorg: Ok just updated the breakpoints back to original after the ThemeUI fix

samanthaandrews (approved):
🚀

PR #436, opened 2022-06-17
OE-553: homepage redesign

Toxiapo:
This PR creates the rest of the miscellaneous layouts of the new homepage design. 

Preview: [/homepage](https://open-ebooks-web-git-oe-553-homepage-redesign-nypl.vercel.app/homepage)


src/pages/homepage/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+            fontSize={{ base: "md", sm: "xl" }}
+            fontWeight={{ base: "normal" }}
+            lineHeight={{ base: "shorter", md: "base" }}
+            my={{ base: 4, lg: 1 }}
+            ml={{ base: 7, sm: 8, lg: 0 }}
+            mr={{ base: 1 }}
+            pb={{ lg: 6 }}
+          >
+            Open eBooks has elementary, middle, and high school
+            <br />
```
samanthaandrews, line 279: Having the `<br />` here matches the mockup but semantically I don't think it makes sense. Can we instead give this text a max width?

src/pages/homepage/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+      </Grid>
+    </Box>
+  );
+};
+
+const ReadAnywhere: React.FC = () => {
+  return (
+    <Box bg="linear-gradient(180deg, #FFFFFF 41.6%, #EEEEEE 100%)" mb={1}>
+      <Flex
+        // maxW="1280px"
```
samanthaandrews, line 292: Stray unused code

src/pages/homepage/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+        templateAreas={{
+          base: `
+          "image header"
+          "text text"
+          `,
+          lg: `
+          "image header"
+          "image text"
+          `
+        }}
```
samanthaandrews, line 227: nice!

samanthaandrews (changes requested):
super small things i noticed. re-request review once you make updates and I'll approve!

src/pages/homepage/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+const steps: Step[] = [
+  {
+    idx: 1,
+    icon: HandIcon,
+    description: "Ask a teacher or school librarian for access"
+  },
+  {
+    idx: 2,
+    icon: PasswordIcon,
+    description: "Sign In"
```
samanthaandrews, line 88: In this context, since it's not a button or heading I think this should be "Sign in"

src/pages/homepage/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+      </ChakraProvider>
+    </Page>
+  );
+}
+
+const Home: React.FC = () => {
+  return (
+    <Box as="main">
+      {/* Hero Image */}
+      <Image
```
samanthaandrews, line 47: I noticed @kristojorg using `next/image` in the grid and carousel components. Should we be using that here? When should we be using one over the other?

kristojorg:
The mobile phone images look a little blurry to me. Maybe we need to export them in higher quality?

src/marketing/theme/foundations/spacing.ts
```diff
@@ -1,4 +1,5 @@
 const defaultChakraSpacing = {
+  "-1": 0,
```
kristojorg, line 2: Shouldn't 0 be 0?

src/pages/homepage/index.tsx
kristojorg: I think we should use next/image always from now on basically

src/marketing/theme/components/button.ts
```diff
@@ -29,15 +29,16 @@ function primaryColors(colorScheme: string) {
...
         color: "ui.white",
         bgColor: "ui.magenta.400",
         borderRadius: "full",
         boxShadow: "button",
-        fontSize: "4xl",
         lineHeight: "shorter",
-        fontWeight: "bold",
-        px: "28",
-        py: "7",
+        fontWeight: { base: "black", md: "bold" },
```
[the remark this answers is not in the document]
Toxiapo: Okay! Thank you

src/marketing/icon/hand.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { StyleProps } from "@chakra-ui/react";
+
+const HandIcon = (
+  props: React.SVGProps<SVGSVGElement> & StyleProps
```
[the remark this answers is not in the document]
kristojorg: If you are using what in the component? I guess let me back up and ask what you want to accomplish by adding the `StyleProps` type to this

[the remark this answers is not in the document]
Toxiapo: The message got cut off for some reasons...

`Would this be okay tho if I am using this in the <Icon as={HandIcon} /> component?`

So the goal was to make it so that it accepts all chakra style props.

src/pages/homepage/index.tsx
Toxiapo: AFAIK the `next/image` forces you to either specify the width and the height of the image or use "layout=fill", which essentially still need the width/height in the parent container.

I could just hard code the width and the height of the images, but do you guys know a way to make it so that you don't need to do that and make it render what ever the size of the image?

samanthaandrews: I think you can use `layout="fixed"` which should behave like the native `` element.

Toxiapo: This is what I get: `Error: Image with src "img/apple-screenshot.png" must use "width" and "height" properties or "layout='fill'" property.`

Toxiapo: It's strange to me that we there's no easy way to make this responsive with just the Image component. I will just use the AspectRatio component for now and set the size based on their maximum size like how I did it for the hero image. Thanks, Sam.

samanthaandrews (approved):

src/marketing/icon/hand.tsx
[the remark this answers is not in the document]
kristojorg: You could pass it to `Icon` without the `StyleProps` type set on the props. You could then style the `Icon` normally with style props. If you want to use it as a chakra component itself (`<HandIcon maxWidth={2} />`), then you have to use the chakra factory setup to do that, and you don't need to pass the component the `StyleProps` type

PR #438, opened 2022-06-22
Refactor JSX Pragma to remove ThemeUI Global

kristojorg:
This PR solves the issue of the ThemeUI-specific JSX transform being applied globally to all pages. It does so by:
1. Disabling the global transform previously found set up in .babelrc and tsconfig. This makes most styles fail on non-chakra pages.
2. Adding the JSX pragma comment to all tsx files in src folder _not_ in the /marketing folder or __tests__ folders. Should be just the app code. This makes the styles on ThemeUI pages resolve correctly again, but it now only applies the custom transform to ThemeUI pages

Once we merge this, I can update my #434 PR and the regular breakpoints setup should work again

**Note:**
When I removed our babel config, it made Next.js switch to using SWC for compiling the app, which is going to be faster. Removing the babel config also made Jest stop working correctly, so I explicitly told Jest to use swc/jest to compile the app as well (so the app and jest both use same compiler). However, the NYPL design system does not seem to work with this. Maybe it would if we updated it, but I think it's better to just remove it completely instead. I saw @samanthaandrews had already removed some Icons, so I'm not going to touch it in this PR. Once we merge this into the theme branch we should make sure it's completely removed and the tests will start working again


kristojorg:
Also the typecheck is failing due to an issue in the marketing/header component which I'm not going to touch so I don't mess with merge conflicts

Toxiapo (approved):
This looks good! 

We are only using Icons from the NYPL design system so I think removing it should be fine for now. 

What was the error you saw regarding the NYPL lib? We might want to switch to using the NYPL design system in the future so it would be nice to document the errors early.

samanthaandrews:
Yeah I removed the whole design system package in my static pages PR. Also those type errors have been fixed in my PR.

PR #440, opened 2022-06-23
Static page cleanup

samanthaandrews:
- Addresses a few content updates from Miguel
- Merges the new Header & Footer into the MDX pages


src/marketing/mdx/MDXComponents.tsx
```diff
@@ -37,7 +30,21 @@ const H1 = ({ children }: MDXComponentProps) => (
 const H2 = ({ children }: MDXComponentProps) => {
   const { id } = createTOCAnchors(children);
   return (
-    <Text textStyle="h4" as="h2" marginTop="8" marginBottom="2" id={id}>
+    <Text
+      textStyle="h4"
```
Toxiapo, line 34: the h4 style is much bigger than h2 here, this is due to the fact that I didn't specify the style for the mobile screens.

Toxiapo (changes requested):
Just that one style update otherwise this is good to go!

src/marketing/mdx/MDXComponents.tsx
samanthaandrews: you're right. fixed!

Toxiapo (approved):

src/marketing/components/MarketingPage.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import SiteHead from "components/Head";
+import { SWRConfig } from "swr";
+import swrConfig from "utils/swrConfig";
+import { UserProvider } from "components/context/UserContext";
+
+/* Wrapper for marketing pages that doesn't include references to theme-ui */
+
+const MarketingPage: React.FC = ({ children }) => {
```
kristojorg, line 9: Why have this as separate from `MarketingPageLayout`?

src/marketing/components/MarketingPage.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <>
+      <SiteHead />
+      <SWRConfig value={swrConfig}>
+        <UserProvider>{children}</UserProvider>
```
kristojorg, line 14: I'm actually wondering if we should use the `<UserProvider>` here or the `<CredentialsProvider>`. We are only trying to determine the auth state of the user. The user provider has to fetch loans (takes forever) to do that, whereas we could also just assume that if the credentials are defined and set then the user is logged in... if they turn out to be wrong or something then the user will get redirected to login anyway once they try to go to the /app page

src/marketing/icon/BackToTop.tsx
```diff
@@ -1,10 +1,8 @@
-import { SxProp } from "theme-ui";
 import React from "react";
+import { Icon } from "@chakra-ui/react";
 
-const BackToTopIcon = (props: React.SVGProps<SVGSVGElement> & SxProp) => (
-  <svg
-    width="16"
-    height="18"
+const BackToTop = (props: React.ComponentProps<typeof Icon>) => (
+  <Icon
```
kristojorg, line 5: Oh this is good to know! I hadn't realized we could just use the `Icon` component this way

src/marketing/mdx/MDXComponents.tsx
```diff
@@ -37,7 +30,21 @@ const H1 = ({ children }: MDXComponentProps) => (
...
+      marginBottom="2"
+      id={id}
+      _before={{
+        display: "block",
+        content: '" "',
+        marginTop: "-3em",
+        height: "3em",
+        visibility: "hidden",
+        pointerEvents: "none"
+      }}
```
kristojorg, line 46: What is this for?

kristojorg (approved):
A couple questions, but I'm going to merge this and we can discuss on the base theme PR

PR #441, opened 2022-06-25
Adjust mdx layout based on feedback from Jen

samanthaandrews:
Jen showed me that the layout for these pages should have equal margin on each side and the text content should have a max width of 544px. I used Chakra's `<Container />` to achieve this and then gave the other MDX elements (h1, h2, p, etc.) a max width of 34em.


src/marketing/mdx/HowItWorks.tsx
```diff
@@ -7,7 +7,12 @@ interface HowItWorksItemProps extends MDXComponentProps {
 
 export const HowItWorks = ({ children }: MDXComponentProps) => {
   return (
-    <Box borderColor="ui.teal.100" borderRadius={50} borderWidth={6}>
+    <Box
+      maxWidth="34em"
```
Toxiapo, line 11: Should we use `rem` here for consistency?

src/marketing/mdx/MDXComponents.tsx
```diff
@@ -1,36 +1,46 @@
...
   <MarketingPageLayout>
-    <Box
-      marginLeft={[2, 8, 16]}
-      marginRight={[2, 8]}
-      marginTop={[4, 8, 12]}
-      marginBottom="24"
-      maxWidth="lg"
-      as="main"
-    >
+    <Container as="main" maxWidth="container.md" marginBottom="8em">
```
Toxiapo, line 28: Also here.

Toxiapo (approved):
The pages look great! Just a small comment regarding the em and rem useage.

src/marketing/mdx/MDXComponents.tsx
samanthaandrews: This is the parent container, which has a larger width than the text below. It mostly gets applied to the video on the /about page and the /faq page table of contents component. The body text is constrained to 544px for UX legibility reasons.

src/marketing/mdx/HowItWorks.tsx
samanthaandrews: Yup. Updated!

PR #442, opened 2022-06-27
OE-567: UX Review

kristojorg:

PR #443, opened 2022-06-27
Header Feedback: Add header link indicators

Toxiapo:
I missed the header link indicator styles where it shows the user where they are on the page. This PR is to add the missing styles.

The final style is subject to change as Jen is currently creating these designs based on our discussion/feedback, for now, I am creating the styles from the current design file.


Toxiapo:
Looks like the updates will get addressed in this [PR](https://github.com/NYPL-Simplified/circulation-patron-web/pull/442), so I am going to close this now.

PR #444, opened 2022-06-27
OE-486: Install web browser agent

Toxiapo:
This PR installs the web browser agent on both the staging and the production site. We will be using the [Copy & Paste](https://docs.newrelic.com/docs/browser/browser-monitoring/installation/install-browser-monitoring-agent/#copy-paste-app) method as it gives more controls on how we want to track errors in the frontend.

This update includes uploading the sourcemap to the Browser UI. I am using the [@edx/new-relic-source-map-webpack-plugin](https://www.npmjs.com/package/@edx/new-relic-source-map-webpack-plugin) plugin which is essentially a wrapper for the publish-sourcemap API, however, this creates two challenges.

1. According to this [post](https://newrelic.zendesk.com/hc/en-us/articles/4409719142039-Source-map-upload-NextJS), the upload API returns an error indicating the use of dynamic routes with brackets, for nextjs projects`[`,`]`. The workaround I proposed was to encode the brackets in the URI by passing in a custom callback function `staticAssetUrlBuilder` during the webpack build.
2. Publishing the sourcemap does not remove the old ones that are already in their server. I cannot find the option to 'override', instead, I had to delete the existing files from their server before starting the upload, otherwise it would return a 409 error. This is because we are using the "URL" as the unique identifier, and they only allow a single instance of the sourcemap to each URL. We could also give it a unique releaseName and releaseId, but I couldn't get it to show up in the UI for som reasons, and it seems like it's not really efficient as we will have to create multiple releaseName and releaseId through the application consider how many sourcemap files we have. 
   - Question regarding this issue. https://discuss.newrelic.com/t/upload-sourcemap-to-the-same-url-without-am-unique-release-name-id-nextjs/187994


src/utils/newRelicSnippet.tsx
```diff
@@ -0,0 +1,87 @@
...
+      sa: 1
+    };
+
+    window.NREUM.loader_config = { ...defaultConfig, ...config };
+    window.NREUM.info = { ...defaultInfo, ...info };
+  }
+}
+
+// Setup the Newrelic browser agent config for different deploy environment
+export function NRBrowserSetup(deployment: string | undefined) {
```
Toxiapo, line 59: This is to switch reporting destinations between QA and PROD, however, NODE_ENV is probably not the correct indicator here, we should maybe detect if the URL is "openebooks" or "staging.openebooks then set the config accordingly.

next.config.js
```diff
@@ -177,6 +194,31 @@ module.exports = async (_phase, { _defaultConfig }) => {
...
+            : "http://localhost:3000"; // localhost for debugging, but we probably don't need it
+        const NRconfig = new NewRelicSourceMapPlugin({
+          applicationId: applicationId,
+          apiKey: NEW_RELIC_API_KEY,
+          staticAssetUrl: staticAssetUrl,
+          staticAssetUrlBuilder: (url, publicPath, fileName) => {
+            const removeLastCharIfSlash = s => (s ? s.replace(/\/$/, "") : "");
+            return `${removeLastCharIfSlash(url)}${removeLastCharIfSlash(
+              publicPath
+            )}/${fileName.replace("[", "%5B").replace("]", "%5D")}`;
```
Toxiapo, line 213: Encoding `[` and `]` before uploading

src/pages/_app.tsx
```diff
@@ -38,6 +39,24 @@ if (process.env.NODE_ENV === "development" && !IS_SERVER && REACT_AXE) {
...
+    ? "production"
+    : /^staging\.openebooks\.org/.test(window.location.hostname)
+    ? "staging"
+    : undefined;
+};
+
+// Start New Relic Browser Monitoring
+if (!IS_SERVER) {
+  const envrionment = productionOrStaging();
+  NewRelicBrowserSetup(envrionment);
```
Toxiapo, line 55: Since this is a frontend monitor, much like GA, we will check the URL here to determine which code snippet to add to the browser.

hokei (commented):
IMO as long as it doesn't make the code worse, it's OK to get this to wherever we can get it, and launch.  We always expect that this is going to need tweaking as we start getting a better sense of how we use the data.

Toxiapo:
@hokei @samanthaandrews I tried to do a test deploy to staging but it wasn't working : ( The GH action build showed that the build was successful, but the code was not updated on the page, it feels like it's building the app from the staging branch, is merging the PR our only option to see this live?

samanthaandrews (approved):

PR #445, opened 2022-06-29
OE-541: Smooth parallax

samanthaandrews:
I noticed that the parallax effect didn't behave smoothly on Firefox and Safari, so I updated to use a CSS-only solution. Now, instead of using a scroll event handler, the parallax hero image is a div with a fixed background image. This means the content below scrolls _over_ the background image, creating a parallax effect without *technically* being parallax. No more jumpiness on Firefox or Safari.


Toxiapo (approved):
🎉

PR #446, opened 2022-07-08
Google analytics

kristojorg:
This adds the Google Tag Manager script and tracks page views and all link clicks. We use Google Tag Manager instead of the Google Analytics script because we want to track all link clicks and Google Analytics alone doesn't handle this. Google Tag Manager then pushes event data to our Google Analytics dashboard. Basically, we have a Google Tag Manager account where we can configure different events that we want to track. Then, we configure those events to push the data to our Google Analytics account.

This [blog post](https://www.analyticsmania.com/post/google-tag-manager-vs-google-analytics/) helped me understand how GTM and GA are different and how they integrate.

Within the Google Tag Manager account, I set up two different tags. Both connect to the Google Analytics ID that Kristo created.
1. Tracking all page views using the "History change" event instead of a "Page view" event. This means we don't have to add any code to our routes to make sure page views are getting sent to Google Analytics.
2. Tracking all link clicks.


Toxiapo (approved):

.github/workflows/release-prod.yml
```diff
@@ -88,7 +88,8 @@ jobs:
           --build-arg NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA=$IMAGE_TAG_SHORT \
           --build-arg NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF=$IMAGE_TAG_SHORT \
           --build-arg GH_PACKAGE_AUTH_TOKEN=$GH_PACKAGE_AUTH_TOKEN \
-          --build-arg NEW_RELIC_APP_NAME="Open Ebooks" .
+          --build-arg NEW_RELIC_APP_NAME="Open Ebooks" \
+          --build-arg NEXT_PUBLIC_GTM_ID=GTM-572L9MV .
```
samanthaandrews, line 92: @Toxiapo do you think this needs to be GTM id value needs to be in quotes? `"GTM-572L9MV"`

Toxiapo: I _think_ what you have there should work

src/analytics/GoogleTagManager.tsx
```diff
@@ -1,30 +1,33 @@
...
   return (
-    <noscript
-      dangerouslySetInnerHTML={{
-        __html: `<iframe src="https://www.googletagmanager.com/ns.html?id=${GTM_ID}" height="0" width="0" style="display:none;visibility:hidden"></iframe>`
-      }}
-    />
+    <noscript>
+      {/* eslint-disable-next-line jsx-a11y/iframe-has-title */}
+      <iframe
+        src="https://www.googletagmanager.com/ns.html?id=${GTM_ID}"
```
Toxiapo, line 26: This should be a template literal, not a string.

samanthaandrews: you are right. fixing now!

PR #448, opened 2022-07-14
OE-487: setup New Relic Error tracking

Toxiapo:
- [x] Remove Bugsnag
- [x] Define metadata fields for new relic reports
- [x] Setup error reports
- [x] Add env var in AWS container for new relic API and license key
- [x] Remove new relic GITHUB_SECRETS


src/dataflow/opds1/parse.ts
```diff
@@ -78,13 +78,11 @@ function isSupportedOpenEBookLink(link: OPDSLink): link is OPDSAcquisitionLink {
...
       return true;
     }
-    track.error(
-      {
-        name: "Unsupported book",
-        message: "A non-AxisNow book format has been detected in the feed"
-      },
-      { severity: "warning" }
-    );
+    track.error(new Error("Unsupported book"), {
```
samanthaandrews, line 81: So the issue with this error is that it will fire a bunch of times in Production because our OPDS feeds currently include both Adobe AND AxisNow links. I'm thinking for now, until this [issue I created](https://jira.nypl.org/browse/OE-587) is addressed, we shouldn't track this error to New Relic.

src/types/globals.d.ts
```diff
@@ -1,4 +1,15 @@
...
+    noticeError(
+      error: Error,
+      customAttributes?: Record<string, string | number>
+    );
+  };
+}
+
+declare module "!file-loader*" {
+  const url: string;
+  export default url;
```
samanthaandrews, line 14: This !file-loader stuff can be removed -- I removed it as part of my offline reading PR but it looks like it snuck back in

samanthaandrews (approved):
Just a couple small things but otherwise looks great!

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ### UNRELEASED CHANGES
 
+- Add: OE-487: Replace BugSnag with New Relic and setup error tracking
```
samanthaandrews, line 5: Let's go ahead and put this in the 4.7.13 release, yeah? Looks like it will probably make it into staging soon and since we already have that 4.7.13 release branch to production, might as well include this too.

Toxiapo: Yeah, we might have to wait for DevOps to deploy the AWS vars to production Monday.

src/dataflow/opds1/parse.ts
Toxiapo: got it! Removing it now.

CHANGELOG.md
Toxiapo: Ok, we are good now. Let's do it!

PR #458, opened 2022-08-02
OE-430: Add production PR template

samanthaandrews:
https://jira.nypl.org/browse/OE-430

Once merged into `staging`, this template will populate every PR, but since it is commented out by default, you will uncomment and use the checklist if making a PR to production branch.


README.md
```diff
@@ -282,7 +282,7 @@ When creating links using `<Link>`, you don't need to worry about whether it is
 # Deploying to production with Github Actions
 
 1. Create a feature branch and update the CHANGELOG.md to include the issues that are going into this release. Also update the version number in `package.json`. Make a pull request with the base branch as `staging`. Once this PR has been approved, squash and merge it to `staging`.
-2. Make a pull request with the base branch as `production` and compare `staging`. Give the pull request a name that includes the version number, for example `Release 4.7.2`. Add a description if it adds additional context for those reviewing, otherwise it's ok to leave it blank.
+2. Make a pull request with the base branch as `production` and compare `staging`. Use the production PR template when you make this pull request by adding `template=production.md` query parameter to the end of the github URL. Give the pull request a name that includes the version number, for example `Release 4.7.2`. Add a description if it adds additional context for those reviewing, otherwise it's ok to leave it blank.
```
Toxiapo, line 285: Is this a manual step we have to take for the template to show up every time we make a PR to prod?

samanthaandrews: Yeah. I've been reading the GH docs and it seems this is the only way to select a template unless we want to make this part of the template for every PR—which is still an option! Not ideal, but we could instead make this part of the template for every PR and have a heading that says `Release to prod checklist` so that it is clear that you can kind of ignore the checklist unless you're making a PR to production? Wait, actually, yeah I think this might be better! I can make the template happen on every PR, but it will remain commented out unless you explicitly uncomment it. How does that sound?

Toxiapo: yeah sounds good!

Toxiapo: I am also curious if there's a way to remind us to either use "squash" or "merge" as part of the checklist. Maybe on a different ticket tho.

samanthaandrews: I added this!

Toxiapo (approved):

PR #460, opened 2022-08-04
OE-499: Add architecture decision log

samanthaandrews:
- Adds two architecture decision record items: one to signify that we will start documenting architecture decisions, and another to note our decision to use Chakra UI component library
- Cleans up the README


Toxiapo (approved):
LGTM!

PR #461, opened 2022-08-05
OE-625: Disable rightclick and copy&paste function

Toxiapo:
This PR disables the right click and copy&paste function on the reader.


samanthaandrews (approved):
Code looks great. I think this is a great solution.

public/js/reader.js
```diff
@@ -0,0 +1,7 @@
+document.addEventListener("copy", e => {
```
samanthaandrews, line 1: Can you add a comment here that explains why we're disabling this? I think it's pretty self-explanatory, but also we might forget in the future.

samanthaandrews (approved):
Let's add a comment that explains why we're disabling copy/paste but otherwise LGTM!

PR #462, opened 2022-08-10
bump version number

Toxiapo:


samanthaandrews:
I think the `package-lock.json` also needs to update with the 4.7.14 version number, right? Maybe just run `npm i` and push up before we merge this in.

Toxiapo:
> I think the `package-lock.json` also needs to update with the 4.7.14 version number, right? Maybe just run `npm i` and push up before we merge this in.

Yes, good catch!

samanthaandrews (approved):

PR #464, opened 2022-08-16
OE-630: Add policies redirect

samanthaandrews:
Redirect `/policies` to `/privacy`


Toxiapo (changes requested):
Since this is the only change, let's bump the version number in this PR as well.

Toxiapo (approved):

PR #466, opened 2022-08-16
Fix GTM build arg

samanthaandrews:
The google tag manager script wasn't getting added to the site because I had put the GTM_ID argument in the wrong place in the Dockerfile. Now it's defined before the build step so the env var is getting built into the image as expected.


Toxiapo (approved):
Let's go!

PR #469, opened 2022-08-24
OE-546: Build contact page with new design and sendGrid

Toxiapo:
This PR includes the redesigns of the /contact page with sendGrid and react-hook-forms


src/pages/api/contact/index.ts
```diff
@@ -0,0 +1,63 @@
...
+const Contact: NextApiHandler = async (req, res) => {
+  const { firstName, lastName, email, inquiry, message } = req.body;
+
+  if (!firstName || !lastName || !email) {
+    res.status(400).json({ error: "Please fill out all the required fields." });
+  }
+
+  try {
+    await sendgrid.send({
+      to: "",
```
Toxiapo, line 16: I think we can put the recipient(Miguel's email) in the GITHUB SECRET instead of hard coding it here. What's your thought?

src/pages/api/contact/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      from: "noreply@nypl.org",
```
Toxiapo, line 17: Miguel wanted this to be the email address of the sender, but sendGrid doesn't let you just add any emails here. I can ask Kristo what he can do with this when he's back next week.

src/pages/api/contact/index.ts
samanthaandrews: I thought the email was going to be sent to `info@openebooks.org`? It makes more sense that this would be a generic group email instead of Miguel's personal email since multiple people can be added to a group email and we would have coverage if he is OOO or something like that.

src/pages/contact.mdx
```diff
@@ -1,7 +1,17 @@
...
+  Divider
+} from "@chakra-ui/react";
+
+import ContactForm from "marketing/components/ContactForm";
+import SectionDivider from "marketing/components/SectionDivider";
 
 # Contact Us
 
-How can we help? If you have any questions about our free online reading platform for kids and teens, [email us](mailto:info@openebooks.org) directly.
+How can we help? If you have any questions, please fill out the form below or email us directly to answer any of your specific questions about our free online reading platform for kids and teens.
```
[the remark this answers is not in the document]
Toxiapo: yes that's a good point

src/pages/api/contact/index.ts
Toxiapo: I think `info@openebooks.org` makes sense because that's what we have on the original contact form.

samanthaandrews (approved):
LGTM! Form is submitting for me and code looks good. I'm thinking we should pass it on to Jen & Ayan for design review and QA review so both can happen at the same time.

PR #477, opened 2022-09-01
Hotfix use Chakra footer as footer everywhere

samanthaandrews:
Use the chakra footer from the homepage in the web app too


Toxiapo (approved):
Nice! So we can use Chakra components even if the file is using `/** @jsxImportSource theme-ui */`

Toxiapo (changes requested):
Could you also remove the old component and update the test?

samanthaandrews:
@Toxiapo Good catch. I made those updates!

Toxiapo (approved):
LGTM, thanks!

PR #478, opened 2022-09-07
OE-598: Enable page regeneration

Toxiapo:
This PR adds the revalidate prop to the book and collection page in the `getStaticProps` function. Statically generated pages fetch data on build time, this raises issue when the API data was updated but the app is still serving the cached data. Adding the revalidate props will always update the cache and server new data without the need for us to redeploy.


src/pages/index.tsx
```diff
@@ -280,13 +280,14 @@ export async function getStaticProps() {
...
+    },
+    revalidate: 60 * 10 // 10minute
   };
-}
+};
 
 export const IS_PRODUCTION_SERVER =
   typeof CM_URL === "string" &&
-  /https:\/\/(www\.)?circulation\.openebooks\.us/.test(CM_URL)
+  /https:\/\/(www\.)?circulation\.openebooks\.(us|org)/.test(CM_URL)
```
samanthaandrews, line 290: The CM_URL should always end in `.us` since this is the url to the backend api routes.

Toxiapo: u r right, updating now.

samanthaandrews (approved):

PR #480, opened 2022-09-13
OE-501: Clean up env vars and docker build file

samanthaandrews:
https://jira.nypl.org/browse/OE-501

Cleans up environment variables that we are using in our Dockerfile.

We're now passing all env vars through in `next.config.js` which means we don't have to set them as runtime vars in Docker. This means that we don't have to keep any environment variables in the task config JSON in AWS -- all env vars are available into the Docker image when we run `docker build`.


samanthaandrews:
I will make a subsequent PR to the Stash repo to remove environment variables from our cloud formation templates. I have a feeling we will get confused about where env vars need to get updated if we don't do this. As far as I can tell, we don't need to add env vars to AWS at all since all of the env vars are built into the image at build time.

.github/workflows/release-prod.yml
```diff
@@ -50,13 +50,6 @@ jobs:
       - name: checkout
         uses: actions/checkout@v2
 
-      - id: nvmrc
-        run: echo ::set-output name=NODE_VERSION::$(cat .nvmrc)
-
-      - name: Set up node with version from nvm
-        uses: actions/setup-node@v2
-        with: { node-version: "${{ steps.nvmrc.outputs.NODE_VERSION }}" }
-
```
samanthaandrews, line 59: we don't use this node version anywhere because the docker file builds from node:16-alpine, so it uses that as the base node image

.github/workflows/release-prod.yml
```diff
@@ -76,22 +69,27 @@ jobs:
 
       - name: Build, tag, and push image to Amazon ECR
         env:
+          NEXT_PUBLIC_CM_URL: https://qa-circulation.openebooks.us
```
Toxiapo, line 72: Typo? it's using the qa URL

samanthaandrews: oops good catch

.github/workflows/release-prod.yml
```diff
@@ -76,22 +69,27 @@ jobs:
...
+          GH_PACKAGE_AUTH_TOKEN: ${{ secrets.GH_PACKAGE_AUTH_TOKEN }}
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           ECR_REPOSITORY: openebooks-nextjs
           IMAGE_TAG: ${{ github.sha }}
           IMAGE_TAG_SHORT: ${{ steps.vars.outputs.sha_short }}
-          GH_PACKAGE_AUTH_TOKEN: ${{ secrets.GH_PACKAGE_AUTH_TOKEN }}
+          NEW_RELIC_APP_NAME: "Open Ebooks"
```
Toxiapo, line 78: This technically should be "Open Ebooks (PROD)" but we can keep it as is now. 

When I first set the monitoring, this was the label I used everywhere but I renamed it in the dashboard so that's why it's still recognizing this old label.

Toxiapo (commented):
So we need to add the secrets to the GitHub Secrets as our next step right?

Could you also remind me who I should ask for the stash access?

Toxiapo (approved):

.github/workflows/release-prod.yml
samanthaandrews: I will change it to "Open Ebooks (PROD)"!

PR #482, opened 2022-09-15
OE-648/upgrade-next

kristojorg:
This PR upgrades to NextJS 12.3 and makes required refactors.

- brings all middleware logic into a single handler (and into a single file since it's now all centralized there)
- removes the access restrictions for book detail pages.
- I added a lot of comments in `middleware.ts` to try and clarify what's happening, but it's still a rather complex set of events where we have to check a lot of conditions, and any given check might result in us responding immediately to the middleware with a redirect or rewrite.
- The middleware.ts file now exports a `config` which tells NextJS to only run the middleware on requests to the specified URLs. 
- Updates to React 18
- I ran a codemod to fix React 18 breaking changes, namely that `Children` now has to be explicitly typed even when using `React.FC`. 
- Updates to Chakra v2 and fixes associated issues
- Switches us to the NextJS recommended Eslint settings
- Runs prettier on full codebase (it previously only ran on src files I believe)


kristojorg:
Closing this in favor of #492

PR #485, opened 2022-09-16
Release version 4.7.17

samanthaandrews:
## ✅ Merging to production checklist ✅

- [x] Do these changes modify environment variables? If so, have you updated env vars in the Dockerfile, the GH Action Workflows & submitted the change to AWS through DevOps?
- [x] Has someone run through the [QA Testing Checklist](https://docs.google.com/document/d/1Sm_mFNAlbUrjT4bfJt5boQMeIA7LoPwoPa4uXqp8kFg/edit?usp=sharing) on https://staging.openebooks.org/ and confirmed everything is working as expected?

⚠️ Make sure you select the "Create a merge commit" option when you merge PRs to the `production` branch.


Toxiapo (approved):
Looks good!

src/pages/_app.tsx
```diff
@@ -40,7 +40,7 @@ const MyApp = (props: AppProps) => {
   );
 };
 
-if (process.env.NODE_ENV === "development" && !IS_SERVER && REACT_AXE) {
```
kristojorg, line 43: Ok one thing I'm seeing is that I think I specifically inclined this check this way so that react-axe would be removed in production by the compiler. It wasn't able to remove it when checking using variables. Not sure if that could cause this though. But we should make a ticket to investigate that.

kristojorg (commented):
One thought here

src/pages/_app.tsx
samanthaandrews: I did some testing on this and checking for the variable `IS_DEVELOPMENT` is working as expected. It is getting set to true when developing locally, and false when doing a production build.

PR #487, opened 2022-09-22
Oe 647/pw automate age lane verification

ayan1229:


samanthaandrews (commented):
In order to get the playwright tests to pass in CI, you'll need to add the env secrets to the `.github/workflows/playwright.yml` file. This is because the `.env.local` file is not committed to git so by default Cucumber/Playwright doesn't have access to them.

samanthaandrews (approved):
Looks great! Go ahead and "Squash and merge"!

PR #488, opened 2022-09-26
OE-521: Add base style stylesheet for FXL books

Toxiapo:
This PR adds the base stylesheet to the FXL books. This PR stems from a bug we found in the previous releases where some FXL contents are showing on the left side of the screen, this is because FXL contents come with their styles, and we don't set styles to FXL explicitly.

The base style I added in this PR does not address all the centering bugs but only the ones that don't come with a fixed width/height. 

Book Tested: Big Cat, Little Cat; Spill Zone 1; Speak


samanthaandrews (approved):

public/css/fixed/default.css
```diff
@@ -0,0 +1,12 @@
+/* A single element layout with left-aligned position*/
+body {
+    margin-left: auto;
+    margin-right: auto;
+}
+  
+@media all and (max-width: 37.5em) {
+    /* When the viewport is larger than then content, the content would shift to right, this will reset the position to default */
```
kristojorg, line 8: Is this just a ballpark max width that we think will be larger than the content?

kristojorg (commented):
Just the one question. Looks good though!

public/css/fixed/default.css
Toxiapo: Yes, so I intended to target the small viewports here(37.5em was the max-width specified in our breakpoint doc). This was to combat where we had to set the `scale` and `transform` properties on the iframe during render, which by default `scale content to the center of the elm`, this is fine when we have enough space, but it breaks for the smaller screen. I am just telling the browser to `scale to the top left corner` here

kristojorg (approved):
Cool, looks good let's give this a shot!

kristojorg (approved):

public/css/fixed/default.css
Toxiapo: It might be better if we include this logic in the webreader and do not have to check it here because we can easily get the content size when we render the iframe. I will create a ticket.

cypress/tests/web-reader/prefetch-content.ts
```diff
@@ -16,11 +16,14 @@ describe("Users can read a book offline", () => {
...
+    cy.log("Visit the book page detail");
+    cy.visit(ALL_ACCESS_DETAIL_BOOK_PATH_ALMOST_MIDNIGHT);
+
+    cy.log("Click on the read button");
+    cy.findByRole("link", { name: "Read" }).click();
 
     cy.log("all 7 css, font and script prefetch links exist");
-    cy.get("link[rel=prefetch][as=style]").should("have.lengthOf", 6);
+    cy.get("link[rel=prefetch][as=style]").should("have.lengthOf", 7);
     cy.get("link[rel=prefetch][as=script]").should("have.lengthOf", 1);
```
Toxiapo, line 27: I had to modify this test from "Read book via a direct link" to "Visit the book detail page and click on read button" for it to work, otherwise I would get a decryptor error on my tests(local works). Have you ran into a similar issue before? @samanthaandrews @kristojorg

PR #490, opened 2022-09-29
OE-580: New app nav

samanthaandrews:


cypress/support/commands.ts
```diff
@@ -82,11 +82,17 @@ Cypress.Commands.add(
     cy.visit(FIRSTBOOK_SIGNIN_PATH);
 
     cy.findByRole("textbox", { name: "Access Code input" }).type(username, {
-      log: false
+      log: false,
+      scrollBehavior: false
```
samanthaandrews, line 86: Since the header is now sticky, Cypress tests were failing because they navigate to an element and then the header covers the element. The workaround is to change the `scrollBehavior: false` as you'll see throughout our tests. Annoying.

src/components/chakra-ui/theme/foundations/color.ts
```diff
@@ -43,6 +43,7 @@ const colors: Colors = {
     },
     gray: {
       dark: "#8A8A8A",
+      darkish: "#909090",
```
samanthaandrews, line 46: i realize this naming is pretty terrible lol. any ideas for a better name??

kristojorg:
image
So I think changing the theme breakpoints might not be a good idea, since the whole marketing site currently depends on those breakpoints. In general, the designs should specify how something responds to the same breakpoints everywhere, though I understand we didn't have designs which accommodated  the "All Books" button. In this case, I think it's probably better to add a breakpoint inline for that specific component, not change them all. We spent significant time with Jen refining how the marketing homepage responds to different sizes, and future designs will use those same breakpoints, so I feel like we shouldn't mess with them

kristojorg:
I also am wondering if we should disable the search button once a user initiates a search but before the page has reloaded? Currently it feels like there is a confusing delay after clicking search where nothing happens. Or maybe a loading state?

src/components/chakra-ui/theme/foundations/color.ts
kristojorg: lol hmm "darker" is slightly better buuuut idk

cypress/support/commands.ts
kristojorg: Oh that is rly annoying haha

kristojorg (changes requested):
whoops meant to request those few changes before approving

src/components/chakra-ui/theme/foundations/color.ts
Toxiapo: +1 on darker haha, or maybe ask/update Jen, maybe it will be used on several other places, in that case, we might want a consistent name for it.

src/components/chakra-ui/AuthButton.tsx
```diff
@@ -0,0 +1,113 @@
...
+import useGetCredentials from "hooks/useGetCredentials";
+import { SIGNIN_PICKER_PATH, SIGNOUT_PATH } from "utils/constants";
+import { SignIn, SignOut } from "components/chakra-ui/icons";
+
+import { menuLinkStyle } from "components/chakra-ui/header/utils";
+
+export type AuthButtonProps = {
+  colorScheme?: string;
+  as?: React.ReactElement;
+} & ThemingProps &
```
Toxiapo, line 19: I don't think you need 'ThemingProps", ChakraProps should be enough.

src/components/chakra-ui/icons/Educators.tsx
```diff
@@ -1,8 +1,8 @@
-import { SxProp } from "theme-ui";
 import React from "react";
+import { Icon } from "@chakra-ui/react";
 
-const EducatorsIcon = (props: React.SVGProps<SVGSVGElement> & SxProp) => (
-  <svg
+const Educators = (props: React.ComponentProps<typeof Icon>) => (
+  <Icon
     width="24"
     height="24"
```
Toxiapo, line 7: Since the svg is being replaced with an `Icon`, these values won't work anymore, you might want to update them to something like `width="24px" / height="24px"`

Toxiapo (changes requested):
Codes look good!

I agree with Kristo regarding the breakpoint, if it's just a one-off breakpoint, we should just inline it.

Toxiapo:
Header UI:
Can we add a little margin to the Search component? It's touching the search bar If you hover over "All Books,"
Screen Shot 2022-10-03 at 2 53 40 PM

cypress/tests/errors/useSwr.ts
```diff
@@ -81,7 +81,7 @@ describe("Mock endpoint that catches 401, 404, and 502 errors in the ErrorBounda
 
     cy.findByRole("button", { name: "Borrow this book" })
       .should("exist")
-      .click();
+      .click({ scrollBehavior: false });
```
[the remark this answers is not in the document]
samanthaandrews: Good idea, I made it the default

src/components/chakra-ui/Carousel/Top10Badge.tsx
```diff
@@ -66,7 +66,6 @@ const Top10Large = (props: SVGProps<SVGSVGElement> & BadgeProps) => {
       </g>
       <defs>
         <filter
-          id="a"
```
samanthaandrews, line 69: @Toxiapo this was the issue with the top10 badges not displaying. I'm not sure how these ids got removed from the chakra svgs?? It must have been a prettier or eslint thing?

kristojorg: How strange! Nice catch

kristojorg:
| If you are on the Home page(web app home), the "All Books" button should be on hover state.

Unfortunately, this will be somewhat difficult currently because the actual URL for different home pages (high school, lower, middle, etc), whereas the all books redirects to "/app" which in turn redirects to "/app/collection/:collectionUrl" after determining your root collection

samanthaandrews:
Ok I've addressed the comments from @kristojorg, @Toxiapo, and @jotten.

- The MarketingHeader is now using custom breakpoints so I didn't change the original breakpoints we were using.
- I created a ticket to work on storing a user's root catalog url so that we can use that value to display the correct selected state on the "All books" header link: https://jira.nypl.org/browse/OE-674
- I will chat with Jen about improving the search loading state and create tickets accordingly.

Toxiapo (approved):
LGTM!

kristojorg (approved):
Looks good!

kristojorg (approved):
Yeehaw!

PR #491, opened 2022-09-30
OE-659: Update contact form email address for non-prod submissions

samanthaandrews:
I want us to be able to test the contact form without cluttering Miguel's inbox and mixing test submissions with real ones. The google group email address openebooksdev@nypl.org is currently set up to send emails to me, @Toxiapo , @ayan1229 , and @kristojorg .


src/pages/api/contact/index.ts
```diff
@@ -8,6 +8,10 @@ sendgrid.setApiKey(SENDGRID_API_KEY as string);
 
 const Contact: NextApiHandler = async (req, res) => {
   const { firstName, lastName, email, inquiry, message } = req.body;
+  const toEmailAddress =
+    req.headers.host === "www.openebooks.org"
```
Toxiapo, line 12: I thought `host` only returns the domain without the subdomain, but I can't find proof so I believe you here!

Toxiapo (approved):
Don't we all love clean inbox!

src/pages/api/contact/index.ts
samanthaandrews: Good catch. As far as I can tell, host does include subdomain. When we make requests to baker and taylor, the host looks like this `contentcafecloud.baker-taylor.com` in the Network tab. So it's subdomain + domain but without any path or `http://`.

samanthaandrews: I created a util function called `determineDeployEnv(hostname)` that will determine whether the deployment is `staging`, `production`, `preview`, or undefined and I made it so that the New Relic setup also uses this function since it was doing a similar check.

Toxiapo (approved):

PR #498, opened 2022-10-14
Disable New Relic in dev mode

samanthaandrews:
A comment that @kristojorg made in the breadcrumbs PR made me realize I could probably clean up some of the logic surrounding the New Relic setup. Previously we were checking for the hostname to determine if we were on staging or production, but I think we can just use the existing `RELEASE_STAGE` env var. Is my thinking right in this?


src/pages/_app.tsx
```diff
@@ -46,16 +45,12 @@ if (IS_DEVELOPMENT && !IS_SERVER && REACT_AXE) {
...
-  return determineDeployEnv(window.location.hostname);
-};
-
 // Start New Relic Browser Monitoring
-if (!IS_SERVER) {
-  const environment = productionOrStaging();
-  NewRelicBrowserSetup(environment);
+if (
+  !IS_SERVER &&
+  (RELEASE_STAGE === "staging" || RELEASE_STAGE === "production")
```
samanthaandrews, line 51: I thought that maybe we didn't need to be checking for the actual hostname and could instead check for the `RELEASE_STAGE` env var. I think this is ok, but want to check with devs to be sure.

Toxiapo: Yup! I think this is fine.

Toxiapo (approved):

PR #503, opened 2022-10-24
Check for New Relic API key before browser initialization

samanthaandrews:
Fixes the client-side react error happening on production builds on staging branch


kristojorg (approved):
Looks good to me! LGTM

PR #504, opened 2022-10-25
OE-648: Disable Cypress & Add middleware Jest tests

samanthaandrews:
- Adds more thorough unit testing for middleware (we were previously at ~75% statement coverage for middleware and now we are at 93.75%)
- Ensures credentials returned by `getCredentials` are sent along as a cookie in middleware responses so that updated credentials are always in the browser cookie


src/middleware.ts
```diff
@@ -261,6 +261,7 @@ function followCollectionRedirect(
   request: NextRequest,
   collectionResponse: Response
 ) {
+  console.log("COLLECTION RESPONSE!!", collectionResponse);
```
samanthaandrews, line 264: I was attempting to see when we get into this code path and wasn't ever getting collectionResponse.redirected to equal `true`. @kristojorg when you're back I'd like to chat it through to understand if this is a bug in the CM or if I'm misinterpreting what the conditions under which we would get here.

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
-- Add: OE-580: New web app navigation
+
+- Refactor: OE-648: upgrade to NextJS 12.3 and mode middleware into a single handler & add middlware unit tests
```
kristojorg, line 5: And disable Cypress I would mention also here

src/__tests__/middleware.bookPath.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    mockedReq.nextUrl.searchParams.set("signinError", "Something went wrong");
+
+    const response: NextResponse = await middleware(
+      mockedReq,
+      {} as NextFetchEvent
+    );
+
+    expect(response.status).toBe(200);
+    expect(redirectSpy).toHaveBeenCalledTimes(0);
+    expect(nextResponseSpy).toHaveBeenCalledTimes(1);
```
kristojorg, line 28: Just a note for the future - I think a better way to test this would be to treat it as a pure function. Just given a certain input (request) do we get the output (response) we expect, instead of asserting on what happens inside the function itself. For example, instead of testing whether these were called, we should be able to test that the response url is the same as the input request url

Anyway, just a thought for the future

src/middleware.ts
```diff
@@ -61,8 +61,8 @@ export async function middleware(request: NextRequest, _ev: NextFetchEvent) {
     );
     if (r3) return r3;
     const r4 = followCollectionRedirect(request, collectionResponse);
-    if (r4) return r4;
-    return NextResponse.next();
+    if (r4) return setAuthCookieServer(r4, credentials);
+    return setAuthCookieServer(NextResponse.next(), credentials);
```
kristojorg, line 65: Interesting, so you had to re-set the credentials even when just calling `next`? Weird that it works that way

src/middleware.ts
```diff
@@ -116,7 +116,7 @@ export async function middleware(request: NextRequest, _ev: NextFetchEvent) {
    * /signin/method, /signin/method/clever:
    *  - Check if there is a signinError in the params and redirect to /signin if so
    */
-  if (isPathMatch(url, "/signin/method*")) {
+  if (isPathMatch(url, "/app/signin/method*")) {
```
kristojorg, line 119: Good catch

src/middleware.ts
kristojorg: This should happen whenever a user requests the root lane directly: `/app/collection/:collectionUrl` where collection url is the all access root lane. 

So if you log in as an early grades user and request this path: `/app/collection/https%3A%2F%2Fqa-circulation.openebooks.us%2FUSOEI%2Fgroups%2F` you should hit this

src/middleware.ts
```diff
@@ -359,10 +360,23 @@ function redirectToSignin(
  * Delete the auth cookie from the response.
  */
 function removeAuthCookie(response: NextResponse) {
+  console.log("we are inside removeAuthCookie");
```
kristojorg: Stray console

kristojorg (changes requested):
Looks good to me! Just a stray console log mainly

src/__tests__/middleware.bookPath.test.tsx
Toxiapo: @samanthaandrews I have a similar tests here fyi.
https://github.com/NYPL-Simplified/circulation-patron-web/blob/32b2be7d54ab8e29dd118e52554f84baf45b35af/src/__tests__/middleware.test.tsx#L75

src/middleware.ts
samanthaandrews: From writing these test I learned that by default calling NextResponse.next() does not, by default, pass the cookies from the request. In our case, since we are calling `getCredentials()` in our middleware which makes a call to `verifyToken()` and sometimes returns a brand new token in the case the token is about to expire, we need to make sure that that new token is set in the cookie.

In short, when we are on the server and we call `getCredentials()`, we need to send the credentials that are returned as a cookie in the Next response. Otherwise the token that is already set in the browser could be an outdated token that is no longer valid.

src/middleware.ts
samanthaandrews: Ok yes I think I've identified a bug (and it may be on the CM side), but when I am logged in as an early grades user, a middle grades user, or a high school user, and I navigate to that path you posted above ⬆️ I am indeed able to see all of the lanes. I am not able to click into each collection to _See more_ content (I hit a 404), but the lanes are visible. This is happening on production, so it is not a result of any of the latest changes made to middleware.

From my testing locally, we are checking that `collectionResonse.redirected` is true, but it is always returning false.

Jen noticed this issue on Thursday, October 20 and posted about it in Slack: https://nypl.slack.com/archives/C0153AFA1RS/p1666284666558289

I created https://jira.nypl.org/browse/OE-720 to investigate the issue. Any context you might have @kristojorg or @Toxiapo please put it in the ticket!

Toxiapo (approved):

PR #505, opened 2022-10-26
OE-713: Explicitly set RELEASE_STAGE instead of parsing git refs

samanthaandrews:
The parsing of git refs to pull out the branch name has not been working as expected. See the screenshot below for our recent production release which should have set the RELEASE_STAGE to `production` but instead it was getting set to `development` because the GIT_BRANCH value, which RELEASE_STAGE var was dependent on, was not parsing out the branch name to be `production` but rather a sha value.

Screen Shot 2022-10-26 at 11 57 07 AM

I spent some time trying to parse the GIT_BRANCH name to ensure it refers to the name of the feature branch, but the behavior of how a github action sets their branch name values isn't as straightforward as expected. Instead, I think it is much clearer to set the RELEASE_STAGE value explicitly as an environment variable in our docker build step.


kristojorg (approved):
This looks good, thanks!

Toxiapo (approved):
Thanks! Yea defining the GIT_BRANCH ourselves makes sense

PR #506, opened 2022-10-27
OE-695 & OE-708: Remove unused `tealNav` color scheme and address contact form accessibility

samanthaandrews:
In my menu nav PR I created a new button color scheme called `tealNav`, but this was not right. I only needed to use the existing `teal` colorScheme value and remove the box shadow for the auth button in the nav 👍 This PR cleans up this issue.

I also addressed some accessibility feedback in the Contact form and the Header component.


samanthaandrews:
This PR is under accessibility review by Clare O'Keefe. I'll mark it as "Ready for review" once she has provided feedback.

src/components/chakra-ui/header/utils.ts
```diff
@@ -78,7 +78,7 @@ export const appHeaderStyles: HeaderStyles = {
   bg: "ui.gray.xlight",
   linkColor: "ui.black",
   hoverLinkBg: "ui.gray.light",
-  authButtonVariant: "tealNav",
+  authButtonColorScheme: "teal",
```
kristojorg, line 81: Using context to pass these things down feels a little off to me, I do think there is probably a simpler way to compose these components and reuse the styles you want to reuse. This way is a little bit difficult to wrap my head around when I want to work on the header. Would it make sense to extract whatever re-used parts you want into a single shared component and then build separate `MarketingHeader` and `AppHeader` from those? 

Doesn't have to be addressed here, just a thought I had last time I was working in this code

samanthaandrews: I agree that using context here isn't the most ideal. We are using a separate `<MarketingHeader />` and <AppHeader />` that both wrap the shared `<Header />` so I could easily pass these styles down through to the descendants, but it would involve quite a bit of prop drilling which is what I was trying to avoid.

With the latest homepage designs that Jen has created (these are supposed to be shown to users in a user research session), we won't need to keep track of two different header styles like this, so using context here was also me anticipating that it would be easier to remove this context entirely in the near future and directly set the styles that we need within each component.

kristojorg: Nice that sounds great!

src/components/chakra-ui/icons/Mail.tsx
```diff
@@ -3,20 +3,30 @@ import { Icon } from "@chakra-ui/react";
 
 const Mail = (props: React.ComponentProps<typeof Icon>) => (
```
samanthaandrews, line 4: I updated the Mail icon to the new one from the Iconography Audit file: https://www.figma.com/file/Y0wri3clDvVO8ftkX6HmVD/OEB-Iconography-Audit?node-id=1678%3A2135

We still need to update all the other icons, and there is a ticket for that here: https://jira.nypl.org/browse/OE-692

src/pages/app/read/[bookUrl].tsx
```diff
@@ -16,7 +15,6 @@ const ReaderPage: React.FC = () => {
   return (
     <Page>
       <main
-        id={CONTENT_ID}
```
Comment (samanthaandrews) [inline src/pages/app/read/[bookUrl].tsx:19]: I don't think it makes sense to include the SkipNavLink or the SkipNavContent here since the header doesn't render in the WebReader anyway, but it made me think that maybe there should be a SkipNav functionality within the web reader itself?

Comment (Toxiapo) [reply src/pages/app/read/[bookUrl].tsx:19]: Yea, this is a good point, typically for screen reading when users knows about the header and just want to start listening the content.

src/components/chakra-ui/ContactForm.tsx
```diff
@@ -135,43 +137,31 @@ const ContactForm = (): JSX.Element => {
...
           <FieldWrapper>
+            <FormLabel>Message</FormLabel>
             <Textarea
-              placeholder="Message"
               {...register(ContactFormFields.message)}
               resize="none"
               rows={13}
               borderColor="ui.gray.dark"
               _placeholder={{ color: "ui.bodyText" }}
+              _focus={{ paddingTop: "20px" }}
```
Toxiapo, line 151: Do we need this? It's shifting the layout when the user focuses on the textarea

src/pages/app/read/[bookUrl].tsx
Comment (samanthaandrews) [reply src/pages/app/read/[bookUrl].tsx:19]: Ok I created this ticket to track it! https://jira.nypl.org/browse/OE-742

src/components/chakra-ui/ContactForm.tsx
samanthaandrews: Good catch. I fixed this by editing the textarea styles in the `form.tsx` theme file. Now there is no layout shift.

Toxiapo (approved):
Looks great!

PR #507, opened 2022-10-28
Oe 698/breadcrumb bug fix

oliviawongnyc:
This PR resolves the breadcrumb issue described in [OE-698](https://jira.nypl.org/browse/OE-698) by only allowing one book to exist at the end of the storedBreadcrumbs array.


src/components/chakra-ui/BookDetail/index.tsx
```diff
@@ -56,8 +56,18 @@ export const BookDetails: React.FC<BookDetailsProps> = ({
 
   React.useEffect(() => {
     setStoredBreadcrumbs(storedBreadcrumbs => {
+      // if there are existing breadcrumbs and the last existing breadcrumb is
+      // a book, then remove the book and add the new breadcrumb to the array,
+      // else, simply add the new breadcrumb to the array
+      const existingBreadcrumbs =
+        storedBreadcrumbs.length &&
+        storedBreadcrumbs[storedBreadcrumbs.length - 1].href.slice(5, 9) ===
```
samanthaandrews, line 64: I think we could use something like `href.includes("/book/")` to avoid having to slice at a specific index

samanthaandrews (changes requested):
One small change, otherwise this looks good. Once updated you can re-request review.

Toxiapo (commented):
Nice! This fixes the issue for me. I will let Sam approve with her requested change.

samanthaandrews (approved):
🚀

PR #508, opened 2022-10-31
Oe 660/pw borrow read return books

ayan1229:


ayan1229:
hi @samanthaandrews 
test is failing due to read-return bug in staging.

playwright/support/mappings.ts
```diff
@@ -46,6 +46,22 @@ export const pages: { [name: string]: Pages } = {
...
   },
+  "book details": {
+    route: "/app/book",
+    regex: /^\/app\/book\//,
+    waitForNavigation: true
+  },
+  "borrowed book detail": {
+    route:
+      "/app/book/https%3A%2F%2Fqa-circulation.openebooks.us%2FUSOEI%2Fworks%2FAxis%2520360%2520ID%2F0017370149",
+    regex: /^\/app\/book\//,
```
samanthaandrews, line 57: We should change this regex to the full path including `https%3A%2F%2Fqa-circulation.openebooks.us%2FUSOEI%2Fworks%2FAxis%2520360%2520ID%2F0017370149`

playwright/support/mappings.ts
```diff
@@ -46,6 +46,22 @@ export const pages: { [name: string]: Pages } = {
     regex: /^\/app\/signin\/method\/clever/,
     waitForNavigation: true
   },
+  "book details": {
+    route: "/app/book",
+    regex: /^\/app\/book\//,
+    waitForNavigation: true
+  },
+  "borrowed book detail": {
```
samanthaandrews, line 54: Consider adding the title of the book to this string so that we know which book it maps to

playwright/features/circulationPage.feature
```diff
@@ -11,9 +11,9 @@ Feature: As a user I should only see the lanes that are appropirate for my age g
     And I click the "sign in button"
     And I am directed to the "open ebook circulation" page
     And the "open ebooks header" should be displayed
-    And the "high school header" should be displayed
+    And the "high school header for all access user" should be displayed
     And the "middle grades header" should be displayed
```
samanthaandrews, line 15: It seems like this "middle grades header" should also change to "middle grades header for all access user" to match the others

playwright/tests/assertions.spec.ts
```diff
@@ -10,18 +11,36 @@ Given(
...
+    return await expect(this.page.locator(element)).toBeVisible({
+      timeout: 50000
+    });
   }
 );
+
+Then(/^I verify "([^"]*)" is displayed$/, async function (this: CustomWorld) {
+  const borrowButton = await this.page.$("text=Borrow this book"); //use it to grab element's existance
+  if (borrowButton === null) {
+    console.log("we are here !!!");
```
samanthaandrews, line 36: Remove stray console.log

playwright/tests/assertions.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { Given, Then } from "@cucumber/cucumber";
 import { CustomWorld } from "../support/setup";
 import { elements } from "../support/mappings";
+import { expect } from "@playwright/test";
 
 Given(
```
samanthaandrews, line 6: I think this is missing an `expect`. Should be `expect(this.page.locator(element)).isEnabled();`

playwright/tests/assertions.spec.ts
```diff
@@ -10,18 +11,36 @@ Given(
...
     const element = elements[elementKey];
-    await this.page.locator(element).isVisible();
+    //return await this.page.locator(element).isVisible();
+    return await expect(this.page.locator(element)).toBeVisible({
+      timeout: 50000
+    });
   }
 );
+
+Then(/^I verify "([^"]*)" is displayed$/, async function (this: CustomWorld) {
```
samanthaandrews, line 33: Since this Then statement is specific to the borrow button, I don't think we should pass in a string here at all. It should just directly say "I verify the borrow this book button is displayed" so that we don't get confused and try to use this assertion outside the context of verifying the borrow a book button.

playwright/tests/assertions.spec.ts
samanthaandrews: We should verify that we're not missing any `expect` values throughout all of these tests, just to be sure

playwright/support/mappings.ts
```diff
@@ -91,6 +108,22 @@ export const elements = {
...
+  "search button": "//button[@type='submit']",
+  "borrow this book button": "//button[text()='Borrow this book']",
+  "return button": "//button[contains(text(),'Return')]",
+  "read button": "//div//a[text()='Read']",
+  "e-reader return button":
+    "//header//a[@aria-label='Return to Open eBooks (QA Server)']",
+  "e-reader table of contents button":
+    "//button[@aria-label='Table of Contents']",
+  "e-reader settings button": "//button[@aria-label='Settings']",
+  "e-reader toggle window button": "//header/div[1]/button[3]",
```
samanthaandrews, line 121: This selector seems pretty brittle since it relies on the position of the div and the buttons. I think this button includes the text "Toggle Fullscreen" that you should be able to use instead.

samanthaandrews (changes requested):
A few changes to make before we can merge this. it's looking really good!

PR #515, opened 2022-11-09
Monorepo!

kristojorg:
## Notes

- When checking this out, you will need to move `.env.local` into `/apps/oew/` manually since it is not tracked by git.
  - Similarly, when checking out a non-monorepo branch, you will need to move `.env.local` back into the root folder. It's easy to delete the left-over `/apps/oew` folder which has node_modules still, but don't forget if you delete this before moving `.env.local`, you'll lose it like I did : )
- We will need to update Vercel to deploy from the monorepo (I no longer have access here)

### How to manage changelogs

We have decided to only update changelogs for released apps and packages. We have yet to figure out how to enforce this in a PR though. For now, I simply deleted that part of the github workflow.

2. Use lerna, which apparently adds this feature
3. Use conventional commits and somehow make that do it automatically?
4. Use [changesets](https://github.com/changesets/changesets)
5. Use [inuit/auto](https://github.com/intuit/auto) to bump versions based on PR labels

### How to globally typecheck?

We can't run `tsc --noEmit` at the top level, because individual workspaces have their own tsconfigs. It needs to be run for each workspace. However, some workspaces don't even have tsconfigs (eg opds), so running `tsc --noEmit` on them doesn't make much sense. 

Packages without tsconfig files will be typechecked when building the consuming app/package. This has been tested in the opds package, where type errors are caught when building oew.

### How to decide module boundaries

This is not necessarily a problem that is new with a monorepo, as code in our apps should be modularized somewhat anyway, but when sharing code across apps it becomes more important to think about where module boundaries are drawn. Do we make a single "shared" module for shared app code? Do we make one for opds1 parsing? How do we modularize authentication code? 

We need to start considering how to define the responsibility of a module. 

I wrote a [document](https://docs.google.com/document/d/1Dmn5R5pWub6BlCvN3LhlGm6hadxjkfCEFNV2pFGt23c/edit?usp=sharing) to start this discussion


kristojorg:
@samanthaandrews can you check why the vercel build is failing for me? Thnx

Otherwise I think this is ready to go finally!

samanthaandrews:
@kristojorg Vercel is failing with the following error:

```
> Build error occurred
--
08:56:45.579 | Error: > Couldn't find a `pages` directory. Please create one under the project root
08:56:45.579 | at Object.findPagesDir (/vercel/path0/node_modules/next/dist/lib/find-pages-dir.js:86:19)
08:56:45.579 | at /vercel/path0/node_modules/next/dist/build/index.js:102:63
08:56:45.579 | at async Span.traceAsyncFn (/vercel/path0/node_modules/next/dist/trace/trace.js:79:20)
08:56:45.579 | at async Object.build [as default] (/vercel/path0/node_modules/next/dist/build/index.js:65:29)
08:56:45.605 | Error: Command "npm run vercel-build" exited with 1

```

kristojorg:
Yep ok we just need to tell it where to find the app. There should be a setting for the project root and we need to set it to apps/oew

samanthaandrews:
@kristojorg that makes sense. after the new sign in modal code is merged i'll make that change in Vercel.

packages/jest-presets/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "jest-presets",
+  "version": "0.0.0",
```
samanthaandrews, line 3: does this need a version number?

packages/jest-presets/base/jest-preset.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+      }
+    ]
+  },
+  // A list of paths to snapshot serializer modules Jest should use for snapshot testing
+  snapshotSerializers: ["@emotion/jest/serializer"],
+
+  // The glob patterns Jest uses to detect test files
+  testMatch: [
+    // "**/__tests__/**/*.[jt]s?(x)",
```
samanthaandrews, line 24: can we remove this commented out line?

package.json
```diff
@@ -1,128 +1,32 @@
...
-    "prettier": "2.5.1",
-    "ts-node": "^10.8.2",
-    "type-fest": "^2.12.0",
-    "typescript": "^4.8.3",
-    "url": "^0.11.0",
-    "urlpattern-polyfill": "^5.0.9"
+    "prettier": "2.5.1"
   },
+  "engines": {
+    "node": ">=16.0.0"
```
samanthaandrews, line 23: wow this is nice! is this a new entry for package.json files? how does it play with the `.nvmrc` file?

apps/oew/src/utils/env.ts
```diff
@@ -22,5 +22,4 @@ export const IS_SERVER = typeof window === "undefined";
 export const APP_VERSION = process.env.APP_VERSION;
 export const BUILD_ID = process.env.BUILD_ID;
 export const GIT_COMMIT_SHA = process.env.GIT_COMMIT_SHA;
-export const AXISNOW_DECRYPT = process.env.AXISNOW_DECRYPT;
```
samanthaandrews, line 25: nice

apps/oew/src/hooks/useComplaints/__tests__/actions.test.ts
```diff
@@ -0,0 +1,102 @@
+import * as actions from "hooks/useComplaints/actions";
```
samanthaandrews, line 1: Everything related to the "Report a Problem" modal was removed in a recent PR: https://github.com/NYPL-Simplified/circulation-patron-web/pull/518

These `useComplaints` hook files must have snuck in here, but we should remove them.

apps/oew/src/components/theme-ui/Head.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
```
samanthaandrews, line 1: I think this file was moved to the `chakra-ui` directory in a recent PR, so we can probably remove this one if the other one is where it needs to be

apps/oew/src/components/theme-ui/SigninOption.tsx
```diff
@@ -0,0 +1,133 @@
+/** @jsxImportSource theme-ui */
```
samanthaandrews, line 1: These files were recently converted to `chakra-ui` by @oliviawongnyc so we just need to make sure her changes are here in the new monorepo and these ones can be removed

apps/oew/cypress/tests/report-problem.ts
```diff
@@ -0,0 +1,47 @@
+import {
```
samanthaandrews, line 1: This can also be removed

apps/oew/cypress/fixtures/problemDoc.ts
```diff
@@ -0,0 +1,14 @@
+const problemDoc = `http://librarysimplified.org/terms/problem/cannot-issue-loan
```
samanthaandrews, line 1: can remove this too

apps/oew/next.config.mjs
```diff
@@ -88,11 +91,17 @@ const withMDX = nextMDX({
...
-module.exports = async (_phase, { _defaultConfig }) => {
-  /**
-   * @type {import('next').NextConfig}
-   */
+/**
+ * NOTES
+ *  - try using canary
+ *  - try with next.config.js instead of mjs.
+ *  - check what is generated in staging
+ */
```
samanthaandrews, line 99: Do we need to create a ticket to take action on these notes?

apps/oew/src/components/chakra-ui/MarketingLayoutPage.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
samanthaandrews, line 1: not sure why this file shows up as a new file instead of just being renamed but it looks ok!

.github/workflows/release-prod.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
           strategy: package
           prefix: v
+          root: "./apps/oew"
```
samanthaandrews, line 27: do we need to add this to the `release-qa.yml` file too?

samanthaandrews (changes requested):
Just a few small things that I noticed related to code that was recently merged into `staging` and a couple questions. Once merging the new sign in modal code we should also make sure that gets correctly pulled into this monorepo branch.

.github/workflows/release-prod.yml
kristojorg: I don't think we use this step in release-qa since we don't tag a git version for QA

apps/oew/next.config.mjs
kristojorg: Nope, my mistake, it was resolved

kristojorg:
Ok @samanthaandrews I think I addressed your comments, thanks for the detailed review!

src/components/chakra-ui/__tests__/AuthButton.test.tsx
```diff
@@ -1,74 +0,0 @@
-import { fixtures, render, waitFor, within } from "test-utils";
```
samanthaandrews, line 1: I think this file was accidentally deleted when it should be moved to `oew`

apps/oew/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 - Refactor: OE-648: upgrade to NextJS 12.3 and mode middleware into a single handler & add middlware unit tests & disable Cypress
 - Chore: OE-736: Upgrade to NextJS 13
+- Refactor: OE-739: Monorepo.
+- Breaking Change: OE-739: Make Axisnow decryptor a required package. Install will fail without access.
```
samanthaandrews, line 14: nit: move these up to "unreleased" section

samanthaandrews (approved):
Just those small things I noticed from the rebasing/merging but otherwise this looks great. This was a ton of work and I'm excited for our new monorepo endeavors!

samanthaandrews:
Vercel is building now, but failing with the following error:

```
Failed to compile.
-- 
./src/test-utils/index.tsx:9:34
Type error: Cannot find module 'jest-fetch-mock' or its corresponding type declarations.

```
Might just need to run npm install on `oew` again?

src/components/chakra-ui/__tests__/AuthButton.test.tsx
kristojorg: Oh shoot thanks. Git told me it had been deleted downstream I thought, but looks like you're right!

PR #517, opened 2022-11-10
OE-377: Update newrelic@next

Toxiapo:
This PR updates @newrelic@next to 0.3 for nextjs middleware support


next.config.js
```diff
@@ -166,34 +166,7 @@ module.exports = async (_phase, { _defaultConfig }) => {
         const staticAssetUrl =
           RELEASE_STAGE === "staging"
             ? "https://staging.openebooks.org"
-            : RELEASE_STAGE === "production"
-            ? "https://www.openebooks.org"
-            : "";
-        const NRconfig = new NewRelicSourceMapPlugin({
```
samanthaandrews, line 172: oh nice so in the latest version we don't need this plugin anymore?

Toxiapo: We still need it, but this code was a duplicate. It must be from one of the merge conflicts.

samanthaandrews: Ahh I see. Good catch.

kristojorg (approved):

PR #518, opened 2022-11-10
OE-707: Remove report a problem modal

samanthaandrews:
Since I'm working on adding the new chakra-ui modal I decided to make sure we had first removed all references to the ReportAProblem modal which Jen & Miguel decided to remove.


src/components/chakra-ui/__tests__/BookDetails.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { BookDetails } from "components/chakra-ui/BookDetail";
+import { BreadcrumbContext } from "components/context/BreadcrumbContext";
+import { fixtures, render } from "test-utils";
+
+test("MyBooks sets its breadcrumbs in BreadcrumbsContext", () => {
```
samanthaandrews, line 5: @oliviawongnyc I started to add this test just to confirm it would run. We should be unblocked from working on https://jira.nypl.org/browse/OE-725 after merging this.

Toxiapo (approved):

PR #519, opened 2022-11-10
OE-710: Signin page does not fill the whole width of the browser

Toxiapo:
Ticket: [OE-710](https://jira.nypl.org/browse/OE-710)

This PR fixes the SigninOption styles

PR #520, opened 2022-11-14
OE-684: convert signin to chakra

oliviawongnyc:
This PR resolves Jira ticket [OE-684](https://jira.nypl.org/browse/OE-684), converting SigninOption and its descendants from theme-ui to chakra-ui components.


src/components/chakra-ui/theme/components/button.ts
```diff
@@ -42,7 +49,8 @@ const ButtonStyle = {
   variants: {
     primary: primaryVariant,
     outline: outlineVariant,
-    menuItem: menuItemVariant
+    menuItem: menuItemVariant,
+    shadow: shadowVariant
```
samanthaandrews, line 53: I think we should rename this to something like "signInPicker" because it is very specific to the sign in selector component and likely won't be used elsewhere in the app. Other buttons in the app do have shadow which is why I think this might get confusing.

src/components/chakra-ui/theme/components/heading.ts
```diff
@@ -0,0 +1,26 @@
+const HeadingStyle = {
+  baseStyle: {},
+  variants: {
```
samanthaandrews, line 3: Historically we have used the `text.ts` file to define our heading styles. For consistency I think we should continue to use the styles defined there. This means headings will be implemented like this instead of using the Chakra Heading component:

```
<Text textStyle="h1" as="h2">
    Heading text
</Text>
```
I am wondering if it would make sense to convert to using the chakra `<Heading />` component in a future ticket though? I'm not sure which is better tbh. @Toxiapo may have thoughts.

src/components/chakra-ui/theme/components/link.ts
```diff
@@ -36,7 +54,8 @@ const LinkStyle = {
   variants: {
     marketingHeader: marketingHeaderVariant,
     appHeader: appHeaderVariant,
-    TOC: TOCVariant
+    TOC: TOCVariant,
+    navLink: navLinkVariant
```
samanthaandrews, line 58: maybe I'm missing it but I couldn't figure out where this is being used?

samanthaandrews (changes requested):

src/components/chakra-ui/theme/components/heading.ts
oliviawongnyc: Okay, I updated. I am overriding the fontWeight and fontSize, however, based on the styles that are currently live.

Toxiapo: Yeah, I would suggest using `text` if possible. It makes sense that we have to override some of the styles in there since this page was done before the design unification, so the base themes is a little different.

src/pages/app/signin/method/clever/callback.tsx
```diff
@@ -1,6 +1,6 @@
 /** @jsxImportSource theme-ui */
```
Toxiapo, line 1: We should be able to remove this now.

src/components/chakra-ui/SigninOption.tsx
```diff
@@ -51,94 +48,99 @@ const SigninOption: React.FC = () => {
...
+  textAlign: "center",
+  fontFamily: "body",
+  fontSize: "lg",
+  fontWeight: "normal",
+  lineHeight: "taller"
 };
 
-const logoHeader: SxProp["sx"] = {
-  display: "flex",
+const logoHeader = {
```
Toxiapo, line 142: For CSS-related objects, we can use `SystemStyleObject` as the type, but we don't need to add it here, just an FYI.

oliviawongnyc: Thank you!

oliviawongnyc: I went ahead an added. Unless there's a reason why we wouldn't add here.

Toxiapo: Yes it is good to add it!

Toxiapo (approved):
Looks good to me!

After @samanthaandrews approves this, can you pass this along with #519's bug to Ayan to QA?

src/components/chakra-ui/theme/components/button.ts
kristojorg: IMO if a variant is only used in one place, it doesn't need to live in the theme. Maybe we can simply inline these styles into a button?

src/components/chakra-ui/SigninOption.tsx
```diff
@@ -54,91 +48,86 @@ const SigninOption: React.FC = () => {
...
-            size="xl"
+            color="ui.blue.clever"
+            variant="signinButton"
+            size="lg"
             iconLeft={CleverBadge}
             sx={{
-              color: "ui.cleverBlue",
               "&:hover": {
-                color: "ui.cleverBlue"
+                color: "ui.blue.clever"
```
samanthaandrews, line 82: chakra has a convenience prop that we might be able to use here and in the First Book button `_hover={{ color: "ui.blue.clever" }}`

samanthaandrews (commented):
Just one small thing! Looks great.

src/components/chakra-ui/SigninOption.tsx
oliviawongnyc: I tried this, but it doesn't work. Not sure why!

src/components/chakra-ui/theme/components/button.ts
oliviawongnyc: Sounds good!

samanthaandrews (approved):
🚀

src/components/chakra-ui/theme/components/heading.ts
kristojorg: I think we didn't use the `Heading` component from Chakra initially because it was a bit confusing what it even did (as in how it was different from Text), so we just opted for Text

PR #522, opened 2022-11-16
OE-739: Extract models

Toxiapo:
This PR extracts the internal models into a package.

Note: Merge this after the monorepo PR to avoid a big PR review.


packages/models/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "./book": {
+      "types": "./book.ts",
+      "import": "./book.ts",
+      "require": "./book.ts"
+    },
+    "./collection": {
+      "types": "./collection.ts",
+      "import": "./collection.ts",
+      "require": "./collection.ts"
+    },
```
kristojorg, line 21: I think it's probably fine to just have a single `index.ts` file exposing all the exports. I don't know if we necessarily need to be able to import from other paths like `models/types`. Probably `import {Book, Collection, bookIsBorrowable} from "models"` is fine

Toxiapo: Yeah, that sounds good! I updated to just have a single export in the package.json, should we leave the types as is? We do need that on a lot of files in the main repo

PR #523, opened 2022-11-17
Fix breadcrumb

Toxiapo:
Ayan discovered the breadcrumb on the staging branch and the production branch look different, this is because we updated chakraui to v2, and they added a default flexbox style to the wrapper element.

This PR
1. Added the Breadcrumb component to the chakra theme file to override some of the styles
2. Made the header logo smaller to account for the screen size that's smaller than 320px (Galaxy fold)


src/components/chakra-ui/theme/components/breadcrumb.ts
```diff
@@ -0,0 +1,21 @@
+const BreadcrumbStyle = {
+  baseStyle: {
+    container: {
+      position: "sticky",
+      py: "2",
+      paddingInline: { base: "4.5rem", lg: "7.5rem" },
+      bg: "ui.black",
+      color: "ui.white",
+      fontSize: "sm",
+      zIndex: "overlay",
```
Toxiapo, line 10: I lowered the zIndex from `header` to `overlay` just so that it doesn't block the drawdown menu when the breadcrumb becomes too large. I am not sure if there's a better way to deal with these stacking issues, or if we should just wait til we have the design for the breadcrumb rework on mobile.

samanthaandrews: I think this is a good solution 👍

samanthaandrews:
I did manual testing in my review and the issue is fixed for me

samanthaandrews:
I just merged my accessibility fix PR, so there are conflicts. We should also check that the "Skip to content" link is still visible with these changes. It should display over the header/breadcrumbs if a user navigates via keyboard.

Toxiapo:
@samanthaandrews Skip content looks good!

PR #527, opened 2022-11-21
OE-582: Make My Books more inviting when no loans are active

Toxiapo:


src/components/theme-ui/MyBooks.tsx
```diff
@@ -75,21 +77,35 @@ const LoansContent: React.FC<{ books: OpenEBook[] }> = ({ books }) => {
...
+            <ChakraLink
+              as={Link}
+              href={CATALOG_ROOT_PATH}
+              fontSize={["lg", "xl", "xl"]}
+              fontWeight="light"
+              color="ui.black"
+            >
+              All books
+            </ChakraLink>{" "}
+            to find something you like
```
samanthaandrews, line 104: I'm thinking we should have a `.` at the end of this sentence

src/components/theme-ui/MyBooks.tsx
```diff
@@ -75,21 +77,35 @@ const LoansContent: React.FC<{ books: OpenEBook[] }> = ({ books }) => {
...
+        <Box m="auto" maxWidth="800px">
+          <Text
+            textStyle="h6"
+            borderRadius="30px"
+            bgColor="rgba(10, 153, 160, 0.24)"
+            boxShadow="0px 4px 4px rgba(0, 0, 0, 0.25)"
+            mx={3}
+            py={[4, 4, 5]}
+            px={[8, 8, 10]}
+            fontWeight="light"
```
samanthaandrews, line 91: Not something we need to address in this PR, but if we do end up re-implementing a similar callout text box thing we should make this into its own component.

package.json
```diff
@@ -40,7 +40,7 @@
     "@next/mdx": "^12.3.1",
     "@nypl/dgx-svg-icons": "^0.3.12",
     "@nypl/epub-to-webpub": "^1.1.0",
-    "@nypl/web-reader": "^3.2.5",
+    "@nypl/web-reader": "^3.2.4",
```
samanthaandrews, line 43: Did we mean to downgrade this?

Toxiapo: nope, not sure how it gets downgraded. updating!

src/components/theme-ui/MyBooks.tsx
Toxiapo: Yeah good call.

src/components/theme-ui/MyBooks.tsx
Toxiapo: Just confirmed with Jen that this should be fine!

samanthaandrews (approved):

kristojorg (changes requested):
I think we should remove the phrase "Take a look at the selection below or " for now, since we don't know when that change is going to be ready and we don't want this to go to production this way. We can add the phrase back in in the next PR

src/components/theme-ui/PageTitle.tsx
```diff
@@ -4,23 +4,22 @@ import { H1 } from "components/theme-ui/Text";
...
       sx={{
-        px: [3, 5],
+        px: [3, 3, 5],
         mt: 4,
         display: "flex",
         flexDirection: ["column", "column", "row"],
         justifyContent: ["flex-start", "space-between"],
         alignItems: ["stretch", "stretch", "flex-end"],
-        mb: [0, 0, 4]
+        mb: [0, 0, "26px"]
```
kristojorg, line 19: Is there a reason we need a non-standard margin value instead of one based on the spacing scale?

Toxiapo: Yeah, the theme-ui and chakra-ui spacing scales are different, and since we are eventually going to delete this component, I wanted to keep this as simple as possible for doing the conversion in the future.

Toxiapo: I think this is my plan going forward with these components, but let me know if you have concerns with this method.

kristojorg: Okay I see, sounds good to me then

kristojorg (approved):

PR #528, opened 2022-11-22
OE-666: New sign out modal

samanthaandrews:
This PR builds the new sign out modal using Chakra and removes the old theme-ui one. Passed visual QA!


src/components/chakra-ui/header/Header.tsx
```diff
@@ -24,7 +24,7 @@ export const Header: React.FC<HeaderProps> = ({
   return (
     <Flex
       position="sticky"
-      zIndex="header"
+      zIndex="sticky"
```
Toxiapo, line 27: I think this value should be higher than what's on the hamburger dropdown, which is `modal`, so then when the dropdown opens, it opens from the background instead of the header foreground

src/components/chakra-ui/theme/components/button.ts
```diff
@@ -149,6 +150,12 @@ function outlineVariant() {
   };
 }
 
+function ghostVariant() {
+  return {
+    backgroundColor: "none",
```
Toxiapo, line 155: Do we need this if it's just one style?

src/components/chakra-ui/theme/foundations/color.ts
```diff
@@ -49,9 +49,7 @@ const colors: Colors = {
       xlight: "#F5F5F5",
       xxlight: "#FAFAFA"
     },
-    red: {
-      firstbook: "#ED1C24"
-    },
+    red: "#ED1C24",
```
Toxiapo, line 52: I wonder if we should rename these brand colors to "ui.firstbook.red / ui.firstbook.blue / ui.clever.blue / ui.clever.black"? I am not sure how many reds will Jen use so this is just a thought.

src/utils/auth.ts
```diff
@@ -170,13 +170,11 @@ export async function verifyToken(
...
   localStorage.clear();
 
   const res = await fetch(SIGNOUT_PATH, { method: "POST" });
   if (res.ok) {
-    router.push(new URL("/", window.location.origin), undefined, {
-      shallow: false
-    });
+    window.location.href = new URL("/", window.location.origin).toString();
   }
 }
```
Toxiapo, line 180: has the Clever signout bug been fixed with next13?

Toxiapo (changes requested):
This looks great!

I just have one question regarding accessibility. When I open the Sign Out popup with my keyboard, the focus was on the "x" text, should it be on the "Sign Out" button?

src/components/chakra-ui/theme/components/button.ts
kristojorg: I think this depends on whether the designs will have ghost buttons from now own. I'd love to get a set of all the button variants, sizes and colors from jen so we can just put them in our theme and not need to wonder what should be a variant vs a one-off

src/utils/auth.ts
kristojorg: In general for internal navigation I think we should use `next/router`

src/components/theme-ui/AppLayout.tsx
```diff
@@ -21,7 +21,7 @@ const AppLayout: React.FC<
     >
       <GlobalStyles />
       <ChakraProvider theme={theme}>
-        <SkipNavLink zIndex="skipNav">Skip to content</SkipNavLink>
+        <SkipNavLink zIndex="skipLink">Skip to content</SkipNavLink>
```
kristojorg, line 24: I don't know that it's related, but I never see this appear anymore

src/utils/auth.ts
kristojorg: But I do remember this being a tricky problem. Maybe things have progressed and there is a better way to sign people out than requesting a url and waiting for it to resolve?


Also, if we continue doing it this way we should consider adding a progress indicator to the spinner. It takes some time before the request to /signout resolves and looks like nothing is happening. We could move this code from `utils/auth` directly into the SignOut button because I don't think it is used anywhere else

src/components/chakra-ui/theme/foundations/color.ts
samanthaandrews: I also wonder. This first book red wasn't actually being used anywhere except the first book logo SVG, and in that case it is not passed through from the theme anyway. I noticed Jen used the same color for this modal "Sign out" action though, so I figured it made more sense to just name it "red" instead of mentioning First Book at all.

src/components/theme-ui/AppLayout.tsx
samanthaandrews: It's there! Navigate to the site and hit "tab" and it should pop up. Lmk if you don't see it though bc that is a bug that we should fix.
Screen Shot 2022-11-28 at 2 45 33 PM

src/components/chakra-ui/theme/components/button.ts
samanthaandrews: I'm going to bring this up to Jen and ask her to create these components when she has some down time. For now, I removed this from the variants list and just used an inline style for these modal buttons.

src/components/chakra-ui/theme/foundations/color.ts
samanthaandrews: I did change the First Book and Clever names to match your suggestion, @Toxiapo 👍

src/components/chakra-ui/header/MenuNav.tsx
```diff
@@ -42,7 +42,7 @@ export const MenuNav: React.FC<React.PropsWithChildren<Omit<MenuNavProps, "aria-
         onClose={onClose}
         finalFocusRef={btnRef}
       >
-        <DrawerOverlay />
+        <DrawerOverlay zIndex="dropdown"/>
```
samanthaandrews, line 45: @Toxiapo i updated the drawer overlay to use "dropdown" so that all of these layers display correctly

src/components/chakra-ui/header/Header.tsx
samanthaandrews: You are correct, but I edited how the hamburger dropdown displays, so now it has a `zIndex="dropdown"`. I tagged you in the spot where I made that change.

src/components/chakra-ui/AuthButton.tsx
```diff
@@ -53,3 +111,19 @@ export const AuthButton = forwardRef<HTMLButtonElement, AuthButtonProps>(
...
+ * a weird bug in Clever signout where `window.location.href` wouldn't
+ * clear the cookie resulting in a login loop.
+ */
+ export async function signOut(router: NextRouter) {
+  localStorage.clear();
+  Cookies.remove(AUTH_COOKIE_NAME);
+
+  router.push(new URL("/", window.location.origin), undefined, {
+    shallow: false
+  });
```
samanthaandrews, line 128: @kristojorg what do you think about this? instead of using the `/api/auth/signout` path, I just remove the cookie and then push users back to the homepage using the router. Then there's also no need for a loading spinner.

src/utils/auth.ts
samanthaandrews: I updated this to not use an API call at all and just remove the cookie. Lmk what you think.

src/components/chakra-ui/theme/foundations/color.ts
```diff
@@ -13,7 +13,7 @@ const colors: Colors = {
     black: "#000000",
     semiBlack: "#222222",
     bodyText: "#424242",
-    error: "#97272C",
+    danger: "#EC445A",
```
samanthaandrews, line 16: I spoke with @jotten and agreed we should replace any red/error color with one value, so that's what you see here. It's now called "danger".

src/components/theme-ui/AppLayout.tsx
kristojorg: Huh, strange. It doesn't show up on safari, but now that I checked with chrome I do see it

src/components/chakra-ui/AuthButton.tsx
kristojorg: Ok tested and it works on safari and chrome at least. Should be fine. Wish I had a decision log so I could look back at why I originally converted from Cookie.remove to the api/auth/signout option

kristojorg (approved):
Looks good!

PR #530, opened 2022-12-01
OE-737: Add Reader Plugin Shell

kristojorg:
This PR adds a framework for our Reader plugin and iOS implementation. It shows:

- Opening a native view via a plugin
- Populating the native view with data from the JS side
- Adding/removing elements to the list on the native side. This triggers events on the JS side
- The JS side updates its state and pushes updates back to native

In the end, state is stored in JS, and stays in sync between iOS and WebView. 

The flow is as follows:

1. JS side has bookmarks state stored locally, but this can also be on a server without issue.
2. JS side calls `Reader.openBook({ bookmarks })`, passing the initial bookmarks through. Eventually, it will also pass the location of the downloaded book and decryption parameters and a last reading location.
3. Native side receives the `openBook` call in `ReaderPlugin`. It can then perform any setup it needs, such as initializing a native `Reader` module. In this case it initializes the `BookmarkModule`. 
4. In the `openBook` call, the native side pushes a new `ViewController` onto the stack.
5. When a user clicks a native button like the "Add bookmark" button, the `BookmarkDelegate` can call the corresponding method on the plugin, which in turn uses `notifyListeners` to trigger an event on the JS side. 
6. The JS code listens for events, and in this case updates its local state. It first adds a bookmark in the loading state, and then after 2 seconds updates it to be loaded (simulating a network call).
7. When the bookmarks state is updated, an effect is triggered which calls `Reader.updateBookmarks` to pass the new set of bookmarks to the native side. This happens twice: first when the loading bookmark is added, then when its state changes to loaded.
8. When the native plugin receives new state, it parses bookmark models out of the JSON passed and passes them to the `BookmarkModel`

NOTE: This also makes the AxisNow decryptor an optional dependency again so that mobile devs without access can successfully run npm install


apps/oeh/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ### UNRELEASED CHANGES
 
-### v0.0.1
+### v3.0.0
```
kristojorg, line 5: I think we will start on the next major version from the current OE app

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,13 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		2CC920C329360EA300AFB5B7 /* Increment.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2CC920C229360EA300AFB5B7 /* Increment.swift */; };
```
kristojorg, line 10: @ettore how does review work for XCode files like this? Do you just assume they're fine and ignore them, since they're auto-generated?

apps/oeh/ios/App/App/Increment/IncrementPlugin.swift
```diff
@@ -0,0 +1,94 @@
+import Foundation
```
kristojorg, line 1: The increment plugin is simpler and doesn't differentiate between pure-native code and capacitor-aware code like the Reader plugin does. But it is maybe helpful to see. I'll leave it here for now but we can remove at any time.

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
ettore: I do look at them for anything suspicious... Xcode sometimes messes up, and git-merge also gets confused by this file sometimes. I'll keep an eye out.

apps/oeh/package.json
```diff
@@ -1,14 +1,14 @@
...
+  "version": "3.0.0",
+  "description": "Open eBooks for iOS and Android",
   "main": "index.js",
   "keywords": [
     "capacitor",
     "mobile"
   ],
   "scripts": {
-    "dev": "vite",
+    "dev": "vite --host",
```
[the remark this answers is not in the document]
ettore: would it be ok to write this kind of info somewhere more permanent / easy to refer to? I for one know that I will forget this instantly.

apps/oeh/README.md
```diff
@@ -1,55 +1,181 @@
...
+## 4: Run the native project
+
+You can run the native projects using the capacitor command line, or the corresponding native editor (XCode and Android Studio).
+
+### Command line
+
+Android: `npx cap run android`
+iOS: `npx cap run ios`
+
+It ensure the native projects are set up right, install any missing podfiles, and
```
ettore, line 81: ... and...? :)

apps/oeh/README.md
```diff
@@ -1,55 +1,181 @@
...
+⚡️  Error: Could not connect to the server.
+```
+
+We are not entirely sure why this error occurs, but it appears that some machines don't allow you to set the localhost ip address directly. Instead, change `server.url` to `http://localhost:8100` in `capacitor.config.ts` then run `npx cap sync` and finally `npx cap run ios` to run the app. Using `localhost` should work.
+
+# Configuration
 
 There are multiple places configuration happens in this app. I will attempt to clear some of this up:
 
 - `vite.config.ts`: Configures the build and bundling of our Vite-powered web app code.
```
ettore, line 173: maybe this is off topic for this README, but what is Vite ?

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
ettore: btw can we manually modify this file? I wasn't sure if this is autogenerated somehow or if it's something that we will maintain manually. It seems like it's the latter but I wanted to make sure.

apps/oeh/ios/App/App/BookmarkModule/BookmarkModule.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+import UIKit
+
+/**
+ 
+ This file is the BookmarkModule, which is entirely native and should not have any knowledge of capacitor.
```
ettore, line 6: is this just placeholder / POC code?

apps/oeh/ios/App/Podfile
```diff
@@ -12,6 +12,8 @@ install! 'cocoapods', :disable_input_output_paths => true
 def capacitor_pods
   pod 'Capacitor', :path => '../../../../node_modules/@capacitor/ios'
   pod 'CapacitorCordova', :path => '../../../../node_modules/@capacitor/ios'
+  pod 'CapacitorFilesystem', :path => '../../../../node_modules/@capacitor/filesystem'
```
ettore, line 15: is it npm that manages the Podfile? since we don't explicitly invoke any `pod` commands....

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
```diff
@@ -0,0 +1,86 @@
...
+        self.bridge?.viewController?.present(alert, animated: true, completion: nil)
+        
+        call.resolve()
+    }
+}
+
+extension ReaderPlugin: BookmarkDelegate {
+
+    func createBookmark(bookmark: Bookmark) {
+        notifyListeners("CREATE_BOOKMARK", data: [
```
ettore, line 76: who implements `notifyListeners`? and/or are these listeners listening on the web side?

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
```diff
@@ -0,0 +1,86 @@
...
+                  let title = element["title"] as? String,
+                  let isLoading = element["isLoading"] as? Bool else { return nil }
+            return Bookmark(id: id, title: title, isLoading: isLoading)
+        }
+        return models
+    }
+    
+    // perform the first time setups
+    @objc func openBook(_ call: CAPPluginCall) {
+
```
ettore, line 42: differently from previous POC code I had seen (I think in nypl-cap), here we do not save the `call` as a instance variable and we do not set it up for "keep alive". Is this because in this POC we are just listening for bookmark creation/deletion, so we don't need a long-lived call?

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
```diff
@@ -0,0 +1,86 @@
...
+    // perform the first time setups
+    @objc func openBook(_ call: CAPPluginCall) {
+        
+        let bookmarks = getBookmarksFromCall(call)
+        
+        DispatchQueue.main.async {
+            self.bridge?.viewController?.present(self.module.viewController, animated: true, completion: nil)
+            self.module.receiveModels(models: bookmarks)
+        }
+    }
```
ettore, line 49: shouldn't we call `call.resolve()` at some point before exiting?

apps/oeh/src/pages/Increment.tsx
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    return () => {
+      listener.remove();
+    };
+  }, []);
+
+  React.useEffect(() => {
+    async function update() {
+      const result = await Increment.updateCount({ count });
```
ettore, line 30: is this the call to the native code to update the native view with the new count?

apps/oeh/src/plugins/increment/definitions.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+   * Listens for increment events to update the count.
+   */
+  addListener(
+    eventName: "INCREMENT",
+    listenerFunc: () => void
+  ): Promise<PluginListenerHandle> & PluginListenerHandle;
+
+  updateCount({ count }: { count: number }): Promise<{ count: number }>;
```
ettore, line 28: this is just a declaration, correct? not a function call. Sorry for the illiterate TS question.

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
kristojorg: I didn't manually add this, so there is at least some auto-generation, but if it's something that you normally manually modify then I assume that is also okay. TBH I'm not entirely sure though

apps/oeh/package.json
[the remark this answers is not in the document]
kristojorg: Yea I tried to add as much as I could to the README in the `oeh` app. I will add a section about how to run it on a physical device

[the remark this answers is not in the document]
kristojorg: Oh actually I already added it haha. It's the "running on a real device" section

apps/oeh/README.md
```diff
@@ -1,55 +1,181 @@
...
+```
+✖ Creating capacitor.config.json in android/app/src/main/assets - failed!
+✖ copy android - failed!
+[error] Error: ENOENT: no such file or directory, open
+        '/Users/joon-seokkim/Projects/circulation-patron-web/apps/oeh/android/app/src/main/assets/capacitor.config.json'
+```
+
+The `capacitor.config.json` file should be automatically generated during `npx cap sync`, but the command seems to fail if the file doesn't exist already. It is not committed to git, and therefore you might see this error the first time you clone the repo. To fix it, we use `npx cap copy` instead:
+
+1. Stop the running development server.
```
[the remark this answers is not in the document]
kristojorg: When you run `npm run dev` it will start a process in that terminal that doesn't exit. You can then use "command + c" to exit like a normal terminal process. Killing the terminal or closing the terminal window/tab should do the same.

apps/oeh/README.md
kristojorg: Good question. I will add explanation at the top

apps/oeh/ios/App/Podfile
kristojorg: Good question. My understanding is that you install a capacitor plugin using `npm install <plugin>`. Running `npx cap sync` will then look for native dependencies in those plugins, add them to the podfile and then run `pod install` (or whatever the pod command is). I will try to make this more clear.

kristojorg: But yea, I've never had to manage the podfile or run pod commands so far. I presume that if you wanted to install some other pod you could do that manually though.

kristojorg: Added as much as I know to the README

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
kristojorg: `notifyListeners` is implemented by `Capacitor`. It generates an event on the web side which we listen for: 
```ts
    const createListener = Reader.addListener("CREATE_BOOKMARK", action => {
       const bkm: Bookmark = {
         id: action.id,
         title: action.title,
         isLoading: true
       };
       addBookmark(bkm);
     });
```

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
kristojorg: Yea this was something I discussed with J as well. It appears there are two ways to communicate:

1. Save the initial call and resolve on that call multiple times
2. Trigger a JS event.

I chose number 2 as it seems cleaner and is a more common method of communicating with external contexts in the web (it's also how a website communicates with iframes, workers, other tabs, etc). I don't really understand the details behind how saving a call works so I thought better to avoid it.

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
kristojorg: That's a good question. Not sure if it is necessary or not. I guess the promise on the web side won't resolve until that happens. We aren't currently waiting for it to resolve, but I will update this to resolve.

apps/oeh/src/pages/Increment.tsx
kristojorg: Yep exactly. This `useEffect` works by running the function any time the count changes, thus calling `updateCount` in the native plugin

apps/oeh/src/plugins/increment/definitions.ts
kristojorg: Yep, and no worries! Just defining the plugin interface

apps/oeh/ios/App/App/BookmarkModule/BookmarkModule.swift
kristojorg: It's just a POC but this can also be the start of the real reader plugin

kristojorg: Yea just a POC showing the plugin interacting with a native module. The real bookmark module and reader module (maybe they're the same? idk) will be installed via SPM from the native reader module.

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
ettore: yeah, normally we do manually modify it (through Xcode UI). Perhaps in this case it was autogenerated, but since we are making changes to it (like team ID, etc), it's safe to assume we won't autogenerate it again and we'll just take ownership of it.

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
ettore: I don't know the difference myself. I thought we had to set keep alive to true if we keep the native view open after we `resolve`. But maybe we don't? Perhaps @jkim-git can shed some light.

jkim-git: Yeah that is one of the ways to send multiple callbacks to hybrid from native (first example in Kristo's bullet points). Using the second approach as it is implemented on this PoC, we don't need to save the call in memory as we are calling `notifyListeners` function that doesn't interact with the plugin call.

jkim-git: Some docs on the subject:
First approach: https://capacitorjs.com/docs/v3/core-apis/saving-calls
Second approach: https://capacitorjs.com/docs/plugins/ios#plugin-events

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
kristojorg: Yep it looks like that's the case. I made some manual changes to it and it seemed fine

kristojorg:
If y'all don't have any problems with this code, I would like to merge it into our staging branch. It is not in production yet obviously, so it's sort of meaningless, but I think it will make our dev simpler to be branching off staging from now on.

cc @ettore @jkim-git @samanthaandrews

apps/oeh/capacitor.config.ts
```diff
@@ -1,18 +1,27 @@
...
+   * This tells Capacitor to look for the web bundle (HTML, CSS, JS)
+   * at the given URL. Running `npm run dev` will run a local dev server
+   * at that url. This is useful for development, but in production, the web
+   * bundle should be statically built (`npm run build`) and copied into the
+   * native projects (`npx cap sync`). Setting this to localhost in production
+   * will result in a blank screen.
+   */
   server: {
-    url: "http://127.0.0.1:8100"
+    url: isProd ? undefined : "http://127.0.0.1:8100"
```
ettore, line 24: very nice: this was a parameter that was a little annoying to change depending what you built it for, so I'm glad we don't have to do that anymore when switching between prod and dev.

apps/oeh/ios/App/App/Info.plist
```diff
@@ -2,10 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<false/>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-        <string>oeh</string>
+	<string>oeh</string>
```
ettore, line 10: usually this is set to $(PRODUCT_NAME) but I am going to do a review of the project later, so no worries. I suppose the only thing I wanted to know if this was done for a particular reason.

ettore (approved):
let's merge it!

samanthaandrews (approved):
🚀

jkim-git (approved):

apps/oeh/capacitor.config.ts
```diff
@@ -1,18 +1,27 @@
 import { CapacitorConfig } from "@capacitor/cli";
 
+const isProd = process.env.NODE_ENV === "production";
```
ettore, line 3: somehow this doesn't seem to work for me when I build for prod with `nom run build` + `npx cap copy`. The `isProd ? ` check down below always return false when I launch the app from Xcode, and it attempts to load from localhost.

apps/oeh/ios/App/App/Info.plist
kristojorg: No I don't actually know how this was set to be honest, but yea I think we should change it. I assume this is why the app has "oeh" in the menu bar when I launched it as a mac app?

apps/oeh/capacitor.config.ts
kristojorg: Yea good catch. I have changed this around a little bit to use an env var we set ourselves called `BUNDLE_TYPE` which can either be `"static" | "server"`. I created utility scripts in `package.json` to make switching between the two easier and updated the README with new info. You can now use `npm run sync:live` or `npm run sync:static` and `npm run ios:live` or `npm run ios:static` depending on your preferred method.

PR #532, opened 2022-12-02
OE-748: Add poc

samanthaandrews:
This PR:
- [OE-748](https://jira.nypl.org/browse/OE-748) : Adds the proof of concept code from @kristojorg [nypl-cap](https://github.com/kristojorg/nypl-cap/) project
- [OE-749](https://jira.nypl.org/browse/OE-749): Ensures basic first book auth works to authenticate users
- [OE-750](https://jira.nypl.org/browse/OE-750): Ensure browse, borrow, and return books works as expected with basic auth tokens

Open issues to address in future tickets:
- Navigating to `/browse/book/:bookId` from the list of books is working but reloading the page throws an error saying the page doesn't exist
- Fetching code is duplicated in both `oew` and `oeh`. This should eventually be pulled out into a package.
- Pull out oeh and oew constants into a package, ensuring the CM_URL env var is set correctly based on environment


apps/oew/next.config.mjs
```diff
@@ -152,6 +152,13 @@ const nextConfig = async (_phase, { _defaultConfig }) => {
         config.plugins.push(
           new webpack.IgnorePlugin({ resourceRegExp: /jsdom$/ })
         );
+      config.plugins.push(
+        new webpack.IgnorePlugin({
+          resourceRegExp: /canvas/,
```
samanthaandrews, line 157: Jsdom was throwing errors in `oew` saying that it couldn't find the `canvas` module. If i remember correctly I think it was originating coming from the `models` package. Lmk if you think I should address the issue elsewhere.

packages/constants/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "constants",
```
samanthaandrews, line 2: i added a constants package because i just wanted to experiment with making a package. currently it is only being used in `oeh` and we need to make a ticket to move over everything from `oew`. the tricky part will be incorporating the different CM_URL env vars into the constants package.

kristojorg: Yea I wonder if it makes sense to have a package called `circulation-manager-service` or something that has all the code for dealing with the circulation manager. This would eventually be swappable for `pjr-backend-service` instead when we switch to that.

kristojorg: It could export a class that you pass ENV vars to to set up a bunch of utilities:

```ts
const CM = new CircManagerService({
  baseUrl: ...,
  librarySlug: ...,
}

// later
const patronEndpoint = CM.patronEndpoint
const meResp = CM.fetchMe()
const book = CM.fetchBook(id) // or url
```

Something like this, to start to extract out the particularities of the CM from the codebases into a defined module with clear boundaries.

apps/oeh/README.md
```diff
@@ -14,7 +14,7 @@ and comes with a very minimal shell for building an app.
 There are two pieces to this app that are run/built independently:
 
 1. The hybrid (web-based) code. This needs to be transpiled from TS to JS and bundled into a small set of static files and made available to the native projects.
-1. The native code. This needs to be compiled and then run in a simulator or on device.
+2. The native code. This needs to be compiled and then run in a simulator or on device.
```
kristojorg, line 17: FYI you can just use 1 1 1 1 1 in a markdown numbered list and it will auto-format properly. so I just use 1s so that if I change/reorder them I don't have to change all the numbers

apps/oew/next.config.mjs
kristojorg: I also got that and I think I just needed to re-install `jsdom` in oew. I may have incorrectly uninstalled it previously

packages/constants/package.json
kristojorg: I started a PR for this here and made a ticket I think we should address in the next sprint probably. This constants package is fine for now though: https://github.com/NYPL-Simplified/circulation-patron-web/pull/533

kristojorg (approved):
LGTM

apps/oeh/vite.config.ts
```diff
@@ -4,6 +4,14 @@ import basicSsl from "@vitejs/plugin-basic-ssl";
 
 export default defineConfig({
   root: "./src",
+  resolve: {
+    alias: {
+      // required to make dompurify and jsdom work correctly
```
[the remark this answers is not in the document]
samanthaandrews: This was the issue I was seeing: https://github.com/vitejs/vite/issues/7555

apps/oeh/README.md
samanthaandrews: ohh nice!

apps/oew/next.config.mjs
samanthaandrews: Ah ok let me try this!

samanthaandrews: Weird. I got the same error even after installing jsdom explicitly.

kristojorg: Huh strange. Did you install in the oew repo with `npm i jsdom -w oew`?

samanthaandrews: yup, that's what i did
Screen Shot 2022-12-05 at 3 27 04 PM

packages/constants/openebooks.ts
```diff
@@ -0,0 +1,16 @@
+export const AUTH_DOC_URL =
+    'https://qa-circulation.openebooks.us/USOEI/authentication_document'
```
Toxiapo, line 2: This is a placeholder for now, right? Cuz it's just the QA link

samanthaandrews: Yup just a placeholder for now

PR #533, opened 2022-12-05
Circ Manager Package

kristojorg:
This PR starts a Circulation Manager package to modularize code that deals with the CM. This includes handling the auth document, setting up URLs to fetch, fetching resources, and turning CM and fetch errors into errors that OEW and OEH will understand


samanthaandrews:
What do you think about closing this PR for now since we are re-evaluating our strategy?

kristojorg:
Yep good idea, closing now

PR #535, opened 2022-12-06
OE-737 Docs and gitignore fixes after building iOS

ettore:
I added some fixes to the documentation based on my experience building the project. 

I also noticed that when building iOS a few files were generated, and those should probably be ignored by git.

I set this to be merged on the main PR for the hybrid app: I can also push the checkins directly to `OE-737/communication-poc` if that's better, but I didn't want to do that right away in case you disagree with these edits.


apps/oeh/README.md
```diff
@@ -38,15 +38,41 @@ Details on how to run the app are below, but the following resources may also be
 - [Capacitor Docs: Development Workflow](https://capacitorjs.com/docs/basics/workflow)
 - [Capacitor Docs: Environment Setup](https://capacitorjs.com/docs/getting-started/environment-setup)
 
-## 1 & 2: Build the Javascript and serve it on a dev server
+## 1: Build the Javascript
+
+From root of `ereading-clients` project:
+1. `nvm install` to install the version of `node` officially used by this project.
+1. `` nvm use `cat .nvmrc` `` to use the correct version of `node`.
```
kristojorg, line 45: This can just be `nvm use`

apps/oeh/README.md
```diff
@@ -38,15 +38,41 @@ Details on how to run the app are below, but the following resources may also be
...
+1. `npm install` to install root JS dependencies.
+1. `cd apps/oeh`
+1. `npm install` to install all JS dependencies for the hybrid apps.
```
kristojorg, line 48: This is not necessary. Running `npm install` in the root will install all the (npm) deps for all of the workspaces in the monorepo

apps/oeh/README.md
```diff
@@ -38,15 +38,41 @@ Details on how to run the app are below, but the following resources may also be
 - [Capacitor Docs: Development Workflow](https://capacitorjs.com/docs/basics/workflow)
 - [Capacitor Docs: Environment Setup](https://capacitorjs.com/docs/getting-started/environment-setup)
 
-## 1 & 2: Build the Javascript and serve it on a dev server
+## 1: Build the Javascript
```
kristojorg, line 41: Let's rename this "set up your environment". Building the javascript is part of `npm run dev`

apps/oeh/README.md
ettore: I can remove it though

apps/oeh/.gitignore
```diff
@@ -5,3 +5,7 @@ node_modules/
 .DS_Store
 .sourcemaps
 dist/
+Installing
+ios/App/App.xcarchive
+ios/App/archive.plist
```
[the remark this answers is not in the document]
ettore: makes sense

apps/oeh/README.md
kristojorg: Yea I think that was just a bug in that npm version, according to npm it should be the same actually no matter where you run it from within a monorepo

PR #537, opened 2022-12-08
OE-767/signin-flow (and other routes)

kristojorg:
This PR adds the sign in flow and a bunch of route shells that are present in the wireframes. I made a few changes:

- There is now an /app route that holds the tabs. The sign in routes don't have tabs. I couldn't figure out how to make the sign in routes not have tabs without nesting the tabs in their own path.
- Each tab has to have its own independent set of routes. [see ionic docs](https://ionicframework.com/docs/react/navigation#working-with-tabs). This means we need a /browse/book/:bookUrl and a /my-books/:bookUrl. They're independent stacks basically. We need to keep this in mind.


samanthaandrews (approved):

apps/oeh/src/pages/sign-in/PrivacyPolicy.tsx
```diff
@@ -0,0 +1,7 @@
+import Page from "../../components/Page";
```
samanthaandrews, line 1: I wonder if this is supposed to link to https://www.openebooks.org/privacy

apps/oeh/src/pages/sign-in/UserAgreement.tsx
```diff
@@ -0,0 +1,192 @@
+import {
```
samanthaandrews, line 1: We do have this page on the web https://www.openebooks.org/app-user-agreement. So we might be able to share the content in this view or link to it, but I know that this ionic page has the accept/reject functionality which seems important.

apps/oeh/src/pages/sign-in/PrivacyPolicy.tsx
kristojorg: Good point. I'll ask

PR #538, opened 2022-12-08
Set up iOS module as SPM package

ettore:
Moves the BookmarkModule code into a local SPM package. Eventually the Swift package will move to a separate remote repo. This is just a proof of concept that Capacitor can work with SPM packages instead of cocoa pods for native code.


jkim-git:
> Eventually the Swift package will move to a separate remote repo

Would this be in conflict with the monorepo structure that's been setup?

jkim-git (approved):

ettore:
> > Eventually the Swift package will move to a separate remote repo
> 
> Would this be in conflict with the monorepo structure that's been setup?

I don't think so, unless we want to have the native code in this repo. I was under the impression that the native reader should be its own repo because it needs to be used by different apps (project-R and Open eBooks) ?

apps/oeh/android/app/capacitor.build.gradle
```diff
@@ -14,7 +14,7 @@ dependencies {
     implementation project(':capacitor-splash-screen')
 
 }
-apply from: "../../../../node_modules/cordova-plugin-ionic/src/android/cordovapluginionic.gradle"
```
kristojorg, line 17: Was this intentional? This is a capacitor plugin I installed that allows for live update functionality

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
```diff
@@ -77,15 +78,15 @@ public class ReaderPlugin: CAPPlugin {
 
 extension ReaderPlugin: BookmarkDelegate {
 
-    func createBookmark(bookmark: Bookmark) {
+    public func createBookmark(bookmark: Bookmark) {
```
kristojorg, line 81: Did these need to be marked as public? Just want to make sure the `BookmarkModule` isn't somehow depending on the `ReaderPlugin`, ideally the dependency would only flow the other direction and `BookmarkModule` is unaware of the `ReaderPlugin`

apps/oeh/android/app/capacitor.build.gradle
ettore: this was the result of me just trying to build the dev server + iOS app. I did not manually modify any of this. Nor did I try to build android, so I am not sure how this was modified. I can restore it if needed, but do you know how the build scripts would have changed this?

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin.swift
ettore: they need to, because these methods are part of a protocol defined in the BookmarkModule, and that protocol needs to be public since it's in a separate module.

kristojorg: Oh I see, in order to conform to the `BookmarkDelegate` they have to be public. Thanks for the explanation

PR #544, opened 2022-12-13
OE-771: Move chakraui and carousel into the UI package

Toxiapo:
This PR

1. Extracts the chakra base theme as its own package(ui), removed the chakra theme folder from OEW
2. Moves the shared carousel logic to the UI folder


apps/oeh/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### UNRELEASED CHANGES
 
+- OE-771: Import chakraui and carousel from the UI package
```
kristojorg, line 5: Actually 3.0.0 isn't released yet either. Let's just get rid of the 3.0.0 heading and put everything in UNRELEASED CHANGES for now

apps/oeh/src/components/BookCard.tsx
```diff
@@ -0,0 +1,38 @@
+import { chakra, Image, Link } from "@chakra-ui/react";
+import { IonCard } from "@ionic/react";
+import { SimpleBook } from "models";
+import { BookCover } from "ui";
+import React from "react";
+import { NavLink } from "react-router-dom";
```
kristojorg, line 6: I think we just need the normal `Link` as the `NavLink` will style the link in a special way when it matches the current url (usually used like in the header bar): https://v5.reactrouter.com/web/api/NavLink

apps/oeh/src/components/BookCard.tsx
```diff
@@ -0,0 +1,38 @@
+import { chakra, Image, Link } from "@chakra-ui/react";
```
kristojorg, line 1: Since this is a chakra component and needs to be wrapped in a chakra provider, I think we should separate it from other components like how we do in oew by making a components/chakra directory

packages/ui/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./chakra";
```
kristojorg, line 1: So this is a case where it might be useful to use the `exports` field of package.json to export a `ui/chakra` entrypoint individually. For the time being we don't need it, but I can imagine in the future we might want to import items from `ui` without getting chakra dependencies wrapped up in there. Just putting this comment here because I couldn't think of an example when we would want that previously.

packages/ui/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "type": "module",
+  "scripts": {
+    "test": "jest"
+  },
+  "dependencies": {
+    "@chakra-ui/react": "^2.4.3",
+    "@emotion/react": "^11.10.5",
+    "@emotion/styled": "^11.10.5",
+    "framer-motion": "^7.6.19"
```
kristojorg, line 21: I think these should all be peer dependencies since the host app is installing them themselves

kristojorg (approved):
Damn this was fast!

It looks like we had to recreate the `BookCard` and `Carousel` in oeh, presumably to make the linking work right. Is that correct? Was it also for styling reasons? This is actually probably not a bad thing to do. We can actually allow the UI for these to drift apart but still share the `useCarousel` hook, which is nice because I imagine we will want to make changes on mobile to the styles.

This looks great! I think the follow up PRs that I can see would be:

- see more button
- list view at end of see more
- rendering the real OE root lane

Toxiapo:
> It looks like we had to recreate the `BookCard` and `Carousel` in oeh, presumably to make the linking work right. Is that correct? Was it also for styling reasons?

This is correct. There are style disparities between OEW and OEH,e.g there are no prev/next buttons and the BookCard does not have title/author for OEH, one thing we can also do in the future is make the `bookToShow` as a param in the hook so we can set how many books to show for different apps if desire, but we don't need to worry about that now.

> This looks great! I think the follow up PRs that I can see would be:
> - see more button
> - list view at end of see more
> - rendering the real OE root lane

Yup! I think it might make more sense to get these updates as part of my `browser` ticket, which would be based on this branch.

apps/oeh/src/components/BookCard.tsx
Toxiapo: ahh, good to know! Updated.

packages/ui/index.ts
Toxiapo: I wanted to give this a try so I went ahead and updated this  : )

Toxiapo: I think we can also add the `icon` and `svg` folders in the UI package.

kristojorg (approved):

PR #545, opened 2022-12-13
OE-717: Re-enable cypress tests

Toxiapo:
This PR adds the MSW interceptors for the Edge run time and fixes some of our outdated cypress tests.

@kristojorg I wanted to add the sign in command in cypress and then toggle auth/unauth users, but unfortunately, I don't see a way to call the interceptor from Edge. I would suppose that when we call the `startNextSever` function, it should give us the option to run something in a specific environment, but I don't think they have that.


kristojorg:
WOOOOOOOOOOOO 🥳 

Excited about this. Also I think we should put a ticket on your board to make a PR to nextjs updating their msw example with this (if you're interested in doing that).

Toxiapo:
> WOOOOOOOOOOOO 🥳
> 
> Excited about this. Also I think we should put a ticket on your board to make a PR to nextjs updating their msw example with this (if you're interested in doing that).

Yeah, I would love to make this happen!

apps/oew/src/middleware.ts
```diff
@@ -146,6 +152,14 @@ export const config = {
...
      */
     "/app/collection/:collectionUrl*"
+  ],
+  /**
+   * This is used to fix the "Dynamic Code Evaluation (e. g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime"
+   * error from the @msw package
+   */
+  unstable_allowDynamic: [
+    "../../node_modules/is-generator-function/index*",
+    "../../node_modules/function-bind/**"
```
Toxiapo, line 162: Soooo this is the issue I am facing right now. It's not able to find the node_module in the root @kristojorg. If you run the app with `npm run dev`, this shouldn't be a problem.

.github/workflows/integration-cypress.yml
```diff
@@ -1,60 +1,60 @@
...
+
+  #     - name: Tar build folder
+  #       run: tar -cf oew-build.tar ./apps/oew/.next
+
+  #     - name: Save build folder
+  #       uses: actions/upload-artifact@v2
+  #       with:
+  #         name: oew-build
+  #         path: oew-build.tar
+  #         if-no-files-found: error
```
Toxiapo, line 52: These steps are failing the CI, and I am not sure why... The problem was that the cypress couldn't find the environment variables during the build process, I suspect it was because when we unzip this build folder, some of the files were missing. 

I removed the whole `install` job and moved them inside the cypress containers, it seems fine now. Usually, the `npm i` step takes the most time to run, and since we are going to run that command as part of the cypress job already, I don't see the benefit of running the install here. Maybe I am missing something?

.github/workflows/integration-cypress.yml
```diff
@@ -76,29 +76,53 @@ jobs:
...
           install: false
           browser: chrome
           record: true
           parallel: true
           group: "OEW - Chrome"
-          spec: apps/oew/cypress/tests/**/*
           command-prefix: "--"
+          spec: cypress/tests/**/*
+          config-file: cypress.json
+          working-directory: apps/oew
```
Toxiapo, line 125: This is to accommodate our multi-workspace setup, if we need to run cypress tests for other apps, we could follow this example: https://github.com/zendesk/cypress-github-action/blob/19ae4d7502bfdedf2a3c47961e02ea209aad367b/.github/workflows/example-start-and-yarn-workspaces.yml#L41

apps/oew/cypress/plugins/startNextServer.ts
```diff
@@ -7,7 +7,7 @@ let app: NextServer;
 const startNextServer = async () => {
   if (app) return;
 
-  const dev = true;
+  const dev = false;
```
Toxiapo, line 10: I changed this back to prod mode, I don't have a preference so let me know if ya prefer dev.

.github/workflows/integration-cypress.yml
kristojorg: The idea here (if I recall correctly) was that we could build the app one time, and then spawn a bunch of parallel cypress containers that just download the built app files and run tests. If we run `npm i` and build inside of the containers, each one of the many containers will have to install and build as well

kristojorg: It's ok to do this and create a ticket to rework this in the future

Toxiapo: hmm okay, let's just keep it that way then. I moved back the install job. (finally resolved the build bug hah ;/)

apps/oew/cypress/tests/signin/signout.ts
```diff
@@ -1,36 +1,37 @@
 import {
   ALL_ACCESS_DETAIL_BOOK_PATH_ALMOST_MIDNIGHT,
-  AUTH_COOKIE_NAME
+  AUTH_COOKIE_NAME,
+  HIGH_SCHOOL_COLLECTION_PATH
   // HIGH_SCHOOL_COLLECTION_PATH
```
samanthaandrews, line 5: nit: remove comment

.github/workflows/playwright.yml
```diff
@@ -19,6 +19,9 @@ jobs:
         with:
           node-version-file: ".nvmrc"
 
+      # - name: Output Github Actions
```
samanthaandrews, line 22: nit: remove if no longer necessary

samanthaandrews (approved):
i wonder why the Playwright tests are reporting as "skipped"? but i think this is something we can investigate in another ticket

Toxiapo:
> i wonder why the Playwright tests are reporting as "skipped"? but i think this is something we can investigate in another ticket

I am not sure why tbh. I can see that it ran when I go to the GitHub Action history, but it is not updating here. Maybe it will fix itself when we open a new PR..

samanthaandrews:
> > i wonder why the Playwright tests are reporting as "skipped"? but i think this is something we can investigate in another ticket
> 
> I am not sure why tbh. I can see that it ran when I go to the GitHub Action history, but it is not updating here. Maybe it will fix itself when we open a new PR..

I remember running into this issue in the past for a different reason. Basically by default GH Actions only keeps track of the last deployment_status and not all of them. We could probably use this action to help us pick the right environment https://github.com/marketplace/actions/await-for-vercel-deployment

but I also don't even think we should be running these tests on every PR, and instead we should make them run on a daily cron instead. i'll make a ticket for @ayan1229 to change this.

kristojorg (approved):
looks great to me!

PR #546, opened 2022-12-13
Fix staging Docker deployment

Toxiapo:
@kristojorg Noticed that staging.openebooks.org hadn't been updated since 11/30, so this PR fixes the issue we had while running the dockerfile


samanthaandrews (approved):
🥇

PR #549, opened 2022-12-16
Pjr 119/add android bridge poc

roywatson:
This PR:

- Adds an Android Plugin proof of concept functionally based on the iOS Plugin POC.
- It demonstrates one possible architecture by leveraging Capacitors invocation of the Plugin derived class as a Singleton allowing the Plugin class to encapulate only the comms with Capacitor while separing the data processing to other classes, in this demo as a form or Repository while the Plugin serves as a data source/sink.
- The comments explore the threading used by Capacitor during the invocation of the Plugin's methods which drove UI update decisions.


apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/repository/BookmarkRepository.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    fun clearBookmarks() {
+        bookmarks.clear()
+    }
+
+
+    fun addBookmark(bookmark: Bookmark) {
+        // This part updates the local list in our activities recycler view
+        bookmarks.add(bookmark)
+        adapter.notifyDataSetChanged();
```
kristojorg, line 58: So I may be misreading this, but I don't think we want to update the native list directly when a user adds a bookmark. The data should flow from native -> hybrid -> native. So you would call `notifyListeners` with a `CREATE_BOOKMARK` action (as you already do), and then the hybrid code will add the bookmark to its internal state (and sync this with the server), and will call `updateBookmarks` to notify you that a new bookmark was added (initially in a loading state). 

Essentially, the native portion should not store the state. The hybrid side communicates with the server and is therefore the source of truth, and all changes to the data should flow through the hybrid app. Does that make sense? LMK if I'm reading this wrong btw

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/repository/BookmarkRepository.kt
```diff
@@ -0,0 +1,99 @@
...
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+import com.getcapacitor.JSObject
+import org.openebooks.app.R
+import org.openebooks.app.plugins.readerplugin.ReaderPlugn
+import org.openebooks.app.plugins.readerplugin.model.Bookmark
+
+object BookmarkRepository {
+
+    var bookmarks = mutableListOf<Bookmark>()
```
kristojorg, line 17: In my mind (which may not represent the realities of development), the `bookmarks` list would be immutable, only updated when `updateBookmarks` is called from the hybrid side with a new list

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/repository/BookmarkRepository.kt
roywatson: My thought process is that the RecyclerView is simply a temporary view that will not be included in the end app so it and any requirements imposed by it will be removed. Of course future requiremnt are yet to be defined.

By "requirements" I mean: 

You're not reading it wrong, except in android a recyclerview (adroid for listview) needs to have a data source that it can pull data from on demand by positional index (ie array index, record number or key for a database source). So by adding a RecyclerView we have added a need for a datasource (list) from which that view can pull data on demand by index. Actually the list gets it data from it's "Adapter" which in turn gets it from the data source available on demand by positional index. 

Another consideration is that Capacitors update call occurs on a different thread than the UI/RecyclerView is running so a direct update is difficult without play thread synchonization games which I thought was beyond the scope of this simple POC. It is however a possible concern for the future. But still wouldn't fullfil the need for an on demand datasource.

The alternative would be to implement a notification call back into Capacitor to have it "update" a single bookmark by position index. Then play the thread game, then update the UI. But every refresh of the UI could/would cause a sequence of these round trips to populate the RecyclerView.

Hope that cleared that up.

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/repository/BookmarkRepository.kt
roywatson: That is a very astute observation.

Suplementing my above comment:

I chose immutable to avoid object creation and destruction churning and the resulting garbage collection. It is axiomatic that hybrid apps are by their very nature less cpu and memory efficient and as the app is to target older , slower, cpu and memory constrained devices (read "Cheaper") I felt any contribution to efficiency that the native code can provide should help.

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/repository/BookmarkRepository.kt
roywatson: Update: I pushed an update that removes the local addition of bookmarks to the native recyclerview. Before it would make the update and receive the update back from Capacitor, now it simply waits for the update to flow back through capacitor. I also added a random uuid for the new bookmark id which I had forgotten to do.

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/repository/BookmarkRepository.kt
roywatson: Update: A little more info on why the mutable list makes sense in this scenario. The Recyclerview is populated by an 
Adapter that requires a datasource, which in this case is an in-memory list. Updates to the content od the recyclerview are managed by the adapter and are triggered by changes in the underlying datasource. That is the standard Android recyclerview usage. So clearly a mutable list is the appropriate datasource here.

kristojorg: That all sounds great, thanks for the thorough explanation! My comment was more to clarify how I intended the plugin to interact with the JS code than it was to question how you were implementing this. My point was that the android code shouldn't update the list of bookmarks directly, so it is conceptually immutable at least. It was interesting to read about how kotlin works and why it makes sense for this to be mutable : ) Cheers!

samanthaandrews:
Hello everyone! @MalcolmMcFly I'd like to get your eyes on this.

@roywatson and @kristojorg should we merge it once Malcolm has reviewed?

Wondering what the next steps are for android development?

kristojorg:
Yep I think once @MalcolmMcFly has reviewed then we are good to merge this!

roywatson:
Removed the mutation in delete. Added threading to make work.

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/ReaderPlugn.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    @PluginMethod
+    fun updateBookmarks(call: PluginCall) {
+        val data = call.data
+        BookmarkRepository.setBookmarks(data)
+        call.resolve()
+    }
+
+    @PluginMethod
+    fun openBook(call: PluginCall) {
```
kristojorg, line 94: One last update: this method was renamed by @ettore to `openBooksmarksList` with the same arguments. We will make a new method called `openBook` in a future PR to actually open a book.

kristojorg (changes requested):
Just the one method name change, but otherwise looks great

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/ReaderPlugn.kt
ettore: btw the function is actually `openBookmarksList` 😅

kristojorg (approved):
Awesome! Thanks @roywatson !

PR #550, opened 2022-12-16
chore: remove old Increment plugin

kristojorg:
This PR remove the old increment plugin. It was the first hybrid/native communication POC and now that we have the bookmark POC implemented on both platforms, we don't need this anymore.


ettore (approved):

ettore:
should we file a ticket for that failing check? I've seen it fail on every PR. Is that something that should run only on deployment (i.e. main, release, feature) branches?

samanthaandrews:
> should we file a ticket for that failing check? I've seen it fail on every PR. Is that something that should run only on deployment (i.e. main, release, feature) branches?

@ayan1229 has a PR up that fixes this! I just need to approve it

PR #552, opened 2022-12-19
Update README for /oeh

ErnestFan:
Minor changes on the README. 

We need to run `npm run build` before  `npx cap sync` or `npm run sync:live`. Otherwise the sync would fail because `apps/oeh/dist` does not exist.


apps/oeh/README.md
```diff
@@ -42,6 +42,7 @@ To run the app for development, these are the conceptual steps:
 1. Install dependencies using NPM.
 1. Build the Typescript source and dependencies into Javascript that browsers understand, and serve it on a local development server.
 1. Make the Javascript available to the native code by serving it on a development server.
+1. Create web bundle
```
kristojorg, line 45: Creating a web bundle is accomplished by the step on line 43 above. Running a dev server creates a bundle, serves it on `localhost:8100` and rebuilds it whenever the source files change.

kristojorg (changes requested):
Thanks for the PR! Let's plan to sync up on what problem you were running in to. There should be no need to run `npm run build` if you are using a `npm run sync:live`. `build` creates a static bundle which can then be sync'd to the native project using `npm run sync:static`. However, if you are using a live bundle (served from a dev server, not a static set of files), then you should only need to run `npm run dev` and then in a new tab `npm run sync:live`. 

It sounds like the source of the confusion is around the two different ways to run the app (live web bundle, static web bundle). Can we plan to chat so I can figure out how to change my explanation so it's more clear?

kristojorg:
For reference, [this is the section](https://github.com/NYPL/ereading-clients/blob/670b5bf3719aa5177067f5f7b4cbc070291ceb57/apps/oeh/README.md#running-with-a-static-web-bundle) of the README that describes the static bundle process:

ErnestFan:
When I ran `npx cap sync` (which I shouldn't) and `npm run sync:live` and it brought me to some errors and warnings, I was confused and followed the suggestion from `npx cap sync` error. I thought `npm run build` is necessary before the sync and therefore created this PR. 
After talking to @kristojorg and re-reading the original README, things cleared up and I am closing this PR.

PR #553, opened 2022-12-20
OE-751 step 1: use full collection

kristojorg:
This is step 1 in my download book ticket. It is to use the full OE collection instead of a subset. It also gets rid of book loading screens by caching the book data from an opds feed in the SWR cache for the book entry. This is then revalidated on book detail page render, which provides the updated loan state (opds feeds don't contain loan info).


apps/oeh/src/chakra/components/BookCard.tsx
```diff
@@ -8,13 +8,15 @@ const BookCard = React.forwardRef<
   HTMLLIElement,
   { book: SimpleBook; className?: string; baseUrl: string }
 >(({ book, className, baseUrl }, ref) => {
+  const encodedBookUrl = encodeURIComponent(encodeURIComponent(book.url));
```
Toxiapo, line 11: I wonder if we should make this a function and put in some comments for when or why we need to double-encode an URL. I recall we had the same problem, but I can't remember why..

kristojorg: Yea I can put in a comment to explain. It's actually not obvious to me why we need to do it either haha

Toxiapo (approved):

PR #554, opened 2022-12-21
Oe 693/error pages

oliviawongnyc:
Updates OEW Error component with new styling.
Ticket: [OE-693](https://jira.nypl.org/browse/OE-693)
[Figma file](https://www.figma.com/file/hL3ywPjlfvx5ij5GNTnmzI/OEB-404-error?node-id=1519%3A60&t=Vhqp7p7PKsNkQvV7-0)


samanthaandrews:
I noticed one weird CSS thing in the header/breadcrumbs while doing manual testing. Screenshot attached.
Screen Shot 2022-12-23 at 2 36 54 PM

apps/oew/src/components/chakra-ui/Error.tsx
```diff
@@ -18,31 +28,104 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
...
   const router = useRouter();
 
+  const { setStoredBreadcrumbs } = useBreadcrumbContext();
+
+  const [data, setData] = useState<SimpleBook[]>();
+  const [isLoading, setLoading] = useState<boolean>(false);
+
+  // TODO: Create the same lane on production and add endpoint here.
+  const lostLaneEndpoint =
+    // IS_PRODUCTION_SERVER
```
samanthaandrews, line 38: Let's definitely implement this before we merge this PR

samanthaandrews:
Let's also make another ticket to implement a Cypress test for these error pages once @Toxiapo 's Cypress test PR gets merged.

Toxiapo:
> I noticed one weird CSS thing in the header/breadcrumbs while doing manual testing. Screenshot attached. 

This was a bug I introduced from one of my refactoring PR : ). I have a fix [here](https://github.com/NYPL/ereading-clients/pull/559). Feel free to merge whenever if it looks good to ya!

apps/oew/src/components/chakra-ui/Error.tsx
```diff
@@ -18,31 +28,104 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
...
+  useEffect(
+    () =>
+      setStoredBreadcrumbs([
+        { text: "All Books", href: "/app" },
+        { text: status ? status.toString() : "Error", href: "/app" }
+      ]),
+    [setStoredBreadcrumbs, status]
+  );
+
+  useEffect(() => console.error(detail), [detail]);
```
Toxiapo, line 64: Do we need this?

apps/oew/src/components/chakra-ui/Error.tsx
```diff
@@ -18,31 +28,104 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
...
+
+  useEffect(() => {
+    if (status === 404) {
+      setLoading(true);
+      fetchCollectionBooks(lostLane).then(data => {
+        setData(data);
+        setLoading(false);
+      });
+    }
+  }, []);
```
Toxiapo, line 53: I think you need the lostLane and data dependencies here.

apps/oew/src/components/chakra-ui/Error.tsx
```diff
@@ -18,31 +28,104 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
...
+
+  useEffect(
+    () =>
+      setStoredBreadcrumbs([
+        { text: "All Books", href: "/app" },
+        { text: status ? status.toString() : "Error", href: "/app" }
```
Toxiapo, line 59: I don't think `number` type has a `toString()` method, this would throw a syntax error. Perhaps use something like `String(status)`

apps/oew/src/components/chakra-ui/Error.tsx
```diff
@@ -18,31 +28,104 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
...
-          <Text as="h1">
-            {status ?? ""} Error: {title}
-          </Text>
-          <Text as="p" my={3}>
-            {detail && `${detail}`} <br />
-          </Text>
-          {shouldRefresh(status) && (
-            <>
+    <AppLayoutPage>
+      <ChakraProvider theme={theme}>
```
Toxiapo, line 68: While we are at it, do you think it's easy to convert this component to chakra and just use the `MarketingLayoutPage` instead? If not, we can do it in a separate PR

apps/oew/src/components/chakra-ui/Error.tsx
```diff
@@ -18,31 +28,104 @@ const ErrorComponent: React.FC<ErrorComponentProps> = ({
 }) => {
   const router = useRouter();
 
+  const { setStoredBreadcrumbs } = useBreadcrumbContext();
+
+  const [data, setData] = useState<SimpleBook[]>();
```
Toxiapo, line 33: Let's rename this to `books` or `bookData`.

Toxiapo (changes requested):
This looks good. Just a few small comments.

apps/oew/src/components/chakra-ui/Error.tsx
oliviawongnyc: I wasn't sure what to do with this detail as the new design doesn't use it. I'll just get rid of it.

apps/oew/src/components/chakra-ui/Error.tsx
oliviawongnyc: I tried this, but doesn't the MarketingLayoutPage look pretty different from the AppLayoutPage? And I believe the designs from Jennifer use the AppLayoutPage. LMK what you think.

apps/oew/src/components/chakra-ui/Error.tsx
Toxiapo: We need the detail for some of the errors, such as the server or decryption errors, because this file is shared by all errors that occurred in the app, showing "Page Not Found" doesn't make sense if it's a valid page caused by an API failure. 

I think we should keep the `detail` variable, and for 404 errors, we can leave this as an empty string, which can be modified directly from the 404.tsx file.

I would also consult with Jen on how she wants to handle the other errors. She has a template for the 400 error code that displays the detail but I don't see the designs for the other ones.

apps/oew/src/components/chakra-ui/Error.tsx
Toxiapo: You are right,  this is fine!

apps/oew/src/components/chakra-ui/Error.tsx
oliviawongnyc: Some comments/questions:

- I am displaying the title, which should change based on the type of error so it won't always read, "Page Not Found." 
- How are you thinking the detail should be used in the Error.tsx component?
- I believe that the non-400 error pages are supposed to look the same as the 400 error pages except that the title and status are updated and there is no "Are you Lost?" lane. Like this: https://www.figma.com/file/hL3ywPjlfvx5ij5GNTnmzI/OEB-404-error?node-id=1533%3A345&t=cr8SISegkZ726dRv-0.

Toxiapo: > I am displaying the title, which should change based on the type of error so it won't always read, "Page Not Found."

ah yes, that makes sense.
>How are you thinking the detail should be used in the Error.tsx component?

I was under the impression that we only show the lost lane and the custom message from Jen's design for the 404 error code, and keep the current setup for the other error with details, such as (i.e. https://github.com/NYPL/ereading-clients/blob/6e470fcfa3c0d2215eccaa7a5e39f63befefbac5/apps/oew/src/components/theme-ui/WebReader.tsx#L140-L144), but maybe I misunderstood?

> I believe that the non-400 error pages are supposed to look the same as the 400 error pages

I see, if that's the case then this is fine then

kristojorg:
For this 404 page, I am wondering if we can use the nextjs custom 404 page setup which would allow us to statically render the page and thus not need to wait for the carousel to load (and potentially have a subsequent error when loading). Documentation is here: https://nextjs.org/docs/advanced-features/custom-error-page#404-page

oliviawongnyc:
Thanks, @kristojorg! Looking into this today.

kristojorg:
@oliviawongnyc let me know when this is ready for review again : )

oliviawongnyc:
> @oliviawongnyc let me know when this is ready for review again : )

@kristojorg oh no! Thought I'd give you the go ahead already. I think I'm ready for another review. Thank you!

samanthaandrews (approved):
🚀

Toxiapo (approved):
LGTM!

PR #557, opened 2022-12-23
OE-768: Add bookdetail page wireframe

Toxiapo:
This PR adds the book detail wireframe to the hybrid app. Note that I am using css-in-js everywhere as we are still deciding what to do with our styling strategies.


apps/oeh/src/pages/BookDetail.tsx
```diff
@@ -9,188 +8,146 @@ import {
...
-  const { isDownloaded, download, removeDownload } = useDownload(data);
+  const { data: book, mutate, isLoading, isValidating } = useBook(bookUrl);
+  const { data: recommendations } = useCollection(book?.relatedUrl as string);
+
+  // Need to refetch the loan endpoint so it gets the most up-to-date book details
+  // This is slow...
+  const { data: loanedBooks } = useSWR(
+    credentials?.token ? [LOANS_ENDPOINT, credentials.token] : undefined,
+    fetchLoans
+  );
```
Toxiapo, line 62: unfortunately, we have to re-fetch data from the loans endpoint until we fix the cache header issue for the single ODPS entry API

apps/oeh/src/hooks/useBook.ts
```diff
@@ -1,17 +1,89 @@
...
+        Authorization: credentials.token
+      };
+      await removeDownload();
+
+      // Spamming a revoke request throws a 404 error hence we wrap this
+      // in a try catch
+      const bookRefetched = await fetchBook(revokeUrl, ROOT_LANE_URL, {
+        headers,
+        method: "PUT"
+      });
```
kristojorg, line 77: What do you mean spamming?

apps/oeh/src/pages/BookDetail.tsx
```diff
@@ -9,188 +8,145 @@ import {
...
+ * FIXME: re-work this logic
+ */
+const basePaths: string[] = ["/app/my-books", "/app/browse"];
+const getBaseUrl = (currentUrl: string) => {
+  return (
+    basePaths.find(baseUrl => {
+      return currentUrl.includes(baseUrl);
+    }) || "/app/browse"
+  );
+};
```
kristojorg, line 47: Let's make a ticket for this describing what needs to be changed

apps/oeh/src/pages/BookDetail.tsx
```diff
@@ -9,188 +8,145 @@ import {
...
   const bookUrl = decodeURIComponent(match.params.bookUrl);
+  const baseUrl = getBaseUrl(match.url);
+
   const { credentials } = useCredentials();
-  const [present, dismiss] = useIonLoading();
 
-  const { data, mutate } = useBook(bookUrl);
-  const { isDownloaded, download, removeDownload } = useDownload(data);
+  const { data: book, mutate, isLoading, isValidating } = useBook(bookUrl);
+  const { data: recommendations } = useCollection(book?.relatedUrl as string);
```
kristojorg, line 56: This could be moved into the recommendations component

packages/models/utils/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./shared";
+export * from "./string";
```
kristojorg, line 2: I would maybe name this something different or just put it in the shared utils file. Not sure that string means anything in this context

kristojorg (changes requested):
A few comments but this looks good!

apps/oeh/src/pages/BookDetail.tsx
Toxiapo: https://jira.nypl.org/browse/OE-782

apps/oeh/src/components/BookDetail/Recommendations.tsx
```diff
@@ -0,0 +1,53 @@
...
+      title: lane.title,
+      url: lane.url,
+      books: []
+    };
+
+    filtered.books = lane.books.filter(curr => curr.id !== book.id);
+    prev.push(filtered);
+
+    return prev;
+  }, []);
```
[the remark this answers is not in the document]
Toxiapo: Thank you!

apps/oeh/src/chakra/components/BookCard.tsx
```diff
@@ -27,6 +27,11 @@ const BookCard = React.forwardRef<
       transition={"all 0.2s"}
       transform="scale(.97)"
       _hover={{ transform: "scale(1)" }}
+      /**
+       * By default, router pushes a page to the route stack, we want to just replace the last page
+       * to avoid the "Back" button loop.
+       */
+      replace={true}
```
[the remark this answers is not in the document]
Toxiapo: Honestly, I am not sure what the expected behavior is, and I would like to hear your thoughts on this.

If you visit a book and land on the detail page, pressing the back button will take you back to the page you came from. So if it's from My loans page, it would take you back to my loans page.

However, there's a scenario where when you are on Book A, and you click on any of the books from the recommended lanes, the back button would take you back to the Book A page, instead of the *root*.

The replace prop will make sure when navigating, we just replace the history instead of adding another page frame to the history stack. Unless we want this?

apps/oeh/src/components/BookDetail/TimeBanner.tsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import { OpenEBook } from "models";
+
+type TimeBannerProps = {
+  bookData: OpenEBook;
+};
+export const TimeBanner: React.FC<TimeBannerProps> = ({ bookData }) => {
```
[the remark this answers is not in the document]
Toxiapo: Yeah, it was removed from the design on the wireframe file after I added this.

apps/oeh/src/hooks/useBook.ts
```diff
@@ -26,3 +98,13 @@ const fetchBookWithToken: Fetcher<
...
   });
 };
+
+async function mutateBookWithKey(book: OpenEBook, token: string) {
+  await globalMutate([book.url, token], book, {
+    revalidate: false
+  });
+}
+
+async function mutateLoans(token: string, book: OpenEBook) {
```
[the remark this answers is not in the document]
Toxiapo: Updated!

apps/oeh/src/hooks/useBook.ts
Toxiapo: When the returns the book, there's a small delay where the "Return" button is still visible on the screen. If the user clicks on the button the second time, it will throw a fatal error something along the line " No active loan or hold"

apps/oeh/src/pages/BookDetail.tsx
Toxiapo: I moved useCollection hook to the recomendations component, I think we still need the mutate function from useBook here.

apps/oeh/src/hooks/useBook.ts
```diff
@@ -1,17 +1,89 @@
...
+      return;
+    }
+
+    present({ message: "Borrowing..." });
+
+    const headers: RequestInit["headers"] = {
+      Authorization: credentials.token
+    };
+
+    const bookRefetched = await fetchBook(borrowUrl, ROOT_LANE_URL, {
```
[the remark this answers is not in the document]
Toxiapo: Updated

apps/oeh/package.json
```diff
@@ -38,6 +38,7 @@
     "framer-motion": "^7.6.19",
     "jest-presets": "*",
     "jsdom": "^20.0.3",
+    "model": "^6.0.1",
```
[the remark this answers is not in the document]
Toxiapo: Whoops, not sure how this got here. Removing!

apps/oeh/src/components/BookDetail/Description.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type DescriptionProps = {
+  description: string | undefined;
+};
+const Description: React.FC<DescriptionProps> = ({ description }) => {
+  if (!description) return null;
+
+  const [isLongText, setIsLongText] = useState<boolean>(
+    description.length > TEXT_LIMIT
+  );
```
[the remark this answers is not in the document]
Toxiapo: OHH you are right, good catch. I will try to add a eslintrc file to it.

apps/oeh/src/chakra/components/BookCard.tsx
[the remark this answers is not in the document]
Toxiapo: I am going to just remove this for now and keep the default behavior... It's acting weird on a fresh sign in account. Sometimes the back button takes the user back to the signin/agreement page

apps/oeh/src/pages/BookDetail.tsx
```diff
@@ -1,196 +1,153 @@
...
+          <IonItem>
+            <ReportAProblem />
+          </IonItem>
+        </div>
+      </IonContent>
+    </IonPage>
+  );
+};
 
+const Loading = () => {
```
samanthaandrews, line 129: Doesn't need to happen in this PR, but if this loading component can be re-used elsewhere we should extract it into its own file.

apps/oeh/src/components/BookDetail/Description.tsx
```diff
@@ -0,0 +1,63 @@
...
+        style={{ margin: 0, paddingBottom: 3 }}
+        dangerouslySetInnerHTML={{
+          __html: isLongText
+            ? truncatedText
+            : description ?? "Description not provided."
+        }}
+      />
+      {isLongText && (
+        <a
+          href="#"
```
samanthaandrews, line 44: Is this href something that we are planning to set in a later ticket? That's fine with me if so, just want to make sure

apps/oeh/src/chakra/components/BookCard.tsx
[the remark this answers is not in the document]
kristojorg: I think we do want to have the stack as that's generally how apps with tabs work (I think?). We can ask jen but for now I think leave it. Consider for example if you click on a user in instagram and then click another and then another and another it just keeps adding them to your stack I believe and you have to go back through all of them

apps/oeh/src/components/BookDetail/Description.tsx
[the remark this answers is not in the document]
kristojorg: Oh yea i thought I had added one initially but I guess not! Thanks : )

apps/oeh/src/chakra/components/BookCard.tsx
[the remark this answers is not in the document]
Toxiapo: Ah yes this makes a lot of sense!

apps/oeh/src/components/BookDetail/Description.tsx
Toxiapo: There shouldn't be any href. It probably makes more sense to convert this to a button. good catch.

apps/oeh/src/hooks/useBook.ts
kristojorg: Should we just make the button disabled when returning is in progress?

Toxiapo: Good idea. Updating.

kristojorg (commented):
I noticed a couple things and pushed a small commit changing them. I hope you don't mind!

apps/oeh/src/components/FulfillmentCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  // Need to refetch the loan endpoint so it gets the most up-to-date book details
+  // This is slow...
+  const { data: loanedBooks } = useSWR(
+    credentials?.token ? [LOANS_ENDPOINT, credentials.token] : undefined,
+    fetchLoans
+  );
+
+  const bookData =
+    loanedBooks?.find((loanedBook: OpenEBook) => book?.id === loanedBook.id) ??
+    book;
```
Toxiapo, line 34: @kristojorg This will always use the cached book because the loan endpoint does not have the book if it's been returned.

kristojorg: Just marking this for the record: we will resolve this via a server side fix to the cache-control header

apps/oeh/src/components/BookDetail/BookTitle.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+type BookTitleProps = {
+  title: string;
+  authors: string[];
+};
+const BookTitle: React.FC<BookTitleProps> = ({ title, authors }) => {
+  return (
+    <>
+      <h1 style={{ margin: 0, fontSize: "18px" }}>{title}</h1>
+      <h5 style={{ marginTop: 0, fontSize: "12px" }}>{authors?.join(", ")}</h5>
```
kristojorg, line 11: Technically for accessibility we are only supposed to go down one level at a time in the headings. I don't think the authors should be a heading though since they're not introducing a new section. Just a div/span would be fine

kristojorg (changes requested):
Okay this looks good and ready, just the one tiny comment about the header tag!

apps/oeh/src/components/BookDetail/BookTitle.tsx
Toxiapo: Updated, ty!

kristojorg (approved):
LGTM!

PR #559, opened 2022-12-23
OE-781: Header height fix and refactor web-theme

Toxiapo:
This PR fixes the header height bug and reactors the chakra theme for web.


apps/oew/src/components/chakra-ui/header/Header.tsx
```diff
@@ -28,7 +29,7 @@ export const Header: React.FC<HeaderProps> = ({
       top={0}
       left={0}
       as="header"
-      height="header"
+      height={HEADER_HEIGHT}
```
kristojorg, line 32: Curious why it doesn't work to have the header height in the theme?

Toxiapo: We are using the shared base theme from the `ui` package. I didn't want to put any oew specific settings there, so I removed this header from the package. If we host our own base theme for oew then we can add this in.

kristojorg: I see I see, that was smart not to include it in the shared ui package. One way we could still have it in the theme is to extend the theme from the ui package adding web-specific items as needed 
```
const webTheme = extendTheme(baseTheme, { sizes: {...} })
```

but this is also a fine solution

Toxiapo: I like it. Let's just refactor it here then so we don't need to worry about it later on.

packages/ui/package.json
```diff
@@ -13,6 +13,11 @@
       "types": "./chakra/index.ts",
       "import": "./chakra/index.ts",
       "require": "./chakra/index.ts"
+    },
+    "./chakra/webTheme": {
+      "types": "./chakra/webTheme/index.ts",
+      "import": "./chakra/webTheme/index.ts",
+      "require": "./chakra/webTheme/index.ts"
```
Toxiapo, line 20: I wanted a web-specific entry point so imports would be more clear, lmk what ya think.

kristojorg: I would actually put the web theme inside the oew source since it is web-specific. It can import and extend from the them in the ui folder. Same for any oeh theme we use

Toxiapo: Updated, thank you!!

samanthaandrews (approved):
LGTM

PR #561, opened 2022-12-28
Basic iOS CI for PRs

ettore:
I'm adding a workflow to build the iOS app and run the native unit tests. While I was working on this, it already caught a failure I was introducing in my previous PR.

Adding @Toxiapo since it seems like you worked on CI in this repo... hopefully there's nothing problematic in what I'm doing here.


apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -34,6 +34,8 @@
 		504EC3131FED79650016851F /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
 		50B271D01FEDC1A000F3C39B /* public */ = {isa = PBXFileReference; lastKnownFileType = folder; path = public; sourceTree = "<group>"; };
 		73001A8729415DE6000BDCED /* BookmarkModule */ = {isa = PBXFileReference; lastKnownFileType = wrapper; name = BookmarkModule; path = SwiftPackages/BookmarkModule; sourceTree = "<group>"; };
+		731A8F3F295B714400AF3257 /* .github */ = {isa = PBXFileReference; lastKnownFileType = folder; name = .github; path = ../../../../.github; sourceTree = "<group>"; };
```
ettore, line 37: adding these to project to ease editing. These files are not copied to app bundle.

apps/oeh/ios/App/Podfile
```diff
@@ -15,7 +15,6 @@ def capacitor_pods
   pod 'CapacitorFilesystem', :path => '../../../../node_modules/@capacitor/filesystem'
   pod 'CapacitorPreferences', :path => '../../node_modules/@capacitor/preferences'
   pod 'CapacitorSplashScreen', :path => '../../../../node_modules/@capacitor/splash-screen'
-  pod 'CordovaPlugins', :path => '../capacitor-cordova-ios-plugins'
```
ettore, line 18: @kristojorg do we need this? This was automatically removed by `npm run sync:static`, so I went with that. The build succeeds with or without this. I also took a look at the files under the `capacitor-cordova-ios-plugins` directory, and it seems like they are pretty much empty.

kristojorg: I had tried to use it to enable live updates but didn't succeed. I'll add it back when I need it, thanks for the heads up!

.github/workflows/ios-unit-testing.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Checkout main repo and submodules
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: ".nvmrc"
+      - name: Init node dependencies
+        run: npm install
```
kristojorg, line 25: Installing node dependencies can be very slow, but can be sped up with caching. The `setup-node` action has a cache option built in. You can enable it like so:
```
- uses: actions/setup-node@v3
   with:
     node-version: 16
     cache: 'npm'
- name: Install node dependencies
   run: npm ci
```

kristojorg (commented):
Sweet, looks good to me! Just the one optimization would be great otherwise this will be very slow

.github/workflows/ios-unit-testing.yml
```diff
@@ -0,0 +1,33 @@
+name: iOS Unit Tests
+on:
+  pull_request:
```
kristojorg, line 3: Does this need a key? Or is it ok this way?

ettore: I haven't used keys in this context, the idea was to have this run on every new pull request being opened.

Toxiapo: This also runs on every commits pushed to the pull request if I understand this correctly. 

I think if we want to run only on PR opens, we need to scope it:
```
on:
  pull_request:
      types: [opened]
```

ettore: hmm, but I think we do want the check to re-run for new commits to the PR, especially if the check fails and we are trying to fix it.

Toxiapo: Ahh sorry, I misread. Yea this is good then!

kristojorg: Sounds good! Just seemed like strange notation but I guess that's just how yml works!

kristojorg (approved):
nice!

PR #562, opened 2022-12-29
PJR-124 Set explicit Xcode scheme name capacitor config file

ettore:
This fixes the build broken by previous PR. I think Capacitor was defaulting to use a scheme name homonymous with the Xcode project name.


Toxiapo (approved):
🎉 Thank you for the PR!

oliviawongnyc (approved):

PR #563, opened 2022-12-30
PJR-129 Add native navigation support & rename `openBook` function

ettore:
Native navigation is hidden by default for the web views, letting Capacitor handle its own web-based nav. Once we need to display a native view, we will show the native navigation. Once the native view disappears, we'll hide the native nav.


ettore:
@Toxiapo do you happen to know why the Playwright check is failing? the log mentions cucumber, but I have not touched that.

Toxiapo:
@ettore Yeah, it's not your changes that caused this. I believe it's an ongoing issue with the Clever test, I will try to retrigger it. 
@ayan1229 may know have more context.

ettore:
@kristojorg  I added 2 more commits that are tangentially related. The "reorder" commit doesn't provide any functional changes. In the last commit ("Rename openBook...") I renamed the `openBook` function to `openBookmarksList` since that's what it actually does. I'd like this change because (1) I want to use the "openBook" function for actually open the reader, (2) it might be good for now to keep the BookmarkModule for reference, and so (3) having a "openBook" function will be confusing in the near term.

Hopefully I changed all the JS files I needed!

@roywatson @MalcolmMcFly this change might affect you too, please let me know of any issues. I'd like to merge this soon to move forward with R2 integration.

ErnestFan (approved):

apps/oeh/ios/App/App/Base.lproj/Main.storyboard
```diff
@@ -1,19 +1,36 @@
...
         <!--Bridge View Controller-->
         <scene sceneID="tne-QT-ifu">
             <objects>
-                <viewController id="BYZ-38-t0r" customClass="CAPBridgeViewController" customModule="Capacitor" sceneMemberID="viewController"/>
+                <viewController id="BYZ-38-t0r" customClass="CAPBridgeViewController" customModule="Capacitor" sceneMemberID="viewController">
+                    <navigationItem key="navigationItem" id="ISh-oU-SZ7"/>
+                </viewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="dkx-z0-nzr" sceneMemberID="firstResponder"/>
             </objects>
+            <point key="canvasLocation" x="-159" y="421"/>
```
kristojorg, line 18: I'm curious how review would normally work for a file like this? Is this generated from an Xcode GUI?

ettore: yeah, reviewing this is kind of a dark art 😄.... In brief, I just edited the storyboard visually, adding a navigation controller at the root.  I can add a screenshot of the storyboard if you want?

ettore: main-storyboard

kristojorg (commented):
I'm glad that using a navigation stack alongside capacitor appears to be working okay! 

A couple notes:

- I'm fine with changing the name of the plugin call since we aren't actually opening a book right now. We will have to update the plugin in the android code as well (which isn't merged yet), but that shouldn't be difficult. Thanks for updating the web code!
- I am sometimes seeing an error that freezes the UI in the native view when I click the "new bookmark" button. This is what the console says:

```
2023-01-04 19:27:33.911582+0100 Open eBooks[64740:9424275] [LayoutConstraints] Changing the translatesAutoresizingMaskIntoConstraints property of a UICollectionViewCell that is managed by a UICollectionView is not supported, and will result in incorrect self-sizing. View: <_UIAlertControllerTextFieldViewCollectionCell: 0x14bd516e0; frame = (0 0; 270 24); gestureRecognizers = <NSArray: 0x6000025157d0>; layer = <CALayer: 0x600002d287c0>>
```

kristojorg (approved):
My previous comment referenced an issue that was some local issue with my simulator. It seems to work fine now! Looks good

ettore:
force push above is to rebase my branch on top of staging. No changes from earlier.

PR #564, opened 2022-12-30
OE-775: Update deployment scripts & instructions for OEW 

samanthaandrews:
To do:

- [x] Update `.github/workflows/oew-release-prod.yml` with new triggers so that prod deployments happen when an `oe-web-<VERSION>` tag is pushed to a `release/oe-web-<VERSION>` branch
- [x] Update `apps/oew/README.md` with the latest deployment instructions
- [x] Change references to `staging` branch to `main`
- [x] Remove references to `production` branch
- [ ] Rename `staging` to `main` in GitHub
- [x] Delete the `production` branch


.github/workflows/oew-release-prod.yml
```diff
@@ -1,51 +1,28 @@
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: ncipollo/release-action@v1
         with:
-          tag_name: ${{ steps.autotagger.outputs.tagname }}
-          release_name: ${{ steps.autotagger.outputs.tagname }}
-          # use the body of the PR commit as the release body
-          body: ${{ github.event.pull_request.body }}
-          draft: false
-          prerelease: false
+          bodyFile: "apps/oew/CHANGELOG.md"
```
samanthaandrews, line 22: Maybe putting the whole `CHANGELOG.md` in the body is annoying? Instead we could just manually update the body of the release with the specific section of the `CHANGELOG.md` that it corresponds to. I'd prefer not to do any file/string parsing in this action.

Note: I wanted to set `generateReleaseNotes: true` but realized this wouldn't really help us since we are using a monorepo and the automatic release notes are generated between the previous release and the current release. In our case the previous release might be a hybrid release, which means the auto-generated release notes wouldn't be accurate.

.github/workflows/oew-release-prod.yml
```diff
@@ -1,51 +1,28 @@
...
+# workflow will only run when a tag with the oe-web-* prefix is pushed to a release/oe-web-* branch
 on:
-  pull_request:
-    branches: [production]
-    types: [closed]
+  push:
+    tags:
+      - "oe-web-*"
+    branches:
+      - "release/oe-web-*"
```
samanthaandrews, line 10: I _could_ change this to be `main`, which means the action would run when the `oe-web-<VERSION>` tag is pushed to the `main` branch, but imo this is more clear. This way, we can be sure that the docker image from the release branch is the one being pushed to ECS instead of looking for the one generated from a previous commit.

.github/workflows/oew-release-prod.yml
```diff
@@ -1,51 +1,28 @@
...
-          # use the body of the PR commit as the tag message
-          tag_message: ${{ github.event.pull_request.body }}
-
       - name: Release
-        id: create_release
-        if: steps.autotagger.outputs.tagname != ''
-        uses: actions/create-release@v1.0.0
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: ncipollo/release-action@v1
```
samanthaandrews, line 20: The `[actions/create-release@v1.0.0](https://github.com/actions/create-release)` action has been deprecated & archived so I switched to use another one that is still getting updated. It was recommended (along with a few others) in the deprecation notice.

.github/workflows/oew-release-qa.yml
```diff
@@ -5,7 +5,7 @@ name: OEW QA/Staging Release
 # workflow will only run when a pull request to the staging branch is closed.
 on:
   pull_request:
-    branches: [staging]
+    branches: [main]
```
samanthaandrews, line 8: I haven't renamed the `staging` branch yet. Will do so once this PR is approved.

apps/oew/README.md
```diff
@@ -246,18 +243,25 @@ When creating links using `<Link>`, you don't need to worry about whether it is
...
+
+Once this is complete, comment on the release reminder in Slack confirming the release has been made. 
 
-On the following Monday, assuming the PR has been approved and all status checks are passing, we generally make the release by completing the merge from `staging` into `production`. Create a merge commit rather than squashing and merging. This merge will trigger the `release-prod.yml` Github Action workflow file, which will automatically tag a release in Github, generate release notes, and build and deploy the latest version of the app to `openebooks.org`. Once this is complete, comment on the release reminder in Slack confirming the release has been made.
+Then, merge the release branch to `main`. If there are conflicts in the release branch, *do not* merge `main` into the release branch. We want the last commit on the release branch to be the release tag. Instead, checkout `main` and merge the release branch into `main` directly, fixing the conflicts manually. Alternatively, if you want to be more careful, you can checkout a new branch from `main`, merge the release branch into your new branch, and then merge your new branch into `main`.
```
samanthaandrews, line 264: If a branch is manually merged to `main` like this, what should we do with the PR? I'm guessing we should just close it, right?

.github/workflows/oew-release-prod.yml
Toxiapo: I don't have a strong preference, but I am ok with the release branch approach.

apps/oew/README.md
```diff
@@ -246,18 +243,25 @@ When creating links using `<Link>`, you don't need to worry about whether it is
...
+1. Request reviews from the team. Do not merge this PR (more on this below).
+1. Add the link to the PR to the release ticket in Jira and tag the appropriate member of QA, the project manager, and the product manager. Move the ticket to "Ready for QA" and assign it to the appropriate member of QA.
+1. Comment on the release reminder in Slack confirming the above has been completed.
+
+If critical bugs are discovered during the QA process for a release, they are fixed on the release branch.
+
+On the following Monday or Tuesday (we don't release on Fridays), assuming the release has been approved by engineers, by QA, and all status checks are passing, create and push a tag to the release branch with the format `oe-web-<VERSION>`, i.e. `oe-web-1.2.3`. Pushing this tag to the release branch will trigger the `oew-release-prod.yml` Github Action workflow file, which will automatically build and deploy the latest version of the app to `openebooks.org`. No additional commits should be made to the release branch after the tag has been pushed.
+
```
Toxiapo, line 261: I have never worked with tags this way, I found an [StackOverflow answer here](https://stackoverflow.com/a/70182050/17573442), is this what we need to do?

samanthaandrews: Yup, that's exactly right. Create a tag and then push it. (I've also never worked with tags so I'm planning on asking @ettore to run through the first deployment with us using this new process)

ettore: yep that works. Although my preferred way to use `git tag` is with the `-a` option because then you can add tag notes (similarly to `git commit`). In summary this is what I do:
```
$ git checkout release/oe-web-1.2.3
$ git tag -a oe-web-1.2.3  8ea4d5533266b5c64728db351285d9a8bf3916b4
<enter release info manually>
git push --follow-tags
```

instead of `git tag -a` you can also do `git tag -m` for a one-liner tag note (just like git commit -m).

apps/oew/README.md
```diff
@@ -246,18 +243,25 @@ When creating links using `<Link>`, you don't need to worry about whether it is
...
+1. Once you've created the release branch, review the "UNRELEASED CHANGES" in the CHANGELOG. Update the "Fix version/s" of the corresponding tickets in Jira with the new release number. Move any tickets to "Ready for Prod" that aren't already there.
+1. In the CHANGELOG, update "UNRELEASED CHANGES" to the new version number. Update the version number in `package.json`. Run `npm install` to update `package-lock.json`.
+1. Once you've made the above changes, push the release branch and make a Pull Request. Give the pull request a name that includes the version number, for example `Release 4.7.21`. Uncomment the production PR template checklist in the PR description and add additional context if helpful.
```
ettore, line 253: unsolicited feedback 😅  but should we include app and platform too here? i.e.  "Give the pull request a name that includes the app, platform and version number, for example `Release OE Web 4.7.21`"

apps/oew/README.md
Toxiapo: @ettore thank you!!

samanthaandrews: Yes, thank you @ettore. I'm going to add this context to the main README and link it to the OEW README. What would the content in `<enter release info manually>` be? The release notes?

apps/oew/README.md
samanthaandrews: @ettore your feedback is never unsolicited!! This is a good catch and I made the change 👍

apps/oew/README.md
ettore: I usually add some more developer-focused release notes, of course keeping in mind that these are still public. 

While the App Store is for patrons and so it's the product team's domain, GitHub is for developers: so adding the same release notes won't be most useful. For example for Open eBooks 2.5.0 the official release notes were:
```
Here’s what we’ve been working on:

- Redesigned sign in experience
- Accessibility in password input
- Bug fixes
```
which are not detailed enough for developers. So on Github we added the following release notes:
https://github.com/NYPL-Simplified/Simplified-iOS/releases/tag/openebooks-v2.5.0

which as you can see mention pretty important details such as an upgrade to R2 and what UI changes actually happened. All of that is also included in the git tag.

samanthaandrews: Got it. This is helpful. In our case it might be good to just copy and paste whatever is in the CHANGELOG for that specific release. I'll add this context to the README.

apps/oew/README.md
kristojorg: Github is smart enough to mark it as merged typically : )

.github/workflows/oew-release-prod.yml
kristojorg: So just to be clear, a docker image is released to production when we push a tag to the release branch, is that right?

samanthaandrews: Yup, that’s exactly right

kristojorg: Okay, it seems a bit scary at first blush for some reason, but let's try it!

apps/oew/README.md
```diff
@@ -246,18 +243,25 @@ When creating links using `<Link>`, you don't need to worry about whether it is
...
-5. Add the link to the PR to the release ticket in Jira and tag the appropriate member of QA.
-6. Comment on the release reminder in Slack confirming the above has been completed.
+1. Create a new branch off `main` with the name `release/oe-web-<VERSION>`, i.e. `release/oe-web-1.2.3`. The name of the branch is important because our GitHub Actions script checks for the `release/oe-web` prefix to make the deployment. In our case, this GitHub Actions script pushes a docker image to elastic container registry (ECR) and updates the elastic container service (ECS) with that new image in AWS. See the [Branching Conventions](../../README.md#branching-conventions) diagram and summary for more context.
+1. Once you've created the release branch, review the "UNRELEASED CHANGES" in the CHANGELOG. Update the "Fix version/s" of the corresponding tickets in Jira with the new release number. Move any tickets to "Ready for Prod" that aren't already there.
+1. In the CHANGELOG, update "UNRELEASED CHANGES" to the new version number. Update the version number in `package.json`. Run `npm install` to update `package-lock.json`.
```
kristojorg, line 252: Instead of updating manually and running npm install, I would use `npm version minor` which will bump the version number in both files without the install

apps/oew/README.md
```diff
@@ -246,18 +243,25 @@ When creating links using `<Link>`, you don't need to worry about whether it is
...
-3. Make a PR to merge `staging` into `production`. Give the pull request a name that includes the version number, for example `Release 4.7.21`. Uncomment the production PR template checklist in the PR description and add additional context if helpful.
-4. Request reviews from the team. Do not merge this PR until you are ready to make the release (more on this below).
-5. Add the link to the PR to the release ticket in Jira and tag the appropriate member of QA.
-6. Comment on the release reminder in Slack confirming the above has been completed.
+1. Create a new branch off `main` with the name `release/oe-web-<VERSION>`, i.e. `release/oe-web-1.2.3`. The name of the branch is important because our GitHub Actions script checks for the `release/oe-web` prefix to make the deployment. In our case, this GitHub Actions script pushes a docker image to elastic container registry (ECR) and updates the elastic container service (ECS) with that new image in AWS. See the [Branching Conventions](../../README.md#branching-conventions) diagram and summary for more context.
```
kristojorg, line 250: Just checking if my understanding is correct:

1. We make the `release/oe-web-1.2.3` branch and push it. This triggers Github Actions to build a new docker image, push it to ECR, and make that live on staging.openebooks.org
2. QA happens on staging.openebooks.org
3. When QA passes, we make a tag and push it to our branch. Ideally this wouldn't re-build the docker image but would simply run the same image created previously in production. 

Is that right?

apps/oew/README.md
samanthaandrews: Cool ok great!

apps/oew/README.md
samanthaandrews: Good idea. Updated.

apps/oew/README.md
samanthaandrews: Thank you for calling this out. I had completely forgotten that we needed to make some changes to our `oew-release-qa.yml` script to ensure Step 1 was happening. The flow you described above is now accurate.

apps/oew/README.md
```diff
@@ -1,52 +1,16 @@
...
-    - [Building With AxisNow Decryptor](#building-with-axisnow-decryptor)
-    - [Running the docker container](#running-the-docker-container)
-    - [From the command line](#from-the-command-line)
-    - [Helpful commands](#helpful-commands)
-    - [Credits](#credits)
+An OPDS web catalog client for Open eBooks patrons
+
+- **Production:** https://www.openebooks.org/
+- **Staging:** https://staging.openebooks.org/ - updated whenever updates are pushed to a `release/oe-web-*` release branch; see [Deploying to Production](#deploying-to-production) for details
+- **Dev:** https://dev-openebooks.vercel.app/ - everything that has been merged to `main`
```
samanthaandrews, line 13: @kristojorg @oliviawongnyc @Toxiapo I added these three URLs to the OE Web README.md. This PR makes an update so that `staging.openebooks.org` is now used for release candidate builds, and is not updated on every push to `main`. Code that is pushed to `main` will be viewable/testable at the Vercel URL `dev-openebooks.vercel.app`

samanthaandrews:
@kristojorg can you give this a final review today? i think it's ready to merge

kristojorg (approved):
The part that I'm still wondering about is whether we actually need to build a new image and push it to ECR to deploy the prod. Ideally we would just use the image that's already built for QA and promote it to production. I guess that involves changing an environment variable or two, but I think that would be ideal. We then really know exactly what's going to prod before it's deployed. Maybe we should just make a ticket to investigate that in the future? What do you think?

samanthaandrews:
> The part that I'm still wondering about is whether we actually need to build a new image and push it to ECR to deploy the prod. Ideally we would just use the image that's already built for QA and promote it to production. I guess that involves changing an environment variable or two, but I think that would be ideal. We then really know exactly what's going to prod before it's deployed. Maybe we should just make a ticket to investigate that in the future? What do you think?

I would love to do this. The issue that I see is that the environment variables are built into the image, so the image used to build staging.openebooks.org includes env vars like `qa-circulation.openebooks.us` instead of `circulation.openebooks.us`. The production image also includes additional env vars like the Google Tag Manager ID, which doesn't exist in the staging image.

I'm sure there is a way to feed in different env vars to the Docker image/Next.js app at runtime, so I will create a ticket to investigate. For now, I think we should merge this PR as it is.

PR #565, opened 2023-01-03
OE-770: Create settings page

oliviawongnyc:
What this PR does
- Builds out V1 of Settings page in Hybrid app according to OEH [wireframes](https://www.figma.com/file/AQBIL9ArUYi2BpNsUmD2AW/OEB-wireframes?node-id=1752%3A4732&t=c37BAWfv8yEh8hGA-0)

Notes
- Sign out works, but syncing bookmarks and deleting server data does not yet – still play around with these buttons as I set up alerts/loading modals/spinners!
- Currently, the "Email" button is wrapped in an <a> tag with an href of mailto: info@openebooks.org. If we want more control over the Subject and mail from values, we may need to implement [Cordova's Email Composer](https://ionicframework.com/docs/native/email-composer). If you are using a simulator, you won't be able to see the default `mailto` behavior, but it should work in browser.
- I made use of [<IonNav>](https://ionicframework.com/docs/api/nav) because it seemed to be made for our particular use case (for the T&Cs and the delete server data page), but it may be better to use React Router. LMK what you think!


Toxiapo:
@oliviawongnyc The icons seem to be broken on the preview site. 

Screen Shot 2023-01-04 at 5 21 01 PM

apps/oeh/src/icons/arrowForward.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
samanthaandrews, line 1: I know these are the icons from the wireframes, but I'm thinking for now it might make sense to use Ionic icons until we've created a strategy for how we want to share icons among OEW & OEH. We can add this issue to the "OEH Tech Debt" google doc.

apps/oeh/src/pages/Settings.tsx
```diff
@@ -1,46 +1,41 @@
...
-            <IonToolbar>
-              <IonTitle size="large">Settings</IonTitle>
-            </IonToolbar>
-          </IonHeader>
-          <p>You are signed in</p>
-          <IonButton expand="block" style={{ margin: 16 }} onClick={signOut}>
-            Sign out
-          </IonButton>
-        </IonContent>
+        <IonNav
```
samanthaandrews, line 21: @Toxiapo @kristojorg Curious to hear your thoughts about using this IonNav component. It seems like this might be a good use-case for it, according to the docs, but I'd probably need to dig a little more to be sure https://ionicframework.com/docs/api/nav

I think this means that NavLinks wouldn't link to their own separate route, like we have with `/app/settings` and other pages

samanthaandrews (commented):
Just one comment about icons and a question for Jiayong & Kristo about the IonNav component. This looks really good!!

oliviawongnyc:
> @oliviawongnyc The icons seem to be broken on the preview site.
> 
> Screen Shot 2023-01-04 at 5 21 01 PM

Fixed with Ionic icons!

apps/oeh/ios/App/App.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,88 @@
...
+        "state": {
+          "branch": null,
+          "revision": "ec32d62d412578542c0ffb7a6ce34d3e64b43b94",
+          "version": "0.9.11"
+        }
+      }
+    ]
+  },
+  "version": 1
+}
```
oliviawongnyc, line 88: Not sure what this file is. Anyone else know?

apps/oeh/src/pages/Settings.tsx
oliviawongnyc: Updated!

apps/oeh/ios/App/App.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
kristojorg: I believe this is auto-generated by swift-package-manager, kind of like the package-lock.json file. @ettore can you confirm? Should we merge changes like this where the Package.resolved changes even though we didn't knowingly change anything with SPM?

ettore: yep, this file is automatically generated by SPM, but what's weird is that I already checked in this file a while ago, so git should not report it as a new file. @oliviawongnyc is this branch rebased on top of `staging`?

apps/oeh/src/components/Settings/SignOut.tsx
```diff
@@ -0,0 +1,56 @@
...
+const SignOut: React.FC = () => {
+  const { clearCredentials } = useCredentials();
+  const history = useHistory();
+  const [present] = useIonActionSheet();
+
+  async function signOut() {
+    await clearCredentials();
+    mutate(LOANS_ENDPOINT);
+    history.push("/sign-in");
+  }
```
kristojorg, line 16: I realized we should probably clear all swr data when a user logs out (not just the loans), and I made a ticket for that here: https://jira.nypl.org/browse/OE-817

apps/oeh/src/components/Settings/SignOut.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <IonButton
+      expand="block"
+      style={{ margin: 16 }}
```
kristojorg, line 21: I would also use `ion-padding` className here

kristojorg (changes requested):
Just a couple comments, but this looks amazing!

apps/oeh/src/components/Settings/UserSettings.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <div
+          style={{
+            display: "flex",
+            flexDirection: "row"
+          }}
+        >
+          <div style={{ width: "45px" }}>
+            {isLoading && (
+              <IonSpinner
+                style={{ position: "absolute", top: "23px", left: "12px" }}
```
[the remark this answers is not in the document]
kristojorg: no longer relevant : )

oliviawongnyc:
> @oliviawongnyc The icons seem to be broken on the preview site.
> 
> Screen Shot 2023-01-04 at 5 21 01 PM

Now using Ionic icons!

apps/oeh/src/components/Settings/TermsAndConditions.tsx
```diff
@@ -0,0 +1,45 @@
...
+                key={agreement}
+                routerDirection="forward"
+              >
+                <IonItem>
+                  <IonLabel>{agreement}</IonLabel>
+                  <IonIcon icon={chevronForwardOutline} slot="end" />
+                </IonItem>
+              </IonNavLink>
+            );
+          })}
```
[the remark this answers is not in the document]
oliviawongnyc: Like this approach! Thank you! Added bonus: when you add the `button` prop to the `IonItem`, the icon is automatically added so I could remove my `IonIcon` component.

kristojorg (approved):
Fantastic!

samanthaandrews (approved):

apps/oeh/src/components/Settings/ReportAProblem.tsx
```diff
@@ -0,0 +1,29 @@
...
+} from "@ionic/react";
+
+const ReportAProblem: React.FC = () => {
+  return (
+    <IonCard>
+      <IonCardContent>
+        <IonItem lines="none">
+          <IonIcon icon={helpOutline} slot="start" />
+          <IonLabel>Report a problem</IonLabel>
+          <a href="mailto:info@openebooks.org">
```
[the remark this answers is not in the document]
kristojorg: good call

PR #566, opened 2023-01-04
OE-751/download axisnow

kristojorg:
This PR adds support for downloading axisnow EPUBs. It uses fetch logic from epub-to-webpub but required me to update that package to not use node-fetch. I have made a PR to do that here: https://github.com/NYPL/ePub-to-webpub/pull/22

In the future, it would be good to pull epub-to-webpub into this monorepo and split out some modules and utilities for epubs that can be shared.

I also removed the fulfillment buttons from the my books list because keeping them in sync with the book details page was proving difficult and will require use of context.

Note: I also added absolute import ability to the typescript config so we can `import BookCard from "components/BookCard"` instead of `from "../../components/BookCard"`. Anything within the `src` directory can be resolved that way.


kristojorg:
@ettore If you don't mind, could you check out this PR and try downloading a book and tell me if you think the structure and location appears correct? I currently have the in the `Documents` directory. All the files for the EPUB are downloaded inside a directory with a name that is the url-encoded version of the ID from the OPDS document. I can place it elsewhere though or name the directory differently if you have a better idea.

Thanks!

ettore:
> @ettore If you don't mind, could you check out this PR and try downloading a book and tell me if you think the structure and location appears correct? I currently have the in the `Documents` directory. All the files for the EPUB are downloaded inside a directory with a name that is the url-encoded version of the ID from the OPDS document. I can place it elsewhere though or name the directory differently if you have a better idea.

will do -- the structure is probably ok because I'm guessing you leave it as-is. The one thing I have to verify is having some key component inside the ebook directory because our DRM module has some expectation in that regard, and it wasn't easy to change IIRC: I always forget the details so I'll let you know.

kristojorg:
@ettore sounds good, we can always update in the future too if there are specific changes needed

apps/oeh/src/components/FulfillmentCard.tsx
```diff
@@ -1,20 +1,55 @@
...
+
+  const showProgress =
+    progress.status === "PREPARING" || progress.status === "DOWNLOADING";
+  const progressPercent =
+    progress.status === "DOWNLOADING" ? progress.progress / progress.total : 0;
+
+  const removeDownloadConfirm = () =>
+    presentAlert({
+      header: "Remove Download",
+      // subHeader: '',
```
Toxiapo, line 37: nit: commented out code

apps/oeh/src/lib/download/axisnow.ts
```diff
@@ -0,0 +1,239 @@
+import { FulfillableBook } from "models";
+import { fetchFullfillment } from "../fetching/fetch";
+import { RemoteFetcher, Epub } from "@nypl/epub-to-webpub";
+import { SetDownloadProgress } from "./useProgress";
+import { Directory, Encoding, Filesystem } from "@capacitor/filesystem";
+
+/**
+ * An axisnow book is exploded, and we need to download each file separately into a single directory.
```
Toxiapo, line 8: Thanks for the comments, this is super useful!

apps/oeh/src/lib/download/axisnow.ts
```diff
@@ -0,0 +1,239 @@
...
+ * The directory structure is as follows:
+ * - /
+ *  - encryption.xml
+ *  - META-INF/
+ *    - container.xml (contains a link to the content.opf file).
+ *  - OEBPS/ (or OPS/ for EPUB 3)
+ *    - content.opf
+ *    - ...book content listed in content.opf
```
Toxiapo, line 16: Is there a directory I can explore myself for debugging purposes after we download these contents?

apps/oeh/vite.config.ts
```diff
@@ -31,4 +31,7 @@ export default defineConfig({
     strictPort: true
     // https: true
   }
+  // optimizeDeps: {
+  //   exclude: ["@nypl/epub-to-webpub"]
+  // }
```
Toxiapo, line 36: Do we still need this?

Toxiapo (commented):
The download works great! @kristojorg Just a couple of questions/notes:

1. The preview link for the hybrid app isn't loading for me; It's a blank page, does it work for you?
2. The download works great for ios, but not for browsers. This is fine since it's a mobile app, just wanted to point this out for devs making updates in the future.

apps/oeh/src/lib/download/axisnow.ts
kristojorg: Yea you have to open the documents folder of the simulator... I have been using this app called rocketsim to do that but it's kind of overkill to just do that. There should be some easy way to get the right directory I think but am not totally sure what it is haha

apps/oeh/vite.config.ts
kristojorg: added a comment!

kristojorg:
> The download works great for ios, but not for browsers. This is fine since it's a mobile app, just wanted to point this out for devs making updates in the future.

Yes good point. I have a ticket open in capacitor to fix this browser issue in their plugin

kristojorg:
Okayyy so after much debuggin @Toxiapo I have figured out that the `image-size` dep in `epub-to-webpub` was using a node-specific variable (`process`), which made things break in the browser. Strangely it works fine in development, but did not work when building for production. 

Anyway, I have updated `epub-to-webpub` with a PR here: https://github.com/NYPL/ePub-to-webpub/pull/23. That PR makes a new `RemoteFetcherBrowser` export that doesn't use the `image-size` dep. 

`epub-to-webpub` is now providing services outside of what it is designed for (general purpose EPUB parsing). It would be great to bring this functionality into this monorepo as a few separate packages. How those are divided up is TBD however and will require some thinking about what functionality we need.

kristojorg:
@Toxiapo I am no longer seeing the download error in the browser... can you tell me if you see it still?

apps/oeh/src/lib/download/useDownload.tsx
```diff
@@ -0,0 +1,82 @@
...
+      });
+  }, [book?.id]);
+
+  /**
+   * If the book is not currently downloadable, return a noop download function
+   */
+  if (!book || book.status === "borrowable")
+    return {
+      download: async () => {},
+      removeDownload: async () => {},
```
Toxiapo, line 50: linting: empty async method

apps/oeh/src/lib/download/axisnow.ts
```diff
@@ -0,0 +1,242 @@
+import { FulfillableBook } from "models";
+import { fetchFullfillment } from "lib/fetching/fetch";
```
Toxiapo, line 2: Incorrect import path. Or is it better to make an alias for the 'lib' folder in the config file?

Toxiapo (commented):
This looks good to me! Just a couple of linting warnings to get this thing built!

> epub-to-webpub is now providing services outside of what it is designed for (general purpose EPUB parsing). It would be great to bring this functionality into this monorepo as a few separate packages. 

Agree, we should make a ticket to do that.

apps/oeh/src/lib/download/axisnow.ts
kristojorg: Strange, my editor is not giving me any errors here. I set up a `baseUrl` in the `tsconfig` which I thought allows us to import from folders under `src` using aliases like this. I'll check out what's going wrong

ettore:
I am trying this, here's my thoughts/observations:
- I noticed that the download is saved in a directory with the extension `.epub.adobe` although it's an Axis book: this looks like an oversight? probably it's not causing issues but for clarity it would be good to remove the `.adobe`  suffix.
- I downloaded "The Night Gardener" (by George Pelecanos) and noticed that `OEBPS` and `META-INF` directories are there (good). The web app also saved a `downloadManifest.json` while the native app did not. This is probably ok. 
- However the native app also saves 2 additional files (user_key and license.json) that the web app does not save. (I'm saying this here because anybody can see those files, so I'm not revealing any secrets. I will not discuss what's inside those files though.) These are needed for our decryption code to work. Our Axis module in native OE implements downloading and saves those files, so I figure the web side will need to do the same. Maybe it already does and saves them somewhere else?

kristojorg:
> I noticed that the download is saved in a directory with the extension .epub.adobe although it's an Axis book: this looks like an oversight? probably it's not causing issues but for clarity it would be good to remove the .adobe suffix.

Ah yes that was an oversight, thanks for catching. It probably shouldn't have any extension since it's a directory. 

> I downloaded "The Night Gardener" (by George Pelecanos) and noticed that OEBPS and META-INF directories are there (good). The web app also saved a downloadManifest.json while the native app did not. This is probably ok.

Yea that is a file I saved so that later we can come back and check the integrity of a download (ie. if it finished successfully) If it causes problems for you I can save it elsewhere, it was just convenient to have them colocated.  

> However the native app also saves 2 additional files (user_key and license.json)

Interesting. Let's chat on slack about what is in those files. Maybe it's the same decryption information I was thinking we would pass in the `openBook` call? If so we can either pass it there or save it to a file for you. Maybe a file is better since it persists?

apps/oeh/android/app/capacitor.build.gradle
```diff
@@ -14,7 +14,7 @@ dependencies {
     implementation project(':capacitor-splash-screen')
 
 }
-apply from: "../../../../node_modules/cordova-plugin-ionic/src/android/cordovapluginionic.gradle"
```
kristojorg, line 17: I removed this package for the time being from npm and it removed it from both native projects.

apps/oeh/jest.config.ts
```diff
@@ -21,7 +21,7 @@ const jestconfig: Config = {
   // Jest doesn't fully support ESM yet, but Ionic ships ESM.
   // See here: https://ionicframework.com/docs/intro/upgrading-to-ionic-6#testing
   transformIgnorePatterns: [
-    "/node_modules/(?!@ionic/core|@ionic/react|@ionic/react-router|@stencil/core|ionicons)"
+    "/node_modules/(?!@ionic/core|@ionic/react|@ionic/react-router|@stencil/core|ionicons|@nypl/epub-to-webpub)"
```
kristojorg, line 24: We want jest to transform epub-to-webpub in this case since we are importing the raw typescript files

ettore:
> Ah yes that was an oversight, thanks for catching. It probably shouldn't have any extension since it's a directory.

Can we keep the `.epub` actually? it would make it simpler to identify content for debugging purposes.

> The web app also saved a downloadManifest.json while the native app did not. This is probably ok.
> 
> Yea that is a file I saved so that later we can come back and check the integrity of a download (ie. if it finished successfully) If it causes problems for you I can save it elsewhere, it was just convenient to have them colocated.

yep, that's totally fine... I don't think it will cause any problems.

> > However the native app also saves 2 additional files (user_key and license.json)
> 
> Interesting. Let's chat on slack about what is in those files. Maybe it's the same decryption information I was thinking we would pass in the `openBook` call? If so we can either pass it there or save it to a file for you. Maybe a file is better since it persists?

Saving it to 2 files named like that would be simplest and it will not be a breaking change for SimplyE. That's how our AxisDRM code expects them and if I remember well, parametrizing them was not trivial, sadly. but yes we can chat about what they actually are.

ettore:
> > Ah yes that was an oversight, thanks for catching. It probably shouldn't have any extension since it's a directory.
> 
> Can we keep the `.epub` extension actually? it would make it simpler to identify the type of content for debugging purposes.

Actually, we must keep the `.epub` extension. R2 uses it as a hint to determine the media type if anything else fails.

kristojorg:
@ettore as discussed on slack, let's wait for milestone 2 to flesh out the details of how we save the decryption info and name the folder. Both are easy to change on our side and will just depend on the needs on the native side

apps/oeh/src/lib/download/useDownload.tsx
```diff
@@ -0,0 +1,86 @@
+/**
+ * We need an api that:
+ *  - Takes a loans feed
+ *  - clears out any downloads not found in the loans feed
+ *  - offers an api to download new loans
+ *  - offers an api to delete downloads
+ *  - indicates download progress when downloading
```
samanthaandrews, line 7: Can we add these as Jira ticket(s)? Either in the `Hybrid App Beta` epic or the `OEH Tech Debt` one.

apps/oeh/src/lib/download/axisnow.ts
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * TO DO
+ *
+ * - When you download from the book details page, it updates its own state, but the my books
+ *   page doesn't update until you refresh the page because it doesn't rerender and re-check if the
+ *   book is downloaded. We could put the download progress in context so that all pages can use a single
+ *   progress state.
+ * - Debug why sometimes the download stops halfway through (in browser).
+ * - Fix clicking fulfillment card in list
+ */
```
samanthaandrews, line 68: Can we add these as Jira ticket(s)? Either in the Hybrid App Beta epic or the OEH Tech Debt one.

samanthaandrews:
> In the future, it would be good to pull epub-to-webpub into this monorepo and split out some modules and utilities for epubs that can be shared.

I'll schedule time to discuss this with Crystal and Jackie, since DRB is the other project that uses `epub-to-webpub`

apps/oeh/src/lib/download/axisnow.ts
kristojorg: done! [OE-815](https://jira.nypl.org/browse/OE-815) and [OE-816](https://jira.nypl.org/browse/OE-816)

apps/oeh/src/lib/download/useDownload.tsx
kristojorg: Yep, added this as part of [OE-815](https://jira.nypl.org/browse/OE-815)

samanthaandrews (approved):
This looks good to me. Are we ready to merge it?

ettore:
> This looks good to me. Are we ready to merge it?

I've not approved only because I am not familiar with JS/web tech, but no opposition from me.

PR #567, opened 2023-01-05
PJR-137 Setup R2 dependency

ettore:
This makes the app compile R2 and link it to the final executable. 

I added R2 as a remote SPM package for now, since it's going to be a while until we actually need to modify that codebase. If we need to, we can change this to a git submodule. 

This also adds an empty "NYPLReader" SPM package. This will be the actual eReader package. Currently it only includes references to the R2 packages to verify linking works fine. I verified R2 is being built and NYPLReader's unit tests are being run.


apps/oeh/ios/App/Podfile
```diff
@@ -15,13 +15,16 @@ def capacitor_pods
   pod 'CapacitorFilesystem', :path => '../../../../node_modules/@capacitor/filesystem'
   pod 'CapacitorPreferences', :path => '../../node_modules/@capacitor/preferences'
   pod 'CapacitorSplashScreen', :path => '../../../../node_modules/@capacitor/splash-screen'
+  pod 'CordovaPlugins', :path => '../capacitor-cordova-ios-plugins'
```
ettore, line 18: I think I had this removed in the past, but every time I run `npm run build`/`npm run sync:static` this is being added back, so I figure it should stay.

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -402,7 +407,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
 				SWIFT_COMPILATION_MODE = wholemodule;
-				SWIFT_OPTIMIZATION_LEVEL = "-O";
+				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
```
ettore, line 410: this is an Xcode recommended setting for release builds.

apps/oeh/ios/App/Podfile
kristojorg: I finally removed the npm package I had installed that is triggering it to be added back. Not sure if we will eventually need it or not, but for now it's safe to ignore

apps/oeh/ios/App/Podfile
```diff
@@ -15,13 +15,16 @@ def capacitor_pods
...
   capacitor_pods
   # Add your Pods here
 end
 
-# post_install do |installer|
-#   assertDeploymentTarget(installer)
-# end
+post_install do |installer|
+  installer.pods_project.build_configurations.each do |config|
+    config.build_settings['CLANG_WARN_DOCUMENTATION_COMMENTS'] = ['NO']
```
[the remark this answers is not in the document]
kristojorg: 👍 gratzie

kristojorg (approved):

apps/oeh/ios/App/Podfile
ettore: sounds good -- I think what we'll do is update the podfile every time npm changes it, so that works.

PR #569, opened 2023-01-09
OE-772: router rendering issue

Toxiapo:
This PR fixes the 404 error issue on page refresh when the URL contains special characters.


apps/oeh/src/pages/BookDetail.tsx
```diff
@@ -45,11 +45,7 @@ const BookDetails: React.FC<BookDetailProps> = ({ match }) => {
     event.detail.complete();
   }
 
-  if (!credentials) {
-    return <Redirect to="/sign-in" />;
-  }
```
Toxiapo, line 50: Moving this after the "loading" check otherwise, it will always go to the sign-in page since `credentials` by default is undefined.

kristojorg (approved):
This is great! So glad you fixed this haha. Thanks!

PR #570, opened 2023-01-10
PJR-135 Setup scaffolding for R2 iOS

ettore:
I copied most of this code from the [R2 test app](https://github.com/readium/swift-toolkit/tree/8b95a2b88066097559db00972217560daebd38ce), sorry for the large PR. There's probably more than what we need now, but somehow this was faster to do, and it seemed safer to match the TestApp as much as possible. There's also some commented out code for the same reason.

This should provide some infrastructure to work on actually opening a book. After that, we can optimize the code as we want.


ErnestFan (approved):

ettore:
forced push just to rebase branch on top of staging.

PR #572, opened 2023-01-12
Chore: Add unoptimized parameter to `next.config.mjs` images

samanthaandrews:
Vercel bills us for optimized images, which is something that is not in NYPL's budget, so we are disabling it for now.


kristojorg:
@samanthaandrews we are only using a couple hundred of the 5k free images vercel provides. I'm fine to disable this, but I hope it doesn't also disable it on our own instance. I think we should investigate a little bit what is going on here

samanthaandrews:
> @samanthaandrews we are only using a couple hundred of the 5k free images vercel provides. I'm fine to disable this, but I hope it doesn't also disable it on our own instance. I think we should investigate a little bit what is going on here

I am going to email Vercel and ask what exactly the `unoptimized` parameter is doing. I'll move this PR to draft state until they respond. I just didn't want to forget to do this so I made the PR quickly yesterday to remind me.

PR #573, opened 2023-01-13
PJR-131 Open ePub in iOS native reader

ettore:
- add code to open book in R2. Most of this is taken from official R2 test app.
- make sure to hide web navigation when opening R2 and show native nav instead.; vice versa in the opposite direction.

notes:
- JS does not call the plugin's `openBook` function yet. 
- we might be able to replace some delegates with Combine in the future. This is just a first iteration to achieve the "open book" Milestone 1.


apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin+OpenBook.swift
```diff
@@ -0,0 +1,67 @@
...
+            } receiveValue: { [weak self] pub in
+                self?.readerModule?.presentPublication(publication: pub,
+                                                       in: navVC,
+                                                       completion: done)
+            }
+            .store(in: &subscriptions)
+    }
+
+    private func extractBookURL(from call: CAPPluginCall) -> URL? {
+//        let docs = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
```
ettore, line 53: once JS passes the bookURL, this will no longer be needed.

ErnestFan (approved):

PR #574, opened 2023-01-13
OE-776: Add collection page and view more button

Toxiapo:
This PR adds the view more button to the carousel sections allowing us to view child lanes.

Updates to the routings:

1. I added a new route, `/app/browse/collection/:collectionUrl` to serve the collection page
2. Renamed `all-book` to `Discover`.


apps/oeh/src/App.tsx
```diff
@@ -35,13 +35,17 @@ import "@ionic/react/css/display.css";
 
 /* Theme variables */
 import "./theme/variables.css";
+/* Global Styles */
+import "./theme/global.css";
```
Toxiapo, line 39: I created this global style just so I can avoid adding `ion-text-center` to multiple components, we can put our global CSS here until we decide which styling route we should go with.

apps/oeh/src/App.tsx
```diff
@@ -90,8 +94,13 @@ const Tabs = () => {
         <Route exact path="/app/browse" component={Browse} />
         <Route path="/app/browse/book/:bookUrl" component={BookDetails} />
         <Route
+          exact
```
Toxiapo, line 97: This is needed for the dynamic route mainly because the collection page shares the `/app/browse` path with the browse page.

apps/oeh/src/App.tsx
kristojorg: I think we've decided right? AFAIK we are using CSS files and the [ionic css utilities](https://ionicframework.com/docs/layout/css-utilities). Either way I think this is okay

kristojorg: Although maybe the utilities have this built in?

kristojorg:
> the  /app/browse/collection/book/:bookUrl route renders the book detail page from the collection page.

Is this necessary? We have the `/app/browse/book/:bookUrl` route already which seems like the same thing no? I think it's okay to route from `/app/browse/collection/:collectionUrl` to `/app/browse/book/:bookUrl`, just not to a different tab like `/app/loans`

apps/oeh/src/App.tsx
kristojorg: I guess we could set up some css files for things like global changes and updates to the ionic theme

apps/oeh/src/App.tsx
kristojorg: Strange. I don't have a problem with it but it doesn't seem necessary, then `/app/browse/book/:bookUrl` also shares that but it doesn't have `exact`

apps/oeh/src/pages/Collection.tsx
```diff
@@ -0,0 +1,94 @@
...
+              />
+            ))}
+          </>
+        )}
+      </IonContent>
+    </IonPage>
+  );
+};
+
+function getRoutePath(path: string): string {
```
kristojorg, line 89: A comment would be helpful here

kristojorg (commented):
Just some questions

apps/oeh/src/theme/global.css
```diff
@@ -0,0 +1,3 @@
+ion-title {
+    text-align: center;
```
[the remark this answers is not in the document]
Toxiapo: The default align was align-left. 

Ah okay, so it's pushed toward the right because of the back button, I will update the css

apps/oeh/src/App.tsx
Toxiapo: The problem was that if I were on the `browse` page, and clicked on a book, it would render the detail page. But if I go to a sub-lane (collection page) and visit the book detail page from there, I can see the URL updates, but it doesn't render the content for me. I've tried many things, but only this method worked for me..

apps/oeh/src/pages/Collection.tsx
Toxiapo: This is to extract the collection URL and remove the ID part, but I can probably just remove this and hardcode that URL since it should always render the collection page.

apps/oeh/src/App.tsx
Toxiapo: Agree. I can do some research on how to utilize ionic CSS best. Maybe an `override.css` is a better fit if we are going to manipulate their default CSS

apps/oeh/src/theme/global.css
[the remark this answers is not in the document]
kristojorg: I think honestly I would leave it to the platform defaults since I believe it is specific to each platform

apps/oeh/src/App.tsx
kristojorg: Strange, but sounds good! Thanks for the explanation : )

apps/oeh/src/theme/global.css
```diff
@@ -1,3 +1,8 @@
 ion-title {
-    text-align: center;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  text-align: center;
```
kristojorg, line 7: At the moment, I think we should leave this as-is for a few reasons:
1. We aren't trying to match designs exactly yet, just get wireframes in place
2. This will update all titles in the app
3. This is applied for both platforms, and I believe the behavior differs for them currently

apps/oeh/src/pages/Browse.tsx
```diff
@@ -48,24 +48,26 @@ const Browse: React.FC<RouteComponentProps> = ({ match }) => {
...
           <IonRefresherContent></IonRefresherContent>
         </IonRefresher>
         {data.lanes.map(lane => (
           <CarouselSection
+            lane={lane}
             key={lane.url}
             books={lane.books}
             title={lane.title}
-            baseUrl={match.url}
+            baseUrl={match.url + "/collection"}
```
kristojorg, line 69: It looks like the `baseUrl` is now always `/app/browse/collection` right? Can we remove it?

kristojorg (changes requested):
Just a few questions

apps/oeh/src/pages/Collection.tsx
```diff
@@ -0,0 +1,88 @@
...
+import BookListView from "../components/Collection/BookListView";
+
+type CollectionProps = RouteComponentProps<{
+  collectionUrl: string;
+}>;
+const Collection: React.FC<CollectionProps> = ({ match }) => {
+  const collectionUrl = decodeURIComponent(match.params.collectionUrl);
+  const { data: collection, mutate } = useCollection(collectionUrl);
+
+  async function doRefresh(event: CustomEvent<RefresherEventDetail>) {
```
samanthaandrews, line 26: might want to rename this or add a comment specifying what we are refreshing here

apps/oeh/src/pages/Browse.tsx
Toxiapo: We do need it because that's being used for the `view more` link. Do you think it makes sense to just remove the /app/browse route and replace it with just the /app/collection route then?

apps/oeh/src/theme/global.css
Toxiapo: Ok, yea that makes sense! I missed the cross-platform style. I am just going to remove this for this PR.

apps/oeh/src/pages/Browse.tsx
Toxiapo: OK, I updated the route to use `/app/collection` as our main entry page, and removed `/app/browse`. I think this may be easier for our mental model since this would be a similar setup as OEW. Let me know what you think!

samanthaandrews:
Let's try to merge this one today or tomorrow if possible!

apps/oeh/src/pages/Browse.tsx
kristojorg: That is a great idea, thanks!

kristojorg (commented):
This looks great, just the one question about the url segments. @Toxiapo and I discussed this on slack but I think it makes sense to have a url segment named per tab, and then sub segments describe the views within the tab. So we would have:

- /app/browse
  - /collection/:collectionUrl
  - /book/:bookUrl
- /app/settings
- /app/my-books
  - /book/:bookUrl
- /app/read

We may want to change /browse to /all-books if we decide that is what we are going to name that tab. That way all the tabs will have a url segment that match

kristojorg (approved):
Looks good to me!

PR #575, opened 2023-01-17
PJR-130 Add unencrypted ePubs

ErnestFan:
Add zipped and exploded ePubs
Add ePubs to iOS project

note:
The reader would stuck on the loading screen when tested on iOS 14.0 simulator. But it works without issue on iOS 15.5 simulator and iOS 15.6 real device. @ettore Which iOS version were you testing with?


apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -251,10 +278,10 @@
 			buildActionMask = 2147483647;
 			files = (
 			);
-			inputPaths = (
+			inputFileListPaths = (
```
ettore, line 281: I'm guessing this was Xcode trying to be "helpful"? 🙄

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin+OpenBook.swift
```diff
@@ -52,6 +52,10 @@ extension ReaderPlugin {
     private func extractBookURL(from call: CAPPluginCall) -> URL? {
 //        let docs = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
 //        return docs?.appendingPathComponent("DivinaCommedia.epub")
+      // Zipped ePub for demo
+//      return Bundle.main.url(forResource: "Metamorphosis-jackson", withExtension: "epub")
```
ettore, line 56: until we have the URL being passed by the web side, I think we should just do this, i.e. open the hard coded ebook and comment out the code I wrote below, to demo/work more easily on the reader. What do you think?

ettore (commented):
this looks good to me. Now that we know a little bit more, I'm wondering if we actually need the exploded epub: maybe the single-file packaged epub is enough. What do you think @kristojorg @roywatson @MalcolmMcFly @samanthaandrews ?

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
ErnestFan: Yeah it's Xcode...

apps/oeh/ios/App/App/ReaderPlugin/ReaderPlugin+OpenBook.swift
ErnestFan: Sure, that works for me. I will make the changes after other devs respond to your question above.

ErnestFan:
> > Add zipped and exploded ePubs Add ePubs to iOS project
> > note: The reader would stuck on the loading screen when tested on iOS 14.0 simulator. But it works without issue on iOS 15.5 simulator and iOS 15.6 real device. @ettore Which iOS version were you testing with?
> 
> @ErnestFan I tested on iOS 15.5 simulator mainly. Can you file a bug ticket for this? This is something we could investigate next sprint. It would actually be good to test on iOS 13 and 16 as well.

Filed a ticket, see [link](https://jira.nypl.org/browse/PJR-149).

ettore (approved):

PR #576, opened 2023-01-17
added jcenter

MalcolmMcFly:
**What does this do?**
- [x] Disabled Android Readium dependencies to unblock @kristojorg 
- [x] Updated simplified Readme with accurate Android build steps.

**How should this be tested?**
The product should build as usual.


roywatson:
We don't need to comment out the readium dependancies to eliminate the duplicate classes issue. If we comment out/remove the above dependancy:

implementation "org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.20.0"

That should eliminate the duplicate classes error and allow the readium depends to stand.

kristojorg (approved):
Looks good to me

kristojorg:
FYI I merged this as the failures were blocking the ios release process

PR #578, opened 2023-01-18
OE-759/http-v2

kristojorg:
This PR allows the android emulator to accept the http traffic used by the live development server. This means we can now use `npm run android:live` to view the app in an android emulator with live reloading. Note that this can be done simultaneously with the ios simulator, and thus we can be previewing both platforms simulaneously during development.


roywatson (approved):
It runs and looks good.

PR #582, opened 2023-01-20
Chore: minor iOS cleanup

ettore:
@ErnestFan : somehow the epubs were showing up red when I opened the project in Xcode. Not sure why because the build was still succeeding. I verified the build still succeeds and the epub is being opened. 🤷 

@kristojorg : after doing a pull from staging + `npm install` + `npm run build`,  I noticed that `npm run sync:static` removed the CordovaPlugins pod from the Podfile. So I am removing it here.


ErnestFan (approved):

kristojorg (approved):
This looks good to me. I'm not sure why that cordova plugin keeps being added and removed to be honest... I thought I removed it previously but either way this is fine!

PR #586, opened 2023-01-24
Release OE iOS 3.0.0 for Milestone 1 demo

samanthaandrews:


ettore:
🤘🤘

ettore:
@samanthaandrews we should probably get these release PR finalized and merged into `main` sooner rather than later (ideally right after a release goes live), to avoid conflicts. Not sure if we are waiting on something else though.

samanthaandrews:
@ettore yeah i originally thought we wouldn't merge these until we actually make the 3.0.0 release to users, but i think you're right we should merge this to `main` during this sprint. do you think we should tag it with something like `oe-ios-3.0.0-alpha`?

ettore:
> @ettore yeah i originally thought we wouldn't merge these until we actually make the 3.0.0 release to users, but i think you're right we should merge this to `main` during this sprint. do you think we should tag it with something like `oe-ios-3.0.0-alpha`?

yeah, I think we should have named the branch `release/oe-ios-3.0.0-alpha` since this was not intended to be released to actual users. But in all other aspects we treated this as a release, so yeah I think we should merge these. Your tag name suggestion sounds great to me.

kristojorg:
We can also change the name of the branch to the one @ettore suggested. I don't think that's a bad idea

ettore (approved):

PR #589, opened 2023-01-25
Release: Android v3.0.0

kristojorg:


apps/oeh/android/app/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
```
MalcolmMcFly, line 10: I would be surprised if Open eBooks has not been submitted with `1` as the `versionCode` before.

kristojorg: This will be updated by appflow when they build the app. I will have to check what the most recent version code was though and set that as the starting point

PR #591, opened 2023-01-25
PJR-149 Update iOS target setting to avoid native reader stuck on loading

ErnestFan:
**What's this do?**
Update iOS target setting to avoid native reader stuck on loading

**Why are we doing this? (w/ JIRA link if applicable)**
[PJR-149](https://jira.nypl.org/browse/PJR-149)

**How should this be tested? / Do these changes have associated tests?**
Login
Press "Open Native" to open reader
Demo book should be readable

**Did someone actually run this code to verify it works?**
@ErnestFan tested on iOS 14-16, unable to test iOS 13 on MacOS Ventura

Note:
@ettore can you test this on iOS 13? I am unable to do so on MacOS Ventura
This also seems to fix the book cover loading issue on catalog page


apps/oeh/ios/App/App/Info.plist
```diff
@@ -2,32 +2,37 @@
...
-	<string>6.0</string>
+	<key>CFBundleSpokenName</key>
+	<string>Open E-Books</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<false/>
+	<key>LSRequiresIPhoneOS</key>
+	<true/>
+	<key>NSAppTransportSecurity</key>
```
ErnestFan, line 29: @ettore Xcode decides to rearrange the properties every time I tried to modify this plist. Only line 29 - 33 contains the actual changes.

ettore:
@samanthaandrews this fix addresses the issue of the reader not opening up on iOS 14. Would you rather not merge this before the demo? It seems pretty safe to me. 

if we merge it, will it kick off a new build to TestFlight?

ettore (approved):
it looks good to me but let's wait to merge until @samanthaandrews chimes in.

samanthaandrews:
I think it's ok to merge. To update the release deployment, we will then need to merge `main` into the `release/oe-ios-3.0.0` branch. Let's aim to do this tomorrow morning at the latest.

samanthaandrews:
Oh! I see this is a PR into the release branch. I think this is ok too.

samanthaandrews:
And yes, in this case, any new push to a `release/oe-ios-*` branch will trigger an Appflow build.

samanthaandrews (approved):

PR #593, opened 2023-01-25
OE-845: Simplify open book flow

kristojorg:
This PR simply removes the old Bookmarks UI and adds a simple button to open the native readers. Just thinking it will make the demo a bit smoother.


Toxiapo (approved):
I pulled down the branch and was able to run the reader on the IOS emulator

PR #596, opened 2023-02-01
OE-790: [OEH] Implement live updates

samanthaandrews:
This PR implements Appflow's live updates feature in the OE hybrid project


apps/oeh/android/app/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,12 @@
     <string name="package_name">org.openebooks.app</string>
     <string name="custom_url_scheme">org.openebooks.app</string>
     <string name="pseudo_book">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</string>
+    <!-- Live Updates Config -->
```
samanthaandrews, line 8: @roywatson @MalcolmMcFly These are the values that I added in order to implement [Appflow's live updates feature](https://ionic.io/appflow/live-updates)

apps/oeh/ios/App/App/Info.plist
```diff
@@ -30,6 +30,20 @@
 	<string>en</string>
 	<key>UIMainStoryboardFile</key>
 	<string>Main</string>
+	<!-- Live Updates Config -->
```
samanthaandrews, line 33: @ErnestFan @ettore These are the values that I added in order to implement [Appflow's live updates feature](https://ionic.io/appflow/live-updates)

apps/oeh/src/App.tsx
```diff
@@ -107,7 +107,7 @@ const Tabs = () => {
           <IonLabel>My books</IonLabel>
         </IonTabButton>
         <IonTabButton tab="settings" href="/app/settings">
-          <IonIcon icon={settings} />
+          <IonIcon icon={sad} />
```
samanthaandrews, line 110: I just changed this for testing, to make sure I could see something change without re-copying the web assets to the native build.

apps/oeh/capacitor.config.ts
```diff
@@ -11,6 +11,11 @@ const config: CapacitorConfig = {
   plugins: {
     SplashScreen: {
       launchShowDuration: 0
+    },
+    LiveUpdates: {
+      appId: "4bad0095",
+      channel: "3.0.0",
```
samanthaandrews, line 17: I should probably change this to `production`? I set it to 3.0.0 just for testing.

apps/oeh/ios/App/App/Info.plist
ettore: IIRC this feature is automatically turned off by Capacitor for release builds, correct? if so, is it being turned off at compile time or at run time?

apps/oeh/ios/App/Podfile
```diff
@@ -15,6 +15,7 @@ def capacitor_pods
   pod 'CapacitorFilesystem', :path => '../../../../node_modules/@capacitor/filesystem'
   pod 'CapacitorPreferences', :path => '../../node_modules/@capacitor/preferences'
   pod 'CapacitorSplashScreen', :path => '../../../../node_modules/@capacitor/splash-screen'
+  pod 'CordovaPlugins', :path => '../capacitor-cordova-ios-plugins'
```
ettore, line 18: lol I like how this keeps coming back and removed every other PR....do you think now it's _actually_ required by Live Updates?

ettore (commented):
looks fine to me, just wanted to understand better how this works. Also, why Cordova keeps on coming back :)

apps/oeh/package.json
```diff
@@ -37,6 +37,7 @@
     "@nypl/epub-to-webpub": "^2.1.0",
     "@trapezedev/configure": "^6.0.7",
     "constants": "*",
+    "cordova-plugin-ionic": "5.5.3",
```
samanthaandrews, line 40: @ettore I do think that the cordova plugin is _actually_ required now. The capacitor docs instructed me to install this to our package.json for setting up Live Updates.

apps/oeh/ios/App/Podfile
samanthaandrews: Answered in a comment below! Yes I do think it is required now

apps/oeh/ios/App/App/Info.plist
samanthaandrews: We can set it to be turned off by setting `DisableDeploy` in the `config.xml` file, but now that I look at it I only see a `config.xml` file in the Android project. We do need to figure out how to disable live updates in development, because the docs say it can interfere with hot reloading. Can you look at what their docs say and let me know if it is clearer to you? https://ionic.io/docs/appflow/quickstart/deploy#disabling-deploy-for-development

From the docs about how to actually deploy a live update, we need to select a commit that we want to build, make a web build from that specific commit, and then assign it to a channel. For testing, I used channel 3.0.0 but I think we would actually want to set it to `production` and use that channel.

TBH how we manage these live updates is not very clear to me. Maybe we need to chat about it and discuss this PR together to get on the same page.

samanthaandrews: I scheduled time for Feb 14. I can show you all how I made a web build and assigned it to a "live updates" channel.

apps/oeh/package.json
kristojorg: I had installed this previously and then not used it. Not sure why it was so hard to remove before. Was pretty annoying haha

apps/oeh/capacitor.config.ts
kristojorg: This deserves some thought (at least before we actually go to production). We can use the channel version number to distinguish which native versions can accept a given live update. We should probably also think about how we version the app with live updates. Do live updates get a new version of their own (3.0.0-live.1)?

apps/oeh/ios/App/App/Info.plist
kristojorg: Yep I think chatting about this is a good idea

kristojorg: And maybe making a ticket for someone to write up a document about possible strategies and things to think about like:

1. How do we use channels and how do we manage which native version can get a live update?
2. How do we version live updates and native releases?

kristojorg (approved):
Looks good to merge, I think we should make a follow up ticket to create a suggested strategy document to circulate ahead of our meeting

apps/oeh/ios/App/App/Info.plist
ettore: I took a look at the docs but I guess I'm confused by what we mean with live updates or in what context we want that. I thought that was mainly for development purposes, and definitely not for production since we can't deploy an app that updates its code thru the Apple App Store. But it seems like you guys want to turn live updates OFF for development. Do we then want it enabled for QA purposes?

apps/oeh/capacitor.config.ts
samanthaandrews: This is definitely something I want to clarify bc I don't know how to manage it currently. I'm thinking I'd like to wait to merge this PR actually until we decide how to manage the live update channels. I'm going to move this PR to draft state until we write the document you suggested.

apps/oeh/ios/App/App/Info.plist
ettore: definitely agree on the document suggestion.... I'd love to understand how this feature works exactly and what we can / cannot do as far as pushing code. I also agree on the thought of using it just for critical bug fixes.

apps/oeh/ios/App/App/Info.plist
```diff
@@ -35,6 +35,20 @@
 	<string>LaunchScreen</string>
 	<key>UIMainStoryboardFile</key>
 	<string>Main</string>
+	<!-- Live Updates Config -->
+	<key>IonAppId</key>
+	<string>4bad0095</string>
+	<key>IonChannelName</key>
+	<string>oe-ios-3.0.0</string>
```
ettore, line 42: would this still work once we update to future versions (say, 3.1) or will we need to keep updating this?

samanthaandrews:
@ettore I won't merge this until we talk about this as a team on March 7. This is the strategy doc I've written and it includes many of the same questions you're asking: https://docs.google.com/document/d/1g6wGivowtIzHZlzou1yJbJUUH2MeNmJVyDVTID888ww/edit#heading=h.m0gaeskz5zeu

Once we discuss as a team, I'll put the info in that strategy doc into the ereading-clients README

PR #597, opened 2023-02-01
Upgrades web-reader package to ^4.0.0

oliviawongnyc:
- Updates @nypl/web-reader version from ^3.2.5 to ^4.0.0 to include latest updates to Settings Panel

One concern I noted:

I noticed that clicking "Scrolling" in the Settings Panel does not work as expected in OEW. If "Scrolling" is selected, the book still appears in a paginated fashion. I moved forward with this change however, because I don't think it has anything to do with the new web-reader version. For one thing, it works fine in the example app in the web-reader repo.  And two, it seems to be an existing issue on openebooks.us staging and prod. Maybe it's a known issue?


apps/oew/package.json
```diff
@@ -39,7 +39,7 @@
     "@next/mdx": "^12.3.1",
     "@nypl/dgx-svg-icons": "^0.3.12",
     "@nypl/epub-to-webpub": "^2.1.0",
-    "@nypl/web-reader": "^3.2.5",
+    "@nypl/web-reader": "^4.0.0",
```
Toxiapo, line 42: I am curious why we are updating this to 4.0.0 instead of 4.2.0?

Toxiapo (changes requested):
Hmm, the scrolling function seems fine to me. I tested the `Lost Brother -  Dixon, Franklin W.` book.

apps/oew/package.json
oliviawongnyc: With the caret, this installs 4.2.0. But I am open to changing it to be more explicit if people prefer!

oliviawongnyc: My thinking was that the only guaranteed number when using the caret is the major version. If we are explicit about ^4.2.0, the dev doesn't know if the minor version is what is installed or if that is outdated so why include it. But I def don't feel strongly about it, and happy to change if it's clearer to the team!

oliviawongnyc:
> Hmm, the scrolling function seems fine to me. I tested the `Lost Brother - Dixon, Franklin W.` book.

@Toxiapo, interesting! This is for me on Chrome Version 109.0.5414.119:

https://user-images.githubusercontent.com/57868332/216354507-5ff7f33c-c3b9-4f91-a236-f6ea7a0b0a23.mov

Toxiapo:
> > Hmm, the scrolling function seems fine to me. I tested the `Lost Brother - Dixon, Franklin W.` book.
> 
> @Toxiapo, interesting! This is for me on Chrome Version 109.0.5414.119:
> 
>  Screen.Recording.2023-02-02.at.9.38.43.AM.mov

Ah okay. I believe the pages you were viewing were meant to be a single-page resource that's why you can't scroll. It's only scrollable when a resource was broken down into smaller frames and can't be fully displayed by the current viewport.

Toxiapo (approved):
This looks good to me!

oliviawongnyc:
> > > Hmm, the scrolling function seems fine to me. I tested the `Lost Brother - Dixon, Franklin W.` book.
> > 
> > 
> > @Toxiapo, interesting! This is for me on Chrome Version 109.0.5414.119:
> > Screen.Recording.2023-02-02.at.9.38.43.AM.mov
> 
> Ah okay. I believe the pages you were viewing were meant to be a single-page resource that's why you can't scroll. It's only scrollable when a resource was broken down into smaller frames and can't be fully displayed by the current viewport.

I see, so if I had kept going, I would have been able to see the pagination? I wonder if it would make sense to disable the pagination button unless it was available.

samanthaandrews:
@oliviawongnyc The behavior you described is intentional, but I agree it is not intuitive. Basically, even in scrolling mode, single-page resources are still paginated and display the "prev" and "next" buttons. Once you navigate to a longer chapter, you will be able to scroll. Then, once you reach the end of that chapter, you have to paginate to the next chapter.

For now, I think this is good to go to Jennifer for VQA (maybe you've already done this!). Addressing the scrolling vs pagination UX can be addressed in a separate initiative.

kristojorg (approved):
Sweet!

PR #598, opened 2023-02-03
OE-782/update-routing-logic

kristojorg:
This PR adds a hook called `useTabPath` that returns the path of the tab that you are currently on so we can construct relative urls within a specific tab. This allows the same <BookCard> for example to construct a path to the book details page in the my-books tab or in the discover tab. 

The benefit of this is that we no longer need to pass a `baseUrl` prop through all our components. The one danger is that it has some hard coding of route names (specifically it expects the tabs to be nested under the /app route). It will throw an error if used not in the right way though, so I think it's safe as we would quickly notice during development if we tried to use it incorrectly.


Toxiapo (approved):
This is much cleaner, thank you!

PR #601, opened 2023-02-06
Xcode 14.2 upgrade + script to rebuild/static-sync all web code

ettore:
Xcode 14.2 highlighted a new warning related to the Podfile copy phase, which has no inputs/outputs but still it was set to run conditionally based on input/outputs. I think we want to run that phase for every build.

Also adding a script to ease building / redeploying of web code consistently with node dependencies, in the context of mobile development. I added this in a new `scripts` directory in `apps/oeh`: @MalcolmMcFly @roywatson maybe this is useful to you as well.


ErnestFan (approved):

kristojorg (approved):
Looks good to me!

PR #602, opened 2023-02-07
Adds lost lane to 404 page

oliviawongnyc:
[OE-851](https://jira.nypl.org/browse/OE-851): Adds `Are you Lost?` lane to `404` page


samanthaandrews (approved):
LGTM

apps/oew/src/components/chakra-ui/fetching.ts
```diff
@@ -59,6 +59,16 @@ export const fetchCuratedCollections = async () => {
...
 };
 
+/**
+ * Fetches the lane data for the 404 page.
+ * This is a list of books about being lost.
+ */
+export const fetchLostCollection = async () => {
+  const lostBooksUrl = makeCMUrl(lostCollectionEndpoint);
+
+  return await Promise.resolve(fetchCollectionBooks(lostBooksUrl));
```
kristojorg, line 69: Just curious about this use of Promise.resolve. I think you should be able to return `fetchCollectionBooks(lostBooksUrl)` without that

kristojorg (approved):
Looks good! Just the one little comment

apps/oew/src/components/chakra-ui/fetching.ts
oliviawongnyc: Oops missed this in the flurry before my vacation. I removed and it still seems to work. I had originally added to mimic the fetchCuratedCollections function above.

kristojorg: Ahh I see I see. In that case, `Promise.all` is used to create one promise that waits for all of the individual promises passed in to it (and `fetchCollectionBooks` returns one)

samanthaandrews:
If this has been approved by QA, let's merge!

PR #603, opened 2023-02-08
OE-699: Update book fetching logic to not check  against loans

Toxiapo:
This PR removes the loans fetching in the fulfillment logic since the new update from the backend, the book endpoint will now no longer sends a caching header


apps/oeh/src/hooks/useBook.ts
```diff
@@ -17,7 +17,7 @@ export function useBook(url: string) {
   const [isBorrowingOrReturning, setIsBorrowingOrReturning] = useState(false);
   const { credentials } = useCredentials();
   const swr = useSWR<OpenEBook, Error>(
-    credentials?.token ? [url, credentials?.token] : null,
+    url && credentials?.token ? [url, credentials?.token] : null,
```
kristojorg, line 20: In what situation would url be undefined?

apps/oew/src/hooks/useBorrowOrReturn.ts
```diff
@@ -50,18 +49,21 @@ export default function useBorrowOrReturn(fulfillmentType: FulfillmentType) {
...
       });
       setLoading(false);
-      // optimistically update our local loans and revalidate
-      mutateLoans([...(loans ?? []), book]);
 
+      const loansKey = [shelfUrl, credentials?.token, credentials?.methodType];
+
+      const loans: OpenEBook[] = cache.get(unstable_serialize(loansKey)) ?? [];
+
+      mutate(loansKey, [...(loans ?? []), book], false);
```
kristojorg, line 62: This won't revalidate, is that intentional?

kristojorg (commented):
Ok basically I think:

1. Swapping a dependency on the loans endpoint for a dependency on every individual book endpoint probably isn't really worth it. Removing the patron-specific information from the book list or some other UX solution seems better.
2. You can still call `useUser` to get access to the `mutate` function without waiting for the loans to arrive. I think this is better than `unstable_serialize`

apps/oeh/src/components/FulfillmentCard.tsx
```diff
@@ -66,20 +59,27 @@ const FulfillmentCard: React.FC<FulfillmentCardProps> = ({ book }) => {
       ]
     });
 
-  const bookData =
-    loanedBooks?.find((loanedBook: OpenEBook) => book?.id === loanedBook.id) ??
-    book;
+  //TODO: Ask Jen for the loading state design
+  if (!book) {
+    return null;
+  }
```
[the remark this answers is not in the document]
Toxiapo: This is correct, books from the collection page would always show the 'borrowable' status as default. To your point that if we had to request every book, it would not be an improvement over fetching the loans endpoint unless it's already been cached by swr.

apps/oeh/src/components/FulfillmentCard.tsx
```diff
@@ -18,26 +17,20 @@ import {
...
-const FulfillmentCard: React.FC<FulfillmentCardProps> = ({ book }) => {
-  const { borrow, returnBook, isBorrowingOrReturning } = useBook(book.url);
+const FulfillmentCard: React.FC<FulfillmentCardProps> = ({ bookSeed }) => {
+  const {
+    data: book,
+    error,
+    borrow,
+    returnBook,
+    isBorrowingOrReturning
+  } = useBook(bookSeed.url);
```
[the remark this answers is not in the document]
Toxiapo: I don't think so. The key we used for swr was the URL + token, so as long as it's been fetched and under the same user, it should use the cached version

apps/oeh/src/hooks/useBook.ts
Toxiapo: Never! It was a debugging code that I forgot to delete.

apps/oew/src/hooks/useBorrowOrReturn.ts
Toxiapo: You are right, this should revalidate

Toxiapo:
> Ok basically I think:
> 
> 1. Swapping a dependency on the loans endpoint for a dependency on every individual book endpoint probably isn't really worth it. Removing the patron-specific information from the book list or some other UX solution seems better.
> 2. You can still call `useUser` to get access to the `mutate` function without waiting for the loans to arrive. I think this is better than `unstable_serialize`

You are right. The loans would just revalidate without messing with the UI. I added back useUser!

Toxiapo:
Following with this ticket: https://jira.nypl.org/browse/OE-699

The removal of the fulfillment card will be discussed further, I think this PR is good to go. @kristojorg

apps/oeh/src/components/FulfillmentCard.tsx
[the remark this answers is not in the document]
kristojorg: That's a good point. We add all the books to the cache when loading the list

kristojorg (approved):
Looks good to me! Hopefully we can remove the fulfillment card from the list page but until then this seems good

samanthaandrews:
Can we merge this one?

PR #604, opened 2023-02-09
Release OE Web 4.8.0

samanthaandrews:
## ✅ Open eBooks Web Release Checklist ✅

- [x] Do these changes modify environment variables? If so, have you updated env vars in the Dockerfile, the GH Action Workflows & submitted the change to AWS through DevOps?
- [x] Has someone run through the [QA Testing Checklist](https://docs.google.com/document/d/1Sm_mFNAlbUrjT4bfJt5boQMeIA7LoPwoPa4uXqp8kFg/edit?usp=sharing) on https://staging.openebooks.org/ and confirmed everything is working as expected?

⚠️ Make sure you select the "Create a merge commit" option when you merge PRs to the `production` branch.


.github/workflows/oew-release-prod.yml
```diff
@@ -1,12 +1,10 @@
 name: OEW Production Release
 
-# workflow will only run when a tag with the oe-web-* prefix is pushed to a release/oe-web-* branch
+# workflow will only run when a tag with the oe-web-* prefix is pushed
 on:
   push:
     tags:
       - "oe-web-*"
-    branches:
```
samanthaandrews, line 8: this actually triggered the prod release action to run 😅 i noticed that the prod release was running when it shouldn't be, and canceled it so prod wasn't actually updated. but it was a close call.

i thought that by including the branch name, it meant that this workflow would run when a tag was pushed to a release branch, but realized that it would actually run if we pushed a tag OR if we pushed to that release branch. so by removing `branches` from the trigger, we ensure this only gets run when an `oe-web-*` tag is pushed.

Toxiapo: Nice catch!!

Toxiapo (approved):
🎉

ayan1229 (approved):

PR #606, opened 2023-02-13
Oe 848/document release process

kristojorg:
This PR just:

- adds documentation surrounding setting up Appflow for OEH
- Renames the trapeze config file to make it easier to distinguish from the appflow config file


samanthaandrews (approved):
LGTM and glad we linked that certificates repo 👍

PR #607, opened 2023-02-13
OE-842: fix contact form bug

oliviawongnyc:
[OE-842](https://jira.nypl.org/browse/OE-842): Fixes styling in OEW contact form so input values are visible and labels don't float on top of the header


Toxiapo (commented):
The input fields are good now! It's just that the textarea's value is still hidden if you lose focus.

samanthaandrews:
> The input fields are good now! It's just that the textarea's value is still hidden if you lose focus.

@Toxiapo I'm not seeing this on Chrome! Which browser? Can you post a video?

Toxiapo:
@samanthaandrews Yea, Chrome@109.0.5414.119 .

I just checked Safari and that seems fine. :/

https://user-images.githubusercontent.com/14434233/218776946-9a452ae0-da2d-473a-81f8-2fab9c385850.mov

kristojorg:
Just a note for our future selves to push back on these types of inputs because while they're fancy, they inevitably incur a maintenance cost (and it's just a contact page after all) : )

Toxiapo:
@samanthaandrews @oliviawongnyc ok false alarm! It's because my Grammarly extension was automiaclly adding an element to the textarea field, making the sibling selector to not work.

@kristojorg yea agree, and this is an incident that proven this could be more maintenance costly than we would like lol

samanthaandrews:
> Just a note for our future selves to push back on these types of inputs because while they're fancy, they inevitably incur a maintenance cost (and it's just a contact page after all) : )

100% agree. The custom nature of the CSS and markup has also made it harder to ensure it is accessible.

samanthaandrews (approved):

Toxiapo (approved):

PR #609, opened 2023-02-15
PJR-147: create `openBook` API

kristojorg:
We previously had an `openBook` api but that was removed for the purposes of the bookmarks POC. I am here adding it back as we are now opening an actual book. I am currently passing a stub path to the book, but this will be updated in the future to be a real path. 

Mobile devs - please let me know if this works for you


.vscode/user-workspaces/kristo-ereading.code-workspace
```diff
@@ -0,0 +1,17 @@
+{
```
ettore, line 1: this looks like it should be gitignored

apps/oeh/src/plugins/Reader/definitions.ts
```diff
@@ -42,7 +42,13 @@ export interface ReaderPlugin {
    *
    * - path: local path to the downloaded book
    */
-  openBookmarksList({ bookmarks }: { bookmarks: Bookmarks }): Promise<void>;
+  openBook({
+    bookmarks,
```
ettore, line 46: interesting, do you think this is how we will pass bookmarks to the reader long term? it seems fine to me, just wondering. 

Also, is this `bookmarks` array currently empty when it's sent or does it include the sample bookmarks we currently had for the temporary bookmarks screen? It would be good to keep including those fake bookmarks for now, to exercise the logic that parses the contents of the `call` object since that was working, even if with dummy data.

Not sure what the other native devs think @roywatson @MalcolmMcFly @ErnestFan

ettore (commented):
looks fine to me, but I want to make sure we are not losing the working example of web-native communication. (I don't care much about the current bookmarks UI.)

.vscode/user-workspaces/kristo-ereading.code-workspace
kristojorg: That is true, good catch. I had it gitignored in another branch but forgot when switching over

apps/oeh/src/plugins/Reader/definitions.ts
kristojorg: I think it makes sense to pass the current set of bookmarks when we make the `openBook` call. We will continue to call `updateBookmarks` when they are updated. Open to other ideas though, that was just my default thought.

kristojorg: I also just added a couple default bookmarks for testing purposes to our `openBook` call

ettore: agree -- this sounds good to me.

ettore:
I still see the `kristo-ereading.code-workspace` file but other than that this looks good to me.

kristojorg:
> I still see the kristo-ereading.code-workspace file but other than that this looks good to me.

Fixed!

@MalcolmMcFly @roywatson can either of you confirm that this looks good and send over an approval? Thank you!

kristojorg:
I updated the API to match what we decided on in [this document](https://docs.google.com/document/d/1Jjpw1m7uyVs3MWkwolOQ3EJgvroRcdeRfAb8Ipy_7u0/edit#). I have not added the `lastReadingPosition` or `UPDATE_READING_POSITION` event nor the `encryptionScheme` parameter. I think we have a little more figuring out to do with those APIs when we get to that point.

apps/oeh/src/plugins/Reader/definitions.ts
```diff
@@ -31,18 +44,21 @@ export type Bookmarks = Array<Bookmark>;
...
+   * - pubPath: local path to the downloaded book
+   * - bookmarks: initial list of bookmarks
    */
-  openBookmarksList({ bookmarks }: { bookmarks: Bookmarks }): Promise<void>;
+  openBook({
+    bookmarks,
+    pubPath
+  }: {
+    bookmarks: Bookmarks;
+    pubPath: string;
```
ettore, line 60: is the `pubPath` relative to the exploded ePub root?

ettore (approved):

apps/oeh/src/plugins/Reader/definitions.ts
kristojorg: So this is something we have to discuss: what is the difference in this api call between the axisnow-specific exploded epub and a normal packaged epub? I guess I was thinking I was implementing `openBook` for a standard packaged epub (since I didn't think we were ready to handle the axisnow case yet). In that case this would be the path to the `.epub` file itself. 

In the axisnow case, what do you think makes sense? Maybe it makes sense to create an entirely separate `openAxisnowBook` API since that is kind of a unique scenario. 

In any case, for the time being, let's consider this the API for a packaged epub and the `pubPath` would be the local path to the `.epub` file

roywatson (approved):
Looks good for Android

apps/oeh/src/plugins/Reader/definitions.ts
ettore: sorry my bad, I misunderstood the meaning of `pubPath` (despite the documentation comment 😅). Can you rename it to `publicationFileURL`?  For what the native side is concerned, there's no difference between the packaged or exploded epub, as long as the packaged file name ends with a `.epub` extension and the exploded path points to the root dir of the epub content. @roywatson @MalcolmMcFly @ErnestFan let me know if I'm in error here.

kristojorg: Okay renaming to `publicationFileURL`. My idea on the exploded one was to point to the `container.xml` file as that is sort of the entrypoint (it contains links to all the other files needed). But if the root dir is better then I can do that instead, just let me know

PR #610, opened 2023-02-16
PJR-150: Setup Koin for DI in native code

MalcolmMcFly:
**What does this do?**
- [x] Add Koin library for dependency injection
- [x] Made `ReaderViewModel` injectable
- [x] Removed all hard-coded strings
- [x] Removed all hard-coded colors
- [x] Removed all wildcard imports
- [x] Added new utility retrieving TAG for log statements

**Why are we doing this?**
Adding dependency injection makes the code more testable and components easier to reuse.
[PJR-150: Setup Koin for DI in native code](https://jira.nypl.org/browse/PJR-150)

**How should this be tested?**
Nothing changes from a UX standpoint, the Android project should build and function the same as before.

**Other**
- These changes DO NOT modify any environment variables.


roywatson (approved):
Looks lie what we went through together

PR #613, opened 2023-02-22
Create an auth package for Clever and Firstbook

Toxiapo:
This PR moves the Clever and FirstBook logic into its package.


apps/oew/src/pages/app/signin/method/firstbook.tsx
```diff
@@ -100,7 +100,10 @@ export const FirstbookAuthHandler: React.FC = () => {
     // try to signin with these credentials
     const basicToken = generateToken(login, password);
     try {
-      const { token, expiresAt } = await fetchFirstbookJWT(basicToken);
+      const { token, expiresAt } = await fetchFirstbookJWT(
+        basicToken,
+        FIRSTBOOK_OAUTH_JWT_ENDPOINT
```
Toxiapo, line 105: I added this param to this function instead of putting it in the constant file because we don't have the setup for QA vs PROD variable yet.

apps/oew/src/pages/app/signin/method/firstbook.tsx
```diff
@@ -4,18 +4,12 @@ import { useForm } from "react-hook-form";
...
 import { FirstbookAuthType } from "opds/opds1";
 import SigninWrapper from "components/chakra-ui/SigninWrapper";
 import { SxProp } from "theme-ui";
-import {
-  fetchFirstbookJWT,
-  performAuthRedirect,
-  setAuthCookieBrowser
-} from "utils/auth";
-import { FirstbookAuthCredentials } from "interfaces";
+import { performAuthRedirect, setAuthCookieBrowser } from "utils/auth";
```
oliviawongnyc, line 12: Why can't this stuff go in the auth package?

samanthaandrews (approved):

oliviawongnyc (commented):
Looks so good!! My only question is why we couldn't move more of utils/auth.ts. Is it because of dependencies the code is using?

apps/oew/src/pages/app/signin/method/firstbook.tsx
Toxiapo: While working on the clever auth, I realized this URL could be extracted from the auth doc. Are there reasons we wanted to hardcode the auth URL for firsbook and not use the version from the backend?

apps/oew/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+
 - Chore: Add comments to AxisNow decryptor and remove unused files
 - Add: OE-852: Automate Contact Form Testing
+- Add: OE-859: Install `auth` and move FirstBook and Clever logics to `auth` package
```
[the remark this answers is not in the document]
Toxiapo: I like this!

apps/oew/src/pages/app/signin/method/firstbook.tsx
Toxiapo: The redirect function was specific for nextjs, and the cookie function uses the browser window.cookie. My intention was to move these auth methods and have them only do fetching and parsing tokens.

oliviawongnyc: I see. Okay that makes sense!

PR #616, opened 2023-02-24
OE-795: Add Clever auth

Toxiapo:
Add Clever auth with capacitor/browser and capcaitor/app plugin


apps/oeh/ios/App/App/Info.plist
```diff
@@ -56,5 +56,17 @@
 	<true/>
 	<key>UIWhitePointAdaptivityStyle</key>
 	<string>UIWhitePointAdaptivityStyleReading</string>
+
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>OpenEbooks</string>
```
Toxiapo, line 65: We need a custom URI for native apps

apps/oeh/ios/App/App/Info.plist
```diff
@@ -2,59 +2,71 @@
...
+    <key>CFBundleURLTypes</key>
+    <array>
+        <dict>
+            <key>CFBundleURLSchemes</key>
+            <array>
+                <string>OpenEbooks</string>
+            </array>
+            <key>CFBundleURLName</key>
+            <string>org.nypl.labs.OpenEbooks</string>
+        </dict>
```
Toxiapo, line 69: @ettore For the redirect to work, I had to add a custom URI so that the app knows where to redirect. This is working fine for ios >= 13, relogging the second time is also fine. I think there was a concern about security risks with using a custom URL scheme but I am not sure what that really means and if that's a concern when you were implementing the clever flow? afaik, companies like airbnb, bestbuy uses custom URL as well. cc. @samanthaandrews  @kristojorg

apps/oeh/android/app/src/main/res/values/strings.xml
```diff
@@ -3,6 +3,6 @@
     <string name="app_name">oeh</string>
     <string name="title_activity_main">oeh</string>
     <string name="package_name">org.openebooks.app</string>
-    <string name="custom_url_scheme">org.openebooks.app</string>
```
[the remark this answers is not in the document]
Toxiapo: @roywatson I am curious what your thought on custom URLs for android and if you had experience with it.

apps/oeh/ios/App/App/Info.plist
ettore: ideally Universal Links should be used instead of custom URLs.  They are newer, more secure and per Apple it's the preferred way of deep linking:
https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app

The security risk for custom url schemes is that different apps could register the same url schemes. Since deep linking with custom URL schemes is all done client-side, it could lead to wrong redirection and other hacks.  See this note in the doc above:

> "If multiple apps register the same scheme, the app the system targets is undefined. There’s no mechanism to change the app or to change the order apps appear in a Share sheet."

Universal Links have a server side component in the flow that prevents that. That's what we did when we re-did the auth flow for OE native last year.  Setting up universal links was a little painful but the app is set up now, and if we keep the same AppID it should continue to work. 

Here are some more articles about the differences between the two:
https://stackoverflow.com/a/45533475/238983

https://medium.com/wolox/ios-deep-linking-url-scheme-vs-universal-links-50abd3802f97

More explanation on how UL work (I think this was my reference when I did the work):
https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html

Toxiapo: Thank you for the links and info! @ettore It seems like Ionic also suffers from the[ redirect problem ](https://github.com/ionic-team/capacitor-plugins/issues/675#issuecomment-957760081)with the universal links you mentioned on the ticket. The in-app browser uses `SFSafariViewController` by default.

I would like to set this up the right way if possible, but I am having difficulties getting the universal link to run. Do you think I can put up some time on your calendar to go over this?

ettore: definitely... I'll need to review how this works in the details because it's been a while 😅 

if I remember well SFSafariViewController is also what I used, and one thing I didn't try was using a WKWebView, which provides more control. Is there a way to control how the web page is opened in the in-app browser in capacitor?

Toxiapo: Thank you!

The package I am using here is @capcitor/browser, which uses SFSafariViewController I don't think I can change that. There's an [alternative plugin](https://github.com/apache/cordova-plugin-inappbrowser) that uses WKWebView but it's not being maintained, unfortunately...

Toxiapo:
Converting this back to Draft while I meet with Ettore on the Universal Link

apps/oeh/ios/App/App/Info.plist
kristojorg: Just dropping a note - if this requires a server side component and we don't want to change the CM, we could use the API routes at openebooks.org pretty easily.

ettore: btw all I meant with "server side component" was that the `/.well-known/apple-app-site-association` file that needs to be present. For the current production apps we have:

https://librarysimplified.org/.well-known/apple-app-site-association

So I think we should be able to reuse this one. We can also move it to a new domain and simply refer to that.

ettore: I reread the [UL docs](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html) and I forgot that to implement this there's a native part as described there. So we need to explore what we can do with Capacitor.

Toxiapo: Seems like the server URL should match the `appID` we set in the `[capacitor.config.ts](https://github.com/NYPL/ereading-clients/blob/39e991976e2db536d2e0e2aac7ef0a7f1e914dd6/apps/oeh/capacitor.config.ts#L7)` file. If we want to continue using `librarysimplifed`, we might need to update our `appId` to something like `org.librarysimplified.OpenEbook`. Is this ok for us to do?  @kristojorg @kristojorg

ettore: so, the appID (which is really the bundleID, Capacitor is confusing the 2 things here) should not change because that's used by the apple ecosystem to identify the app in the App Store. However, we can change the rest: both the domain that hosts the `apple-app-site-association` resource, its contents, as well as how we refer to it (the "associated domain") from Xcode as described in the [apple docs](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html). 
Is there documentation about universal links and capacitor maybe? If we can't add native code we'll need to pass thru capacitor somehow.

Toxiapo: Yes, the [capacitor doc ](https://capacitorjs.com/docs/guides/deep-links#ios-configuration) never mentioned the appID to be associated with the hosting domain, so I am guessing we could just define the associated domain ourselves.

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -457,6 +458,9 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED = YES;
+				CODE_SIGN_ENTITLEMENTS = "Open eBooks.entitlements";
```
ettore, line 461: These are a little odd because they appear to be on the `Pods-Open eBooks` configuration but I'm not sure how.... did you do these changes explicitly or was it Xcode?

Toxiapo: I don't think I have touched that file so it could be Xcode

Toxiapo: would this affect our build? I can remove this line and try if can build the app locally.

ettore: I checked out your branch and I think _some_ of these changes are not necessary. This one for CODE_SIGN_ENTITLEMENTS is needed because this is how we get applinks to work.

I'll comment on the others explicitly to avoid confusion.

I also figured out where in Xcode we can see these changes -- that's the part I could not understand from the PR (because of Xcode cryptic project file format).

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -457,6 +458,9 @@
...
+				CODE_SIGN_IDENTITY = "Apple Development";
```
ettore, line 462: to see these changes in Xcode you should click on the "App" project on the far left where the source files are. Then click on the Open eBooks target and "Build Settings". Scroll down to Signing and you should see something like this:

xcodebuildsettings

This "Levels" view shows you how Xcode applies build settings, which is kinda like CSS, that is in increasingly specific layers; so from right to left we have:
- defaults layer (unchangeable, Xcode factory defaults)
- project layer (settings affecting the whole project, overriding the defaults)
- config files layer (.xcconfig files such as our `Pods-Open eBooks.debug.xcconfig` file are editable and are the foundation of targets; they override the project's)
- target layer (settings affecting an app's build directly, overriding the config files)

So for example any settings at target layer overrides anything else. 

As you can see the customization for "Provisioning Profile" (which corresponds to PROVISIONING_PROFILE_SPECIFIER in the project source)  is unneeded because it's the same as the Project. If you select it and press the Delete key it will remove it and the app will build the same.

Same reasoning for the Code Signing style (CODE_SIGN_STYLE): unneeded duplicated customization.

For CODE_SIGNING_IDENTITY, can you see what happens if you keep the project default "iOS Developer"?  You can also do the same trick here, since the default is in fact "iOS Developer". I changed and it still built. It would be good to keep the same code signing identity.

Another way to do all this is to manually edit the project file in a text editor, but I thought it would be good to show how these settings translate on the UI. Let me know if you want to pair on this, I'm available.

Toxiapo: Thanks! This is helpful.

apps/oeh/README.md
```diff
@@ -250,6 +250,35 @@ You can run this app with live-reloading on a real device as well as a simulator
...
+    buildTypes {
+        release {
+            ...
+            signingConfig = signingConfigs.getByName("debug")
+            ...
+        }
+    }
+}
+```
+
```
Toxiapo, line 281: @MalcolmMcFly I added this instruction for local development. I believe this is all we need?

apps/oeh/src/pages/sign-in/clever/callback.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default function Callback() {
+  const { credentials, setCredentials } = useCredentials();
+  const history = useHistory();
+
+  useEffect(() => {
+    App.addListener("appUrlOpen", (event: URLOpenListenerEvent) => {
+      console.info("Getting the url from the appUrlOpen listenerer");
+      validateAndRedirect(new URL(event.url)?.hash, setCredentials, history);
+    });
+    validateAndRedirect(window?.location?.hash, setCredentials, history);
```
samanthaandrews, line 27: why do we run `validateAndRedirect` twice?

samanthaandrews (changes requested):
Just a few small things regarding the xcode changes. this was a ton of work and it looks really good.

apps/oeh/src/pages/sign-in/clever/callback.tsx
Toxiapo: Good catch, we don't need two.

apps/oeh/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -478,6 +483,9 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED = YES;
+				CODE_SIGN_ENTITLEMENTS = "Open eBooks.entitlements";
+				CODE_SIGN_IDENTITY = "iPhone Developer";
+				CODE_SIGN_STYLE = Automatic;
```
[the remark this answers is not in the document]
Toxiapo: I removed these, but then it came back after building the app...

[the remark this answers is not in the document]
samanthaandrews: Ok, that's fine. I am currently talking with Appflow support so they can help me understand what these files are supposed to include, so I'm not going to worry about it until they tell me what these values should be!

apps/oeh/src/lib/auth/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const AUTH_REDIRECT_LANDING_PAGE = "/app/discover";
+
+/**
+ * Helper function to:
+ * 1. Extract the access token from the callback URL
+ * 2. Store the credentials
+ * 3. Redirect to a landing page
+ */
+export const validateAndRedirect = (
```
samanthaandrews, line 14: nice!

samanthaandrews (approved):

apps/oeh/README.md
MalcolmMcFly: Yes this is good.

apps/oeh/android/app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
kristojorg: I honestly don't remember adding this but it's possible I did?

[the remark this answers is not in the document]
Toxiapo: OK! I think this is fine because we don't use these values.

apps/oeh/src/pages/sign-in/SignIn.tsx
```diff
@@ -7,13 +7,42 @@ import {
...
+  React.useEffect(() => {
+    /**
+     * Listens to the Clever redirect event
+     */
+    App.addListener("appUrlOpen", (event: URLOpenListenerEvent) => {
+      const url = event.url;
+      // Andorid automically adds the www subdomain to the URL.
+      const cleanUrl = url.replace(/^https:\/\/www\./i, "https://");
+
+      console.info("Getting the url from the appUrlOpen listenerer.", cleanUrl);
```
kristojorg, line 34: Stray console

kristojorg (approved):
This looks great! And seems to work great too : )

It's a complicated feature and I have a bit of trouble tracing the flow through the various files myself. It might be worthwhile to write a little description of the process in the README. Like a high level, "these are the steps" kind of thing.

Toxiapo:
> This looks great! And seems to work great too : )
> 
> It's a complicated feature and I have a bit of trouble tracing the flow through the various files myself. It might be worthwhile to write a little description of the process in the README. Like a high level, "these are the steps" kind of thing.

Good idea!

PR #619, opened 2023-02-28
[WIP] OE-787/first book oauth

oliviawongnyc:


apps/oeh/src/components/CredentialsContext.tsx
```diff
@@ -1,17 +1,22 @@
...
+import { tokenExpired, verifyToken } from "auth";
+import { FIRSTBOOK_OAUTH_JWT_ENDPOINT } from "constants/openebooks";
 
 const CREDENTIALS_STORAGE_KEY = "user-credentials";
 
 export type CredentialsContext =
   | {
-      credentials: CredentialsState | undefined;
-      setCredentials: (creds: CredentialsState) => void;
+      getVerifiedCredentials: () => Promise<
```
kristojorg, line 15: I would add a comment here to specify why we have `credentials` and `getVerifiedCredentials`. That the latter is to be called before actually using the credentials for making a fetch request

apps/oeh/src/components/CredentialsContext.tsx
```diff
@@ -39,6 +46,31 @@ export function CredentialsProvider({
...
+    if (credentials) {
+      try {
+        const parsedCredentials = JSON.parse(credentials);
+        const verifiedCredentials = await verifyToken(
+          parsedCredentials,
+          FIRSTBOOK_OAUTH_JWT_ENDPOINT
+        );
+
+        if (tokenExpired(verifiedCredentials)) {
+          setCredentialsJson(verifiedCredentials as FirstbookAuthCredentials);
```
kristojorg, line 62: Why do we need to cast this type here? It would be best if `verifyToken` returns the correct type

kristojorg: It looks like it is because `verifyToken` returns `OpenEbooksAuthCredentials`, not specifically the `FirstbookAuthCredentials`. However, what I notice is that `verifyToken` will always return the specific type of credentials that you pass in `FirstbookAuthCredentials` in, `FirstbookAuthCredentials` out. There is actually a way to indicate this in typescript, called [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). This is how I would write the signature of that function with an overload:
```ts
export async function verifyToken(
  credentials: FirstbookAuthCredentials,
  oauthEndpoint: string
): Promise<FirstbookAuthCredentials>
export async function verifyToken(
  credentials: CleverAuthCredentials,
  oauthEndpoint: string
): Promise<CleverAuthCredentials>
export async function verifyToken(
  credentials: OpenEBooksAuthCredentials,
  oauthEndpoint: string
): Promise<OpenEBooksAuthCredentials> {
```

This tells Typescript that if you get out what you pass in. I'm not exactly sure if it's what you want here, but in any case I'd love to remove this casting : )

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -22,19 +25,28 @@ const FirstbookForm: React.FC = () => {
...
+        methodType: FirstbookAuthType,
+        token,
+        basicToken,
+        expiresAt
+      };
+
+      setCredentials(credentials);
+    } catch (err: any) {
+      setCredentials(undefined);
+      console.error("Authentication failed. Please try again.");
```
kristojorg, line 47: Personally, I think we should start using an [IonAlert](https://ionicframework.com/docs/api/alert) for these situations where we don't have specific UI for the error yet. It's prettymuch as easy as console.error and it at least presents something to the user for now

apps/oeh/src/hooks/useBook.ts
```diff
@@ -15,9 +16,11 @@ export function useBook(url: string) {
     Indicates the start and the end of the return / borrow action
   */
   const [isBorrowingOrReturning, setIsBorrowingOrReturning] = useState(false);
-  const { credentials } = useCredentials();
+  const { credentials, getVerifiedCredentials } = useCredentials();
   const swr = useSWR<OpenEBook, Error>(
-    credentials?.token ? [url, credentials?.token] : null,
+    credentials?.token
+      ? [url, getVerifiedCredentials, credentials?.token]
+      : null,
```
kristojorg, line 23: I am realizing that if we don't want to repeat this pattern everywhere, we could make a SWR [middleware](https://swr.vercel.app/docs/middleware#request-logger). It can be kind of like the logger example in that it extends the fetcher with the verifyToken functionality.

kristojorg: Just an idea, not totally necessary

apps/oeh/src/lib/download/axisnow.ts
```diff
@@ -68,7 +70,7 @@ export async function downloadAxisnow(
    */
   await removeAxisnowDownload(book.id);
 
-  const fulfillmentInfo = await fetchFullfillment(
+  const fulfillmentInfo = await fetchFulfillment(getVerifiedCredentials)(
```
kristojorg, line 73: I think we don't need to pass token into this function anymore since we are going to use the verified credentials

apps/oeh/src/lib/download/useDownload.tsx
```diff
@@ -52,7 +52,12 @@ export default function useDownload(book: OpenEBook | undefined) {
       console.error("Cannot download without a book");
       return;
     }
-    await downloadAxisnow(book, credentials.token, setDownloadProgress);
+    await downloadAxisnow(
```
kristojorg, line 55: I think it would also be valid to just call `const token = getVerifiedCredentials()` at the start of the `download` function and then pass it directly in to `downloadAxisnow`. Either way but I don't think we need to pass the token and the verify function

apps/oeh/src/lib/fetching/fetch.ts
```diff
@@ -58,16 +88,27 @@ export const entry2Book = (catalogUrl: string) => async (entry: OPDSEntry) =>
 
 export const feed2Collection =
   (info: RequestInfo) => async (feed: OPDSFeed) => {
+    console.log("feed2collection");
```
kristojorg, line 91: Stray log

apps/oeh/src/lib/fetching/fetch.ts
```diff
@@ -76,9 +117,16 @@ export const fetchFullfillment = (info: RequestInfo, opt?: RequestInit) =>
...
-  const collection = await fetchCollection(url, {
+  [
+    url: string,
+    getVerifiedCredentials: () => Promise<
+      OpenEBooksAuthCredentials | undefined
+    >,
+    token: string
+  ]
+> = async ([url, getVerifiedCredentials, token]) => {
+  console.log("fetching loans");
```
kristojorg, line 128: stray log

apps/oeh/src/lib/fetching/fetch.ts
```diff
@@ -76,9 +117,16 @@ export const fetchFullfillment = (info: RequestInfo, opt?: RequestInit) =>
...
+  const collection = await fetchCollection(getVerifiedCredentials)(url, {
```
kristojorg, line 129: I think we need to review the token on line 130 since that one is not verified and it should now be added by `fetchCollection`

apps/oeh/src/pages/MyBooks.tsx
```diff
@@ -11,20 +11,23 @@ import {
...
+const MyBooks: React.FC<RouteComponentProps> = () => {
+  const { credentials, getVerifiedCredentials } = useCredentials();
+
   const {
     data: books,
     mutate,
     isValidating
   } = useSWR(
-    credentials?.token ? [LOANS_ENDPOINT, credentials.token] : undefined,
+    credentials?.token
```
kristojorg, line 28: I don't know if we can pass a function in as an argument in SWR. It uses the identity of these arguments to know when to refetch the value. I think we should be passing this directly to the fetcher: `fetchLoans(getVerifiedCredentials)`

kristojorg (commented):
This is looking great, just adding a few comments/suggestions!

I think it's unfortunate that we have to pass around this `getVerifiedCredentials` function everywhere. It would be nice to find a cleaner way, and I do think one exists, but maybe that's best left for a future cooldown ticket

apps/oeh/src/lib/download/axisnow.ts
oliviawongnyc: I thought that, but was scared to remove 😅.  I think this is true in multiple places. Will update.

apps/oeh/src/components/CredentialsContext.tsx
oliviawongnyc: Do we actually need `credentials`? `getVerifiedCredentials` returns credentials, and we can also find credentials in local storage. Is storing it in state necessary?

apps/oeh/src/components/CredentialsContext.tsx
oliviawongnyc: Woah! Cool!

apps/oeh/src/components/FirstbookForm.tsx
oliviawongnyc: I like that.

apps/oeh/src/hooks/useBook.ts
oliviawongnyc: For this line, specifically, do you mean?

`credentials?.token ? [url, getVerifiedCredentials] : null` 

...or do you mean putting the `getVerifiedCredentials` logic in the middleware rather than before each fetch request? I think the latter makes sense, but maybe worth saving for the cool down sprint?

apps/oeh/src/lib/fetching/fetch.ts
oliviawongnyc: Yes! Thank you.

apps/oeh/src/pages/MyBooks.tsx
oliviawongnyc: I think we _can_ pass a function (from [the docs](https://swr.vercel.app/docs/arguments): "In some scenarios, it's useful to pass multiple arguments (can be any value or object) to the fetcher function."). But are you worried that it is causing fetchLoans to run at the wrong times?

apps/oeh/src/hooks/useBook.ts
Toxiapo: This sounds like a good idea, I would love to learn more about the middleware pattern in swr. cooldown sprint make sense!

apps/oeh/src/lib/fetching/fetch.ts
```diff
@@ -8,36 +8,65 @@ import {
...
-    .then(handleCMError(info))
-    .then(text)
-    .then(parser.parse)
-    .catch(handleParsingError(info))
-    .then(validate(OPDSEntry, info))
-    .then(entry2Book(catalogUrl));
+// Before each fetchBook, fetchCollection, and fetchFulfillment request,
+// we verify the JWT token using getVerifiedCredentials. If the token is
+// nearing its expiration, we fetch new credentials to use in the upcoming
+// fetch call. If not, we return we return the existing credentials and proceed.
```
Toxiapo, line 22: nit: Let's make this a block comment `/** */`

apps/oeh/src/components/CredentialsContext.tsx
```diff
@@ -1,34 +1,49 @@
...
-  React.createContext<CredentialsContext>(undefined);
+export const CredentialsContext = createContext<CredentialsContext>(undefined);
 
 export function CredentialsProvider({
   children
 }: {
   children: React.ReactNode;
 }) {
-  const [credentials, setCredentials] = useState<CredentialsState>();
+  const [credentials, setCredentials] = useState<FirstbookAuthCredentials>();
```
Toxiapo, line 41: same here.

apps/oeh/src/components/CredentialsContext.tsx
```diff
@@ -1,34 +1,49 @@
...
 
 export type CredentialsContext =
   | {
-      credentials: CredentialsState | undefined;
-      setCredentials: (creds: CredentialsState) => void;
+      // Existing credentials show that the user was logged in
+      // at some point. getVerifiedCredentials ensures credentials
+      // are valid, updating them if they are close to expiring. It
+      // is run prior to every fetch request.
+      credentials: FirstbookAuthCredentials | undefined;
```
Toxiapo, line 24: Update type, and let's make this a block comment as well

apps/oeh/src/components/CredentialsContext.tsx
```diff
@@ -1,34 +1,49 @@
...
 }: {
   children: React.ReactNode;
 }) {
-  const [credentials, setCredentials] = useState<CredentialsState>();
+  const [credentials, setCredentials] = useState<FirstbookAuthCredentials>();
   const [isLoaded, setIsLoaded] = useState(false);
 
-  async function setCredentialsJson(creds: CredentialsState | undefined) {
+  async function setCredentialsJson(
+    creds: FirstbookAuthCredentials | undefined
```
Toxiapo, line 45: type

Toxiapo (changes requested):
Just the `Type` change, otherwise, this looks good to me!

Toxiapo:
@oliviawongnyc Could you update the playwright tests are well? It may be because the first book logins are now different.

apps/oeh/src/pages/MyBooks.tsx
kristojorg: Yea that would be my worry. Passing it as a key here makes the fetch run whenever it changes, whereas that isn't really our intent (the `getVerifiedCredentials` function shouldn't ever change I think). So while it may work, I think it's better to pass it directly to the `fetchLoans` function

oliviawongnyc: Okay, I removed getVerifiedCredentials from all cache keys. Let me know if this is what you had in mind. There are four instances I changed, in `BookListView`, `useBook`, `useCollection`, and `MyBooks`.

oliviawongnyc:
@Toxiapo, can you clarify which `Type` change you were talking about here?

Toxiapo (approved):
LGTM!

PR #620, opened 2023-03-02
PJR-195/playwright

kristojorg:
Adds a basic playwright test and sets it up in CI for PJR


samanthaandrews (approved):
once we have real endpoints we can add `playwright-msw` to mock them. looks good to me 🚀

PR #623, opened 2023-03-03
IOS-469 Convert makeReaderVC call to Combine

ettore:
This changes the way we open the reader to use Combine. The main challenges here were:
- find a way to map errors that could happen in different apis to a single Combine chain, so that we can process them in one place. Currently errors are handled in 3 different guards/catch blocks.
- find a way to handle nil/empty array situations in a Combine-friendly way
- avoid complicating the code 
- handle the call to the `completion` handler in a single place if possible


apps/oeh/ios/App/SwiftPackages/NYPLReader/Sources/ReaderModule.swift
```diff
@@ -130,38 +130,34 @@ extension ReaderModule {
...
-            viewController.hidesBottomBarWhenPushed = true
-            navVC.pushViewController(viewController, animated: true)
+            vc.navigationItem.backBarButtonItem = backItem
+            vc.hidesBottomBarWhenPushed = true
+            navVC.pushViewController(vc, animated: true)
         }
 
         guard let format = formats.first(where:{ $0.supports(publication) }) else {
-            delegate.presentError(ReaderError.formatNotSupported, from: navVC)
+            delegate?.presentError(ReaderError.formatNotSupported, from: navVC)
```
ettore, line 142: ideally I'd like to handle this error in the `sink` down below, so that all errors are in one place

apps/oeh/ios/App/SwiftPackages/NYPLReader/Sources/ReaderModule.swift
```diff
@@ -127,41 +127,36 @@ extension ReaderModule {
...
-        } catch {
-            delegate.presentError(error, from: navVC)
+            vc.navigationItem.backBarButtonItem = backItem
+            vc.hidesBottomBarWhenPushed = true
+            navVC.pushViewController(vc, animated: true)
         }
 
-        completion()
+        _ = Just(formats).setFailureType(to: ReaderError.self)
+            .flatMap { formats in
```
ettore, line 142: this is necessary to wrap the whole logic inside a unique publisher. We are taking the `formats` array and wrapping into a `Just` publisher, so that the array is being published as a single object. At that point `flatMap` receives one element (our array) and now we can filter and make the VC, and return errors of as many different types as we want.  Note that we need to call `setFailureType` so that the error type of the Just publisher (which I believe is `Never`) can match the error type of the publisher returned by `makeReaderViewController` (ReaderError).

This works fine but I wonder if this can be optimized further, since wrapping with `Just` and flatMapping a single element is kind of weird.

In any case, this way we can process all errors and we can call the `completion` closure in one place i.e. in the `sink` instead of having them in multiple `guard` blocks.

apps/oeh/ios/App/SwiftPackages/NYPLReader/Sources/ReaderModule.swift
ettore: thank you Ernest, this was very useful to make me understand that if there are different error situations that prevent the enactment of the chain, we need to wrap those in one place, in one "chain link".  For example `makeReaderViewController` is a very specific API that only `ReaderFormat` publishers have, so we can transform/filter/process all we want, but we will need to handle the type difference in one place. In that same place we need to handle the errors related to the type.

I've posted a possible solution, which I think has some advantages, although a part of me thinks it's more complex. I hope it's just the novelty factor of Combine.

apps/oeh/ios/App/SwiftPackages/NYPLReader/Sources/ReaderModule.swift
```diff
@@ -127,41 +127,34 @@ extension ReaderModule {
...
+            vc.navigationItem.backBarButtonItem = backItem
+            vc.hidesBottomBarWhenPushed = true
+            navVC.pushViewController(vc, animated: true)
         }
 
-        do {
-            let readerViewController = try format
-                .makeReaderViewController(
-                    for: publication,
+        _ = {
```
ettore, line 141: I ended up removing the `Just` / `flatMap` combination and changing that to a simple closure that returns a Publisher. I realized that Just + flatMap was simply mapping the `formats` array to a custom Publisher that may return errors based on various situations:
1. array emptiness
2. publication format not supported
3. error creating the VC

However.... I can do that myself without Just/flatMap!

The formal declaration of the closure (unneeded because of type inference) would be:
```
( ) ->  AnyPublisher<UIViewController, ReaderError>
```

We could also extract that into its own function, but since it's only used in this context, I actually think it's simpler to leave it there, so that we can see the flow. 

Open to thoughts. Not urgent!

apps/oeh/ios/App/SwiftPackages/NYPLReader/Sources/ReaderModule.swift
```diff
@@ -127,41 +127,34 @@ extension ReaderModule {
...
-        }
-
-        completion()
+                    resourcesServer: self.publicationServer)
+            } else {
+                return Fail(error: .formatNotSupported).eraseToAnyPublisher()
+            } }()
+            .sink { sinkCompletion in
+                if case .failure(let error) = sinkCompletion {
+                    self.delegate?.presentError(error, from: navVC)
```
ettore, line 152: @hokei actually I don't think this is the bug I thought it was.... because without a delegate we cannot present any errors. We should log this though, but we don't have a way to do so yet. I'll add a TODO for the future and file a ticket.

PR #624, opened 2023-03-06
OE-810: Skin onboarding

Toxiapo:
This PR:
- Updates the Agreement page
- Updates the Sign in page
- Updates the First Book Form


apps/oeh/src/chakra/theme/index.ts
```diff
@@ -0,0 +1,12 @@
+import { extendTheme } from "@chakra-ui/react";
+import { baseTheme } from "ui/chakra";
+import global from "./foundation/global";
+
+const hybridTheme = extendTheme({
+  ...baseTheme,
+  styles: {
+    global
+  }
+});
```
Toxiapo, line 10: I created this theme for OEH to rest some of the global styles.

apps/oeh/src/components/Header.tsx
```diff
@@ -2,24 +2,10 @@ import { useEffect, useState } from "react";
...
-  useEffect(() => {
-    setIsDarkMode(prefersDark.matches);
-    const listener = (event: MediaQueryListEvent) => {
-      setIsDarkMode(event.matches);
-    };
-    prefersDark.addEventListener("change", listener);
-    return () => {
-      prefersDark.removeEventListener("change", listener);
-    };
-  }, [prefersDark]);
```
Toxiapo, line 22: @oliviawongnyc I moved this and created a hook from it, I needed to check the color mode for the sign-in page as well.

apps/oeh/src/main.tsx
```diff
@@ -10,7 +10,7 @@ ReactDOM.createRoot(document.getElementById("root") as HTMLElement).render(
   <React.StrictMode>
     <UserAgreementProvider>
       <CredentialsProvider>
-        <ChakraProvider theme={baseTheme}>
+        <ChakraProvider theme={baseTheme} resetCSS={false}>
```
Toxiapo, line 13: This is to turn off the reset CSS that's being injected in the header from Chakra. This includes removing the default HTML styles for semantic tags like headings, p, a, etc. If we need to bring this back, we should do it in the ionic CSS.

apps/oeh/src/pages/Settings.tsx
```diff
@@ -1,44 +1,37 @@
...
-                  </IonText>
-                  <SignOut />
-                  <HelpAndFeedback />
-                  <TermsAndConditions />
-                </IonContent>
-              </>
-            );
-          }}
-        />
-      </IonPage>
```
Toxiapo, line 32: I am unsure why we have a first book sign-in form on the settings page. I think we should just redirect the user to the sign-in page instead.

I also think that the "My books" page should also perform a auth redirect. Currently, it shows a "You are not signed in" message. But let me know what you all think!

apps/oeh/src/styles/variables.css
```diff
@@ -74,6 +74,35 @@ http://ionicframework.com/docs/theming/ */
...
+  --zindex-docked: 10;
+  --zindex-dropdown: 1000;
+  --zindex-sticky: 1100;
+  --zindex-banner: 1200;
+  --zindex-overlay: 1300;
+  --zindex-modal: 1400;
+  --zinex-popover: 1500;
+  --zindex-skipLink: 1600;
+  --zindex-toast: 1700;
+  --zindex-tooltip: 1800;
```
Toxiapo, line 105: I copied this from our chakra/theme folder; it would be easier for us to work with variables instead of random digits.

Toxiapo: This should probably be a `zindex.css` file too... for discoverability.

apps/oeh/src/components/UserAgreement.tsx
```diff
@@ -1,132 +1,543 @@
 const UserAgreement: React.FC = () => {
   return (
     <div className="ion-padding">
+      <h1>Open eBooks App End User License Agreement</h1>
+      <p>Last Updated: 05/01/2016</p>
```
Toxiapo, line 5: This page has the same exact copy from https://www.openebooks.org/app-user-agreement

oliviawongnyc (changes requested):
Looks sooooo good! And thank you for all that great work you did to share code and styles. You've set us up well for moving forward!

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -1,63 +1,139 @@
...
+      code: ""
+    });
+    setCode(code);
   };
 
+  const pinChange = (e: CustomEvent<InputChangeEventDetail>) => {
+    e.preventDefault();
+    setErrors({ ...errors, pin: "" });
+    // Numbers only
+    const pin = e.detail.value?.replace(/\D/g, "") ?? "";
```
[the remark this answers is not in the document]
samanthaandrews: I do think these questions should be answered in a meeting/conversation with Jen!

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -1,63 +1,139 @@
...
+
+    // TODO: Handle CM errors better when we have the error setup.
+    presentAlert({
+      header: response?.title ?? "Connection failed",
+      message: response?.detail ?? "The request timed out.",
+      buttons: ["OK"]
+    });
+  };
+
+  const codeChange = (e: CustomEvent<InputChangeEventDetail>) => {
```
samanthaandrews, line 77: Let's rename this to `accessCodeChange` for clarity

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -1,63 +1,139 @@
...
+    e.preventDefault();
+    setErrors({ ...errors, pin: "" });
+    // Numbers only
+    const pin = e.detail.value?.replace(/\D/g, "") ?? "";
+    setPin(pin);
+  };
+
+  const containsError = !!errors.code || !!errors.pin;
+  const formFilled = pin && code;
+  const shouldDisableBtn = containsError || !formFilled;
```
samanthaandrews, line 97: let's rename this to `buttonIsDisabled`

apps/oeh/src/components/UserAgreement.tsx
samanthaandrews: We will eventually turn this into a shared package so that we don't duplicate content, but for now this is fine!

samanthaandrews (commented):
Just a couple small things from me! I do think sitting down briefly with Jen & @oliviawongnyc to make sure everyone's on the same page about error handling for sign in is a good idea. It's the first time we're explicitly asking for error handling designs so there will be some good discussion about what devs can do versus what design is thinking about.

apps/oeh/src/components/FirstbookForm.tsx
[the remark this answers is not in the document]
Toxiapo: Good idea, I will set up a meeting.

Re: Do we really want to not allow users to type letters into the PIN input? 

I should take out that regex rule because we don't need it anymore. Per Jen's request, the Pin field should show the number keyboard, hence I set the type to `"tel"`. If we run the app on a device, the input keyboard wouldn't show letter keys.

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -1,63 +1,139 @@
...
-      console.log("success", await resp.json());
+      console.log("success", response);
       setCredentials({ token: header });
     } else {
-      console.error("fail", await resp.json());
+      console.error("fail", response);
     }
+
+    // An invalid credentials is entered.
+    if (resp.status === 401) {
```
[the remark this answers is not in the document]
Toxiapo: Yeah, this is something I should confirm with Jen as well. The API doesn't tell us which one was correct so we will have to either take the errors out from the input or put it elsewhere.

apps/oeh/src/styles/settings.css
```diff
@@ -58,6 +53,10 @@
   margin-right: 1.25rem;
 }
 
+.settings .help-card ion-button::part(native) {
```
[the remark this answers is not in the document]
Toxiapo: OHH it must be a wrong read on the value. I will update this.

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -1,63 +1,139 @@
...
+      <IonItem className={`${errors.pin ? "ion-invalid" : ""}`}>
+        <IonLabel position="floating">PIN</IonLabel>
         <IonInput
-          value={password}
-          placeholder="Enter Pin"
-          type="password"
-          onIonChange={e => setPassword(e.detail.value ?? "")}
+          value={pin}
+          placeholder="Enter PIN"
+          type="tel"
```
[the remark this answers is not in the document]
Toxiapo: This is so that it shows the number-only keyboard on a mobile device.

apps/oeh/src/components/FirstbookForm.tsx
```diff
@@ -1,63 +1,139 @@
...
+            className="visibility"
+            onClick={() => setReveal(false)}
+          />
+        )}
+        {pin && !reveal && (
+          <Visibility className="visibility" onClick={() => setReveal(true)} />
+        )}
       </IonItem>
-      <IonButton type="submit" expand="block" style={{ margin: 16 }}>
+      <IonButton type="submit" expand="block" disabled={shouldDisableBtn}>
```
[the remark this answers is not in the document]
Toxiapo: It is using #50555C. I think it could be that Ionic has an `opacity: .5` set for disabled buttons. I will remove that.

apps/oeh/src/hooks/useColorMode.ts
```diff
@@ -0,0 +1,24 @@
...
+    const listener = (event: MediaQueryListEvent) => {
+      setIsDarkMode(event.matches);
+    };
+    prefersDark.addEventListener("change", listener);
+    return () => {
+      prefersDark.removeEventListener("change", listener);
+    };
+  }, [prefersDark]);
+
+  return [isDarkMode];
```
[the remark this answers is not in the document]
Toxiapo: Good catch. It should be just a boolean. Updated.

apps/oeh/src/pages/sign-in/SignIn.tsx
```diff
@@ -24,36 +37,46 @@ const SignIn: React.FC = () => {
...
+            </IonButton>
+          </IonCard>
         </div>
-        <div
-          style={{
-            display: "flex",
-            flexDirection: "column"
-          }}
-        >
+        <FooterLinks>
```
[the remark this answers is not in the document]
Toxiapo: I see some frames with extra spacings. I will remove them for now and ask Jen what she thinks.

apps/oeh/src/styles/variables.css
Toxiapo: I think we should do that for all the commonly used values. I will make a follow-up ticket.

apps/oeh/src/pages/sign-in/firstbook/FirstbookPage.tsx
```diff
@@ -20,28 +16,14 @@ const FirstbookPage: React.FC = () => {
...
-        </IonToolbar>
-      </IonHeader>
-      <IonContent fullscreen>
-        <IonHeader collapse="condense">
-          <IonToolbar>
-            <IonTitle>First Book</IonTitle>
-          </IonToolbar>
-        </IonHeader>
+    <IonPage class="firstbook">
+      <Header showBackButton />
```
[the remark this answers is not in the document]
Toxiapo: Did you land on this page via a restart? I think it's because the "back" button relies on the browsing history, and if it can't find it, it will not show the button.

Toxiapo:
> Oooh also there should be that error message that appears when authentication fails, but I think we may still be waiting on Jen for that copy and placement.

Yes! Jen is working on that.

apps/oeh/src/components/Header.tsx
```diff
@@ -1,36 +1,23 @@
...
+  const isDarkMode = useColorMode();
 
   return (
     <IonHeader>
-      <IonToolbar color="secondary">
+      <IonToolbar color="secondary" mode="ios">
```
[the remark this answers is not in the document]
oliviawongnyc: @Toxiapo I'm still curious about this

oliviawongnyc (approved):
I think this looks good! Nice that Ionic has a maxLength for the input. Did we want to add 10 for the maxLength of the Access Code? I forget. Also, did we decide not to add a "Please enter valid PIN" message if the user types letters instead of numbers? We have had a lot of back and forth on this so I think this is one of those things we move forward with and fix later if Jen has more feedback.

Toxiapo:
> I think this looks good! Nice that Ionic has a maxLength for the input. Did we want to add 10 for the maxLength of the Access Code? I forget. Also, did we decide not to add a "Please enter valid PIN" message if the user types letters instead of numbers? We have had a lot of back and forth on this so I think this is one of those things we move forward with and fix later if Jen has more feedback.

Right, we decided not to set the length for the access code because we are not sure what's the constraints for the field

PR #625, opened 2023-03-06
PJR-179: Default iOS and Android starters

samanthaandrews:
Changes in this PR
- Added the app name "org.nypl.projectr" to the `capacitor.config.ts` file
- Removed the iOS and Android folders in the pjr app and re-generated them from scratch, to make sure they're all using the correct app name. This means that none of our reader code is in the pjr app, but I thought that might actually be a good thing for now, since it seems like we don't want to have to duplicate that native code in two places and should instead figure out a way to modularize it.
- Enables clearText traffic for developing on Android in dev mode
- Ensures successful iOS and Android builds. On the base branch, I was getting failures.


kristojorg (approved):
Nice! Thank you for doing this : )

PR #628, opened 2023-03-13
SMA-425: Implement User Settings

MalcolmMcFly:
**What does this do?**

- [x] Add compose dependencies to project in order to use Jetpack Compose
- [x] Moved common dependency versions to `variables.gradle` for better maintainability.
- [x] Added Proto DataStore for saving user reading settings
- [x] Now properly injecting `ReaderViewModel` with Koin DI
- [x] Added new reader settings menu item to `ActionBar`
- [x] TODO: Update reader to apply settings from DataStore

## ✅ Open eBooks Web Release Checklist ✅

- [ ] Do these changes modify environment variables? If so, have you updated env vars in the Dockerfile, the GH Action Workflows & submitted the change to AWS through DevOps?
- [ ] Has someone run through the [QA Testing Checklist](https://docs.google.com/document/d/1Sm_mFNAlbUrjT4bfJt5boQMeIA7LoPwoPa4uXqp8kFg/edit?usp=sharing) on https://staging.openebooks.org/ and confirmed everything is working as expected?


roywatson (approved):
Your review was excellent

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/ui/EpubReaderFragment.kt
```diff
@@ -49,46 +50,42 @@ class EpubReaderFragment : VisualReaderFragment(), EpubNavigatorFragment.Listene
...
-            val backgroundsColors = mutableListOf(
-                resources.getString(R.color.white),
-                resources.getString(R.color.oldLaceWhite),
-                resources.getString(R.color.black)
-            )
-            navigatorFragment.resourcePager.setBackgroundColor(Color.parseColor(backgroundsColors[appearancePref]))
+        /**
+         * Continuously listen for changes to ReaderSettings and apply them to the reader
+         */
+        readerSettingsJob = lifecycleScope.launch {
```
russellcullen, line 79: Do we need to manage the job ourselves because `lifecycleScope` is tied to the activity lifecycle owner rather than the fragment view lifecycle? 

If that's the case, I believe we can use `viewLifecycleOwner.lifecycleScope` here and it should just work

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/ui/ReaderActivity.kt
```diff
@@ -11,13 +14,17 @@ import kotlinx.coroutines.launch
...
 import org.openebooks.app.plugins.readerplugin.ui.viewmodel.ReaderViewModel
+import org.readium.r2.shared.ExperimentalReadiumApi
 
+@ExperimentalReadiumApi
 class ReaderActivity : AppCompatActivity() {
 
     private val model: ReaderViewModel by viewModels()
     private lateinit var binding: ActivityReaderBinding
     private var path: String? = null
+    private lateinit var readerSettingsBottomSheet: ReaderSettingsBottomSheet
```
russellcullen, line 27: Do we need this as a class instance for any reason, is it just used in the local method?

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/ui/components/ReaderSettings.kt
```diff
@@ -0,0 +1,169 @@
...
+    var sliderPosition by remember { mutableStateOf(readerSettings.brightness) }
+    var theme by remember { mutableStateOf(readerSettings.theme) }
+    var fontSize by remember { mutableStateOf(readerSettings.fontSize) }
+
+    LaunchedEffect(readerSettingsState.value) {
+        readerSettings = readerSettingsState.value
+        sliderPosition = readerSettings.brightness
+        theme = readerSettings.theme
+        fontSize = readerSettings.fontSize
+    }
```
russellcullen, line 66: If i understand this correctly, this lamba will be called whenever `readerSettingsState.value` changes, which then assigns the state to local vars. I assume this is because we're initially loading with an empty state before we grab the data so the `remember'd` vars wont update when the data does?

I believe we can acheive the same thing by using the arg version of `remember`:
```
val sliderPosition by remember(readerSettingsState) { mutableStateOf(readerSettingsState.value.brightness) }
val theme by remember(readerSettingsState) { mutableStateOf(readerSettingsState.value.theme) }
val fontSize by remember(readerSettingsState) { mutableStateOf(readerSettingsState.value.fontSize) }
```

russellcullen (approved):
Looks great! I posted some questions, mostly for my understanding

apps/oeh/android/app/src/main/java/org/openebooks/app/plugins/readerplugin/ui/components/ReaderSettings.kt
russellcullen: Another thought is that we might not need to manage the state internally here ourselves and can just use the values from the input state. But I assume there might be a little delay time between the user action and when state is updated since this is on disk, so updating the state in place lets us show the changes immediately

MalcolmMcFly:
@samanthaandrews I think I am just going to close this PR and move my changes to a new branch and see what happens.

kristojorg:
@MalcolmMcFly we can also rebase this on top of main and force push it. I use that strategy quite often when things like this happen. Happy to lend a hand if you want

PR #629, opened 2023-03-15
Adds search bar, displays results on Collections page

oliviawongnyc:
- [OE-889](https://jira.nypl.org/browse/OE-889): Add search bar and display Collection Page with results


apps/oeh/ios/App/Podfile
```diff
@@ -15,6 +15,7 @@ def capacitor_pods
   pod 'CapacitorFilesystem', :path => '../../../../node_modules/@capacitor/filesystem'
   pod 'CapacitorPreferences', :path => '../../../../node_modules/@capacitor/preferences'
   pod 'CapacitorSplashScreen', :path => '../../../../node_modules/@capacitor/splash-screen'
+  pod 'CordovaPlugins', :path => '../capacitor-cordova-ios-plugins'
```
kristojorg, line 18: I don't totally understand why this keeps appearing in various PRs... Do you have any ideas @samanthaandrews ? I assume it is supposed to be here since we enabled live updates previously. So I guess the question is, why isn't it already on main? It just seems to randomly pop up all the time when running `cap sync`

apps/oeh/src/components/Collection/Search.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const Search: React.FC<SearchProps> = ({ searchData, setIsSearching }) => {
+  const history = useHistory();
+  const [searchTerms, setSearchTerms] = useState<string>("");
+
+  const onSearch = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!searchData)
+      throw new Error("Cannot perform search. No search template available.");
```
kristojorg, line 19: I think we should present this in a message to the user using [Ion Alert](https://ionicframework.com/docs/api/alert). I know we don't have error handling strategies for all of these yet, but my worry is that if we throw these errors around, the user won't see any error message and nothing will happen. My personal feeling is that showing an error message alert (even if it's not perfect) is a better default. I would encourage us to do this anywhere we don't have error handling designs yet. What do yo think @oliviawongnyc @samanthaandrews ?

apps/oeh/src/lib/fetching/parseSearchData.ts
kristojorg, line 1: Okay this is something I could see us sharing between oew and oeh. The whole `lib/fetching` module actually could be moved into a `circ-manager` package with any code specific for interacting with the CM between OEH and OEW. Maybe we should make a ticket for this? Could be a good cool down sprint project

packages/models/Search/types.ts
kristojorg, line 1: This could be placed directly in the `Search/index.ts` file just for the future. No need to change it but generally the only reason to split it out is if there is too much for one index.ts file to be legible

kristojorg (approved):
Looks good to me!

apps/oeh/src/lib/fetching/parseSearchData.ts
Toxiapo: This makes sense. I think we should formalize the error handling package first before moving these.

apps/oeh/src/components/Collection/Search.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  return (
+    <form onSubmit={onSearch} role="search">
+      <IonSearchbar
+        onIonChange={e => setSearchTerms(e.detail.value ?? "")}
+        onIonCancel={() => setIsSearching(false)}
+        showCancelButton="always"
+      />
+    </form>
```
Toxiapo, line 41: Maybe this is something we can address later, but one possible accessibility issue I see is that if I click on the search icon, the focus isn't on the input field, I had to manually click on the input to start searching.

apps/oeh/ios/App/Podfile
samanthaandrews: i am so confused by this. when i added live updates, i was instructed to install `cordova-plugin-ionic`, which i did. but this didn't add anything to the Podfile when i was making these changes, and i was running the app a lot to make sure that live updates were actually working. https://github.com/NYPL/ereading-clients/pull/596/files

but this looks like it's adding `capacitor-cordova-ios-plugins` from the local path, not from `node_modules`. I just checked and `capacitor-cordova-ios-plugins` is a folder that is auto-generated by capacitor and is not committed to git (it's in the `apps/oeh/ios/.gitignore` file). so i'm not sure if this is supposed to be here or not!

apps/oeh/src/components/Collection/Search.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    const encodedSearchTerms = encodeURIComponent(searchTerms);
+    const url = createSearchUrl(
+      searchData.template,
+      searchData.url,
+      encodedSearchTerms
+    );
+    if (!url) return;
+
+    const link = `/app/discover/collection/${encodeURIComponent(url)}`;
```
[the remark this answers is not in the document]
oliviawongnyc: Good point. I added a message if the page is empty.

apps/oeh/src/components/Collection/Search.tsx
oliviawongnyc: I think this is a good catch. I'll fix it now.

PR #633, opened 2023-03-20
PJR 226/227/228: MSW, Catalog list, book details list

kristojorg:
This PR:

1. Sets up MSW with a catalog fixture based on the existing catalog feed from the backend (but slightly edited to add IDs)
2. Adds a page to view the catalog in a list format
3. Adds a book detail page with basic info and a "get" button that currently just opens the sign in modal
4. Adds stub pages for search and my books
5. Adds a simple sign in modal that currently doesn't do anything
6. Changes jest to expect test files with `.test.ts` names. We will use `.spec.ts` for playwright instead so that the jest runner doesn't accidentally pick up playwright tests and vice-versa.
7. Removes jest from PJR. We don't need it at the moment and if/when we do, we will evaluate whether jest or vitest is better. Vitest is supposed to be basically the same but use the same vite build setup that the actual app uses (one less build pipeline to maintain).

I apologize for the size of this PR... in retrospect I should have split this up into multiple


apps/pjr/ios/App/App.xcworkspace/xcshareddata/xcschemes/Open eBooks.xcscheme
```diff
@@ -15,8 +15,8 @@
             <BuildableReference
```
samanthaandrews, line 15: Interesting. Do we need this? If so, maybe we change the name to `ProjectR.xcscheme`?

apps/pjr/src/pages/WorkPage/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+          <IonToolbar>
+            <IonTitle>Loading...</IonTitle>
+          </IonToolbar>
+        </IonHeader>
+        <IonContent fullscreen>
+          <IonRefresher slot="fixed" onIonRefresh={pullToRefreshHandler}>
+            <IonRefresherContent></IonRefresherContent>
+          </IonRefresher>
+          <div>loading...</div>
+          {/* <ListLoader /> */}
```
samanthaandrews, line 54: Can we remove this?

apps/pjr/src/pages/Discover.tsx
```diff
@@ -10,36 +10,37 @@ import {
...
+          <IonToolbar>
+            <IonTitle>All Books</IonTitle>
+          </IonToolbar>
+        </IonHeader>
+        <IonContent fullscreen>
+          <IonRefresher slot="fixed" onIonRefresh={pullToRefreshHandler}>
+            <IonRefresherContent></IonRefresherContent>
+          </IonRefresher>
+          <div>loading...</div>
+          {/* <ListLoader /> */}
```
samanthaandrews, line 40: Can we remove this?

apps/pjr/src/pages/WorkPage/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { Api } from "backend";
+import { PjrWork } from "backend/pjr-opds";
+import classNames from "classnames";
+import BookCover from "components/BookCover";
+import { arrayToString } from "lib/opds-utils";
+import { useEffect, useRef, useState } from "react";
+import { RouteComponentProps } from "react-router";
+import useSWR from "swr";
+
+import classes from "./style.module.css";
```
Toxiapo, line 23: Will this module.css method works with styling Ionic's shadow nodes where you had to style them using the `::part` selector? or how difficult would be if we had to do that? @oliviawongnyc and I have been trying to scope CSS for different pages. For now, we are just scoped with a global CSS class name and prefix the CSS styles with that, but I think the module method might be a better solution and easier to work with.

kristojorg: That is a good question, I haven't tried styling the shadow parts yet... I'll test it out today

apps/pjr/ios/App/App.xcworkspace/xcshareddata/xcschemes/Open eBooks.xcscheme
kristojorg: I think this is now gone

apps/pjr/src/pages/WorkPage/index.tsx
kristojorg: I haven't had a chance to test the shadow part styling... I think it should work but am not positive. I think maybe it makes sense for all of us to chat about how styling has been going at the start of the next cycle and evaluate whether we need to change our process at all, use modules or use postcss or something?

apps/pjr/src/App.tsx
```diff
@@ -31,6 +31,9 @@ import "@ionic/react/css/display.css";
 import "./theme/variables.css";
 
 import Discover from "pages/Discover";
+import WorkPage from "pages/WorkPage";
+import Search from "pjr/pages/Search";
+import MyBooks from "pjr/pages/MyBooks";
```
Toxiapo, line 36: What are the differences between these two imports? Shouldn't the root directory be "pjr"  if we are on the pjr app?

apps/oeh/vite.config.ts
```diff
@@ -40,7 +40,6 @@ export default defineConfig({
     outDir: "../dist",
     minify: false,
     emptyOutDir: true,
-    target: "esnext",
```
Toxiapo, line 43: Did you remove this or it was unintentional?

apps/pjr/src/components/Loader.tsx
```diff
@@ -0,0 +1,20 @@
+import { IonSpinner } from "@ionic/react";
+
+export default function Loader() {
+  return (
+    <div
+      style={{
+        display: "grid",
+        // gridTemplateRows: "auto auto",
```
Toxiapo, line 8: nit: comments

apps/pjr/src/backend/pjr-opds.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+}
+
+export interface PjrFeed extends Feed {
+  metadata: {
+    title: string;
+    id: string;
+  };
+  publications: Array<PjrWork>;
+}
```
Toxiapo, line 25: These are now scoped by prj,  but I am assuming we can re-use these on OEH in the future?

Toxiapo (commented):
Just a couple config questions!

apps/pjr/src/env/index.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * This is hardcoded for the moment because vite exposes env variables on import.meta.env while
+ * playwright uses process.env in commonjs mode. So I can't figure out how to access an env variable in both
+ * vite and my playwright tests and my application code at the same time. I tried turning the whole
+ * project into an es module, but that would require adding file extensions to every import.
+ *
+ * I also tried transpiling the playwright tests with tsup, but this didn't work either as it didn't replace
+ * the import.meta. I could try with babel instead, but for now I will hardcode the backend url.
+ */
```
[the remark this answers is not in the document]
samanthaandrews: in OEH i think i remember adding the file extension but only on imports within the playwright tests themselves, which seemed _ok_ although still not ideal

apps/oew/cypress/downloads/downloads.html
samanthaandrews, line 1: what even is this lol

samanthaandrews (approved):
just a few small comments. otherwise this looks good to me.

apps/pjr/src/App.tsx
samanthaandrews: also wondering about this. as far as i can tell, i think the `pjr/` prefix can be removed?

kristojorg: Good question. I was at one point struggling with getting playwright to recognize our imports relative to the typescript `baseUrl`. That's what allows us to do `import WorkPage from "pages/WorkPage";`. So I tried setting an explicit path alias from `pjr` to `./src`. The app itself never had a problem with recognizing either one, so both of these worked. It was only within files that playwright imported that I had an issue. 

I _think_ I tracked the problem down to something else though, so I'll try undoing this path alias

apps/oeh/vite.config.ts
kristojorg: I noticed a warning from Vite that said the `legacy` plugin used above was overriding this setting, so we should remove one or the other. I removed this one since what we want is the target set by the `legacy` plugin.

apps/pjr/src/env/index.ts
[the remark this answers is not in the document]
kristojorg: Yea I can't figure out whose fault it is tbh. Vite is using the ES Modules way of doing things (`import.meta`) but turning your entire project into es modules is still quite a challenge, especially once typescript gets involved. The underlying problem is that each of these packages has its own build system, and they have incompatible requirements. When you include Jest too, there are then three different build pipelines. What would help this is if:
1. Vite could transpile the tests. Vite has Vitest which is basically a drop in replacement for Jest that uses the Vite build pipeline. Would be nice if they could do something similar for Playwright, although I'm not sure how that would work.
2. We could configure our whole app as an ES Module. This is _maybe_ possible now that Typescript 5 is out, which allows us to import ES Modules without using the `.mjs` extension on every import. Maybe I give this a shot during cooldown.

apps/pjr/src/pages/WorkPage/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+   *
+   * We may not even use that element anyways so this point is probably moot.
+   */
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    // prevent browser reload
+    e.preventDefault();
+    const form = e.currentTarget;
+    const formData = new FormData(form);
+    const data = Object.fromEntries(formData.entries());
+    console.log(data, e.currentTarget.elements);
```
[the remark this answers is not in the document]
kristojorg: This whole component is currently unfinished waiting on designs and the borrow/return endpoint

apps/oew/cypress/downloads/downloads.html
kristojorg: I have no idea haha

kristojorg: I'm gonna gitignore it I think

kristojorg: Aaaand it's magically gone

PR #634, opened 2023-03-22
OE-881: Add splash screen and icons

Toxiapo:


Toxiapo:
Hi @russellcullen, I've noticed that on Android 12 and above, the launch screen image is rendered from the mipmap folder, whereas for versions 11 and below, the icon from the drawable folders is used instead. Can you please confirm if this behavior is expected or if there is a configuration option available to use the same resource for both versions, if desirable? 

Additionally, I have separate icon designs for night and light modes, and I have included the night folder for both mipmaps and drawable. Though, this has resulted in a large number of images for the app. Do you think this is OK or can we reduce the number of resources? AFAIK, the minimum android we support is Android 6. Thanks!

russellcullen:
@Toxiapo I believe that is expected for the launch icon but I will double check and get back to you to be sure. 

And regarding the additional images, this should be ok as not all the resources get shipped to the end user. I believe we build the app using App Bundles, which ensures that only the resources needed for a particular device are bundled to the end user. 

That being said, both night and light mode images will be bundled for a particular density but nothing we can do about that. A bit of expected extra images we can't avoid

apps/oeh/ios/App/App/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,116 +1,116 @@
...
     {
-      "size" : "29x29",
+      "filename" : "AppIcon-29@2x~ipad.png",
       "idiom" : "iphone",
-      "filename" : "AppIcon-29x29@2x-1.png",
-      "scale" : "2x"
+      "scale" : "2x",
+      "size" : "29x29"
     },
     {
```
Toxiapo, line 21: Hi @ErnestFan, would you mind helping me take a look at these settings and let me know if everything looks good, just making sure I am not breaking anything!

These icons were generated using a Capacitor plugin, they should support dark and light modes, for both the app icons and splash screens. thank you!

Toxiapo: Sure! I will send it through slack.

apps/oew/package.json
```diff
@@ -38,7 +38,7 @@
     "@newrelic/next": "^0.3.1",
     "@next/mdx": "^12.3.1",
     "@nypl/dgx-svg-icons": "^0.3.12",
-    "@nypl/opds": "*",
+    "@nypl/opds": "^0.1.0-alpha.0",
```
kristojorg, line 41: This should be changed back to use the `"*"`

kristojorg (changes requested):
Just the one thing. Crazy how many icon files there are in an app!

apps/oew/package.json
Toxiapo: Updated!

samanthaandrews:
it really is crazy how many icon files there are. thank you for doing this @Toxiapo !

apps/oeh/android/app/src/main/res/drawable-v24/ic_baseline_add_24.xml
russellcullen, line 1: nitpick: is this file used? Might be safe to delete if not

Toxiapo: Yup, the `activity_bookmarks.xml` config file is referencing it.

russellcullen: I think this might be the same as the one in `drawable/ic_baseline_add_24.xml`, which will be used as a default if this file is missing. In theory this icon will be used on android versions 24+ (since its in `res/drawable-v24`), but if theres no difference then I dont think it's needed

Toxiapo: ah yes, you are correct, they are the same! I am happy to remove this if it's a duplicate.

russellcullen (approved):

kristojorg (approved):
Looks good!

PR #636, opened 2023-03-27
OE-882: Automate my book page workflow

ayan1229:


apps/oew/playwright/features/borrow-return-books.feature
```diff
@@ -42,3 +42,16 @@ Scenario: As a user I should be able to borrow a book, open it, read it
...
+ Scenario: As a user I should not see the returned book on My Books page
+    Given I go to the "sign in" page
+    Then I click the "First Book sign in button"
+    And I fill in the "access code input" with "OPENEBOOKS_ALL_ACCESS_USER_USERNAME"
+    And I fill in the "PIN input" with "OPENEBOOKS_ALL_ACCESS_USER_PW"
+    And I click the "sign in button"
+    And the "open ebooks header" should be displayed
+    Then I click the "my books link"
+    And the "my books header" should be displayed
+    And the "Can You Keep A Secret?" should not be displayed
```
samanthaandrews, line 57: I kind of think we should just put these three assertions in the previous scenario. That way we don't have to do the full login process just to check that the book was returned.

samanthaandrews, line 57: Let's rename this `"Can You Keep A Secret?"` to `"Can You Keep A Secret? link"` so that it's clear what element/locator we are looking for on this page.

samanthaandrews (changes requested):
Just a few small things!

samanthaandrews (approved):
LGTM!

